Issue #3, opened 2019-06-04
Replace SomePlaceInMemory with VolatileRef


sameo:
In the `read_obj/write_obj` unit tests methods.

bonzini:
Fixed by #5.

PR #5, opened 2019-06-05
 Replace SomePlaceInMemory with VolatileRef

bonzini:
The virtio tests use a SomeplaceInMemory struct which is pretty much a duplicate of `vm_memory::VolatileRef`, let's drop it to avoid duplication and to provide a realistic example of how to use `VolatileMemory`/`VolatileSlice`/`VolatileRef`.


bonzini:
Fixes #3

PR #10, opened 2019-10-31
Refine the VirtioDevice trait

jiangliu:
This is an RFC PR, which depends on https://github.com/rust-vmm/vm-device/pull/9 and https://github.com/rust-vmm/vm-device/pull/11

It refines the VirtioDevice trait to better support virtio backends.
Also introduce VirtioQueueConfig and VirtioDeviceConfig to manage
virtio device configuration information.

Signed-off-by: Xingjun Liu <xingjun.liu@linux.alibaba.com>
Signed-off-by: Liu Jiang <gerry@linux.alibaba.com>

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -4,11 +4,85 @@
 //
 // SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
 
+use super::queue::Queue;
 use super::*;
-use std::sync::atomic::AtomicUsize;
 use std::sync::Arc;
+use vm_device::interrupt::{InterruptIndex, InterruptSourceGroup, InterruptSourceType};
+use vm_device::resources::DeviceResourceRequest;
 use vm_memory::GuestMemory;
 use vmm_sys_util::eventfd::EventFd;
+
+/// Data struct to manage an virtio queue.
+pub struct VirtioQueueConfig<'a, M: GuestMemory> {
+    /// A virtque.
+    pub queue: Queue<'a, M>,
+
+    /// And associated event notification fd.
+    pub queue_evt: EventFd,
+}
+
+/// Configuration information for an virtio device.
+///
+/// The ownership of an virtio device configuration object will be moved into the VirtioDevice
+/// object when VirtioDevice::activate() method gets called, and it should be returned back
+/// if VirtioDevice::activate() fails or when VirtioDevice::reset() gets called.
+pub struct VirtioDeviceConfig<'a, M: GuestMemory> {
+    /// Memory object to access the guest memory.
+    pub mem: M,
+
+    /// Virtqueues of the virtio device.
+    pub queues: Vec<VirtioQueueConfig<'a, M>>,
+
+    /// A group of interrupts to notify the guest for queue events.
+    pub intr_group: Arc<Box<dyn InterruptSourceGroup>>,
+}
+
+impl<'a, M> VirtioDeviceConfig<'a, M>
+where
+    M: GuestMemory,
+{
+    /// Constructs an virtio device configuration object.
+    pub fn new(
+        mem: M,
+        queues: Vec<VirtioQueueConfig<'a, M>>,
+        intr_group: Arc<Box<dyn InterruptSourceGroup>>,
+    ) -> Self {
+        VirtioDeviceConfig {
+            mem,
+            queues,
+            intr_group,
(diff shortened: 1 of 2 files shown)


sameo:
@andreeaflorescu you may want to have a look here.

andreeaflorescu:
Yap, I'll take a look next week once I am back home.

src/device.rs
```diff
@@ -4,12 +4,86 @@
 //
 // SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
 
+use super::queue::Queue;
 use super::*;
-use std::sync::atomic::AtomicUsize;
 use std::sync::Arc;
+use vm_device::interrupt::{InterruptIndex, InterruptSourceGroup, InterruptSourceType};
```
andreeaflorescu: We have to think this through. Right now it looks like a dependency on vm-device from vm-virtio will introduce a cyclic dependency between those two. vm-virtio device depends on vm-device because it needs interrupts and vm-device depends on vm-virtio because the DeviceManager manages devices including VirtioDevices.

What are your thoughts on this?

src/device.rs
```diff
@@ -4,12 +4,86 @@
...
+        match self.intr_group.get_type() {
+            InterruptSourceType::LegacyIrq => self.intr_group.trigger(0, flags),
+            InterruptSourceType::MsiIrq | InterruptSourceType::VfioMsiIrq(_, _) => {
+                self.intr_group.trigger(index, 0)
+            }
+        }
+    }
+
+    /// Get all irqfds associated with vritio queues.
+    pub fn get_vring_irqfds(&self) -> Vec<&EventFd> {
```
andreeaflorescu: Is there a usecase where both LegacyIrq and MsiIrq are supported at the same time? It looks like the implementation of the DeviceConfig is wildly different depending on the interrupt system. Each function here is converted in a match on the type of interrupt.

sameo:
On Mon, Nov 04, 2019 at 04:20:48AM -0800, Andreea Florescu wrote:
> andreeaflorescu commented on this pull request.
> >  use std::sync::Arc;
> +use vm_device::interrupt::{InterruptIndex, InterruptSourceGroup, InterruptSourceType};
> 
> We have to think this through. Right now it looks like a dependency on vm-device from vm-virtio will introduce a cyclic dependency between those two. vm-virtio device depends on vm-device because it needs interrupts and vm-device depends on vm-virtio because the DeviceManager manages devices including VirtioDevices.
>

I think DeviceManager or the whole vm-device crate should not have any
dependency against vm-virtio. DeviceManager will manage devices
implementing a set of vm-device defined traits (like e.g. the IO one).

Cheers,
Samuel.

jiangliu:
I have given an overview of the rust-vmm project on OpenInfra SH forum.
The status becomes a little bit clear with the following pic, only dependency from higher block to lower block is allowed.
image

src/device.rs
jiangliu: Per PCI spec, a device will work in Legacy mode after reset, and may switch to/from PCI MSI/MSI-x mode when OS programs the interrupt configuration registers. So it's a must for a PCI device to support both legacy and MSI/MSIx interrupts.

src/device.rs
jiangliu: As @sameo  mentioned, the vm-virtio has dependency on the vm-device crate, but not vice versa.

andreeaflorescu: Ok, I'll look again at the vm-device crate. What PR should I look at to understand the design of interrupts?

src/device.rs
andreeaflorescu: In this case, can we have `VirtioDeviceConfig` as a trait as well? What we have here can be the implementation of the trait and in Firecracker's case we can implement our own Device Configuration which should be much simpler.

sameo:
On Tue, Nov 05, 2019 at 01:33:25AM -0800, Andreea Florescu wrote:
> >  use std::sync::Arc;
> +use vm_device::interrupt::{InterruptIndex, InterruptSourceGroup, InterruptSourceType};
> 
> Ok, I'll look again at the vm-device crate. What PR should I look at to understand the design of interrupts?
>

https://github.com/rust-vmm/vm-device/pull/11

jiangliu:
This has been moved to vm-device, so close it.

PR #13, opened 2020-02-05
Remove field `mem: &'a M` from Struct Queue

### LINKED ISSUE #12: Remove field `mem: &'a M` from Struct Queue
Currently the Struct Queue is defined as:
```rust
#[derive(Clone)]
/// A virtio queue's parameters.
pub struct Queue<'a, M: GuestMemory> {
    mem: &'a M,

    /// The maximal size in elements offered by the device
    max_size: u16,

    next_avail: Wrapping<u16>,
    next_used: Wrapping<u16>,
}
```
The `mem` field is reference to an object of GuestMemroy. We encountered several issues with this design,
1) we need to explicitly carries the lifetime parameter,
2) we need to explicitly declare the `M` type parameter,
3) it enforces the guest memory object used by the Queue object is the same type, which forbids optimization for enabling memory hotplug, which may have a GuestMemory type and GuestMemory snapshot type.
So how about reverting to the original design by removing field `mem` from struct Queue?

jiangliu:
The `struct Queue` has a field named `mem`, which is a shared refrence
to a `GuestMemory` object, which has several drawbacks:
1) must explicitly carries the lifetime parameter,
2) must explicitly declare the M type parameter,
3) enforces the guest memory object used by the Queue object to be the
same type, which forbids optimization for enabling memory hotplug.

For virtio device to support memory hotplug, the programming pattern
may be:

```
impl<M: GuestMemory> VirtioPseudoDevice<M> {
	fn new(mem: M) {
		VirtioPseudoDevice {
			mem,
			queue: Queue::new(256),
			...
		}
	}

	fn process_virtio_queue(&mut self) {
		let snapshot = self.mem.snapshot();
		let iter = self.queue.iter(snapshot);

		while iter.has_next() {
			// reuse snapshot to improve performance
			....
		}
	}
}
```

Fix #12

### DIFF
--- a/coverage_config_aarch64.json
+++ b/coverage_config_aarch64.json
@@ -1,5 +1,5 @@
 {
-  "coverage_score": 79.9,
+  "coverage_score": 74.6,
   "exclude_path": "",
   "crate_features": ""
 }
--- a/coverage_config_x86_64.json
+++ b/coverage_config_x86_64.json
@@ -1,5 +1,5 @@
 {
-  "coverage_score": 84.3,
+  "coverage_score": 80.0,
   "exclude_path": "",
   "crate_features": ""
 }
--- a/src/device.rs
+++ b/src/device.rs
@@ -46,7 +46,7 @@ pub trait VirtioDevice<A: GuestAddressSpace>: Send {
         mem: A,
         interrupt_evt: EventFd,
         status: Arc<AtomicUsize>,
-        queues: Vec<Queue<A::M>>,
+        queues: Vec<Queue>,
         queue_evts: Vec<EventFd>,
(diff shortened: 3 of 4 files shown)


bonzini:
Actually, please open a separate issue. I do not think we should remove the mem field, the point of the hotplug support was to make it unnecessary.

jiangliu:
> Looks good but can you add a testcase?

Found it's a little hard to add a test case for this scenario:(

jiangliu:
> Actually, please open a separate issue. I do not think we should remove the mem field, the point of the hotplug support was to make it unnecessary.

Sure, I will open another PR for the bugfix/improvements/new features.

PR #14, opened 2020-03-19
update for vm-memory 0.2.0

bonzini:
* update rust-vmm-ci because the old version fails with vm-memory's usage of `try_from`

* compile with latest vm-memory (`new` -> `from_ranges` rename)

* avoid checked_offset and instead use `get_slice` and `VolatileArrayRef`

* replace associated type `VirtioDevice::M` with type parameter `VirtioDevice<M>`

* replace `&'a M: GuestMemory` with `GuestAddressSpace`, which is possible now that `GuestAddressSpace::T` is cloneable.

There is no abstraction cost for static guest memory map.


.buildkite/pipeline.yml
```diff
@@ -8,7 +8,7 @@ steps:
       platform: x86_64.metal
     plugins:
       - docker#v3.0.1:
-          image: "fandree/rust-vmm-dev"
+          image: "rustvmm/dev:v3"
```
andreeaflorescu: This should be deleted if we're switching to rust-vmm-ci.

bonzini: I just copied what vm-memory is doing... Is that wrong as well?

andreeaflorescu: The vm-memory setup is correct. The vm-memory has a specialized pipeline because it has features and the default buildkite pipeline does not have support for features. You'll notice that the Buildkite pipeline in the vm-memory repository only runs the tests related to the mmap feature: https://github.com/rust-vmm/vm-memory/blob/master/.buildkite/pipeline.linux.yml

For the custom, default tests we can use the existing pipeline.

Cargo.toml
```diff
@@ -7,15 +7,16 @@ repository = "https://github.com/rust-vmm/vm-virtio"
...
+edition = "2018"
 
 [dependencies]
 byteorder = "=1.2.1"
 epoll = "=4.0.1"
 libc = ">=0.2.39"
 log = "=0.4.6"
-vm-memory = { git = "https://github.com/rust-vmm/vm-memory" }
+vm-memory = ">=0.2.0"
 vmm-sys-util = { git = "https://github.com/rust-vmm/vmm-sys-util" }
```
jiangliu: Could you please also change to use package from crates.io?

Cargo.toml
```diff
@@ -7,15 +7,16 @@ repository = "https://github.com/rust-vmm/vm-virtio"
...
 license = "Apache-2.0 OR MIT"
+edition = "2018"
 
 [dependencies]
 byteorder = "=1.2.1"
 epoll = "=4.0.1"
 libc = ">=0.2.39"
 log = "=0.4.6"
-vm-memory = { git = "https://github.com/rust-vmm/vm-memory" }
+vm-memory = ">=0.2.0"
```
jiangliu: How about "~0, >= 0.2.0"?

sameo:
@bonzini @jiangliu @andreeaflorescu PR updated with all comments addressed. PTAL.

Cargo.toml
```diff
@@ -7,6 +7,7 @@ repository = "https://github.com/rust-vmm/vm-virtio"
 keywords = ["virtio"]
 readme = "README.md"
 license = "Apache-2.0 OR MIT"
+edition = "2018"
```
andreeaflorescu: This commit is not signed :(

tests/coverage
```diff
@@ -1 +1 @@
-86.9
+85.6
```
andreeaflorescu: This commit is also not signed. I assumed it was created with GitHub?

Cargo.toml
sameo: Good catch. It should be fixed now.

tests/coverage
sameo: This one is removed.

sameo:
> The readme needs to be updated as well to remove the section about testing and link the documentation from rust-vmm-ci instead.

Done.

PR #15, opened 2020-04-04
Partially revert #14

jiangliu:
#14 changes struct Queue to use GuestAddressSpace instead of GuestMemory, which may potentially introduce bugs when memory hotplug is enabled. So partially revert #14 to use GuestMemory for Queue.

### DIFF
--- a/coverage_config_aarch64.json
+++ b/coverage_config_aarch64.json
@@ -1,5 +1,5 @@
 {
-  "coverage_score": 79.9,
+  "coverage_score": 78.5,
   "exclude_path": "",
   "crate_features": ""
 }
--- a/coverage_config_x86_64.json
+++ b/coverage_config_x86_64.json
@@ -1,5 +1,5 @@
 {
-  "coverage_score": 84.3,
+  "coverage_score": 85.7,
   "exclude_path": "",
   "crate_features": ""
 }
--- a/src/device.rs
+++ b/src/device.rs
@@ -18,7 +18,7 @@ use vmm_sys_util::eventfd::EventFd;
 /// called and all the events, memory, and queues for device operation will be moved into the
 /// device. Optionally, a virtio device can implement device reset in which it returns said
 /// resources and resets its internal state.
-pub trait VirtioDevice<M: GuestAddressSpace>: Send {
+pub trait VirtioDevice<A: GuestAddressSpace>: Send {
     /// The virtio device type.
     fn device_type(&self) -> u32;
 
@@ -43,10 +43,10 @@ pub trait VirtioDevice<M: GuestAddressSpace>: Send {
     /// Activates this device for real usage.
     fn activate(
         &mut self,
-        mem: M,
+        mem: A,
         interrupt_evt: EventFd,
         status: Arc<AtomicUsize>,
-        queues: Vec<Queue<M>>,
+        queues: Vec<Queue<A::M>>,
         queue_evts: Vec<EventFd>,
     ) -> ActivateResult;
 
--- a/src/queue.rs
+++ b/src/queue.rs
@@ -12,8 +12,7 @@ use std::num::Wrapping;
 use std::sync::atomic::{fence, Ordering};
 
 use vm_memory::{
-    Address, ByteValued, Bytes, GuestAddress, GuestAddressSpace, GuestMemory, GuestUsize,
-    VolatileMemory,
+    Address, ByteValued, Bytes, GuestAddress, GuestMemory, GuestUsize, VolatileMemory,
 };
 
 pub(super) const VIRTQ_DESC_F_NEXT: u16 = 0x1;
@@ -56,8 +55,8 @@ struct Descriptor {
 unsafe impl ByteValued for Descriptor {}
 
 /// A virtio descriptor chain.
-pub struct DescriptorChain<M: GuestAddressSpace> {
-    mem: M::T,
+pub struct DescriptorChain<'a, M: GuestMemory> {
+    mem: &'a M,
     desc_table: GuestAddress,
     queue_size: u16,
     ttl: u16, // used to prevent infinite chain cycles
@@ -79,9 +78,9 @@ pub struct DescriptorChain<M: GuestAddressSpace> {
     pub next: u16,
 }
 
-impl<M: GuestAddressSpace> DescriptorChain<M> {
+impl<'a, M: GuestMemory> DescriptorChain<'a, M> {
     fn read_new(
-        mem: M::T,
+        mem: &'a M,
         desc_table: GuestAddress,
         queue_size: u16,
         ttl: u16,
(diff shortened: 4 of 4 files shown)


bonzini:
> First, the typical flow to handle a virtio queue descriptor is:
> 
> 1) get the next available descriptor.
> 
> 2) access guest memory according to the guest physical address in the descriptor.
> 
> With current implementation, we have no way to ensure step 1 and step 2
> uses the same GuestMemory snapshot when memory hotplug is enabled.

Why would it be a problem to access memory using the same snapshot in the two steps?

jiangliu:
> > First, the typical flow to handle a virtio queue descriptor is:
> > 
> > 1. get the next available descriptor.
> > 2. access guest memory according to the guest physical address in the descriptor.
> > 
> > With current implementation, we have no way to ensure step 1 and step 2
> > uses the same GuestMemory snapshot when memory hotplug is enabled.
> 
> Why would it be a problem to access memory using the same snapshot in the two steps?

Let take an imaged Virtio driver as an example:)
```
struct VirtioDevice<A: GuestAddressSpace> {
       guest_as: A,
       queue: Queue<A>,
}

impl<A: GuestAddressSpace> VirtioDevice<A> {
       fn process_queue(&mut self) {
             let snapshot1 = self.geust_as.memroy();
              for avail_desc in self.queue.iter() { // snapshot2 = self.queue.mem.memory();
                      snapshot1.read_obj(avail_desc.addr);
                      self.queue.add_used(avail_desc);      // snapshot3 = self.queue.mem.memory();    
              }
      }
}
```
By now, we have no mechanism to ensure snapshot1, snapshot2 and snapshot3 are in consistence.
So we suggest to change the pattern to
```
struct VirtioDevice<A: GuestAddressSpace> {
       guest_as: A,
       queue: Queue<A::M>,
}

impl<A: GuestAddressSpace> VirtioDevice<A> {
       fn process_queue(&mut self) {
             let snapshot1 = self.geust_as.memroy();
             for avail_desc in self.queue.iter(snapshot1) { 
                    snapshot1.read_obj(avail_desc.addr);
                    self.queue.add_used(snapshot1, avail_desc);
             }
      }
}
```
By this way, the VirtioDevice could guarantee the same guest memory snapshot is used to process a specific Virtio Queue Descriptor.

bonzini:
> By now, we have no mechanism to ensure snapshot1, snapshot2 and snapshot3 are in consistence.

That's not a problem. If they are not, it is a guest bug.

jiangliu:
> > By now, we have no mechanism to ensure snapshot1, snapshot2 and snapshot3 are in consistence.
> 
> That's not a problem. If they are not, it is a guest bug.

Please refer to https://github.com/cloud-hypervisor/cloud-hypervisor/blob/master/vm-virtio/src/block.rs#L605 for the code pattern mentioned above.
Actually we can't trust the guest OS, I'm concerning about security risks:)
And when AtomicMmap is used, the proposed way also reduces atomic memory accesses caused by Arc operations.

bonzini:
>  Actually we can't trust the guest OS, I'm concerning about security risks:)

What is the security risk?

Anyway, if you need the particular `GuestMemory` that was used to access the descriptor chain, you can get it from `DescriptorChain`; just add an accessor for the `mem` field.

> And when AtomicMmap is used, the proposed way also reduces atomic memory accesses caused by Arc operations.

I think you're referring to `self.mem.clone()` in `AvailIter::next()`. I agree it's expensive, but I think it's correct. The c-h code can only keep the `GuestMemory` alive in the `for avail_desc in queue.iter(&mem)` loop because it handles I/O requests serially. In a more performant implementation, there is no requirement for the `AvailIter` to outlive the `DescriptorChain`, and therefore it's correct to clone `self.mem` when creating the `DescriptorChain`.

slp:
This is an interesting question. On one hand, using `GuestAddressSpace` removes the lifetime restrictions making each object completely independent. For instance, in CH we added some helper functions to create a new `DescriptorChain` from an existing one referencing a different `GuestMemory`. That won't be necessary with this approach.

On the other hand, I'm also kind of worried about the cost that adds on iterating not only over `DescriptorChain`s in the `Queue`, but also on the descriptors from the same chain. As someone already ran some numbers on this?

bonzini:
> On the other hand, I'm also kind of worried about the cost that adds on iterating not only over DescriptorChains in the Queue, but also on the descriptors from the same chain. As someone already ran some numbers on this?

Yes that is a problem. However I already have some patches for that, turning `DescriptorChain` into an `Iterator<Item = Descriptor>` (http://github.com/bonzini/vm-virtio/commits/desc-chain-iter). I have not pushed them to avoid forking vm-virtio further from the CH versions.

alexandruag:
Hi! Besides getting rid of the extra lifetime, are there other good reasons to use `GuestAddressSpace` instead of `&GuestMemory` in `AvailIter` and `DescriptorChain`? I haven't tried out the current `vm-virtio` implementation, but is the presence of the additional lifetime bothersome during use, or is it basically transparent?

Also, if we remove the current `&GuestMemory` member from `Queue`, does adding a `GuestAddressSpace` member help with anything other than getting rid of an extra `&GuestMemory` parameter required for methods such as `is_valid`, `iter`, and `add_used`? 

It seems to me that a bit of extra verbosity is worth trading for delegating the `memory()` calls on `GuestAddressSpace` to a higher level. At the same time, I'm also trying to understand if there are other factors to be considered here.

bonzini:
> does adding a GuestAddressSpace member help with anything other than getting rid of an extra &GuestMemory parameter required for methods such as is_valid, iter, and add_used?

Conceptually, the address space is not something that changes across the device's life. The `GuestAddressSpace` concept was added exactly to allow the memory map to change, but allow devices to know what's constant, and therefore keep the dereference of the memory to whatever needs to access the memory.

The current situation with a constant `&GuestMemory` is indeed the least flexible and in some sense the worst of both worlds. But `GuestAddressSpace` is meant to fix that problem.

slp:
> Yes that is a problem. However I already have some patches for that, turning `DescriptorChain` into an `Iterator<Item = Descriptor>` ([http://github.com/bonzini/vm-virtio/commits/desc-chain-iter](https://github.com/bonzini/vm-virtio/commits/desc-chain-iter)). I have not pushed them to avoid forking vm-virtio further from the CH versions.

I see, that looks nice. As for the divergence with CH, I think I'm responsible for most of those changes, at least in Queue. Most of them are no longer needed, after the introduction of `GuestAddressSpace`. The other are just some helpers for dealing with EVENT_IDX related fields.

Once you merge your iterator for `Descriptor`, I can rebase the EVENT_IDX changes on top of it, and then update the dependent CH's crates. @bonzini @sameo @sboeuf @rbradford what do you think?

sboeuf:
@slp 
>Once you merge your iterator for Descriptor, I can rebase the EVENT_IDX changes on top of it, and then update the dependent CH's crates. @bonzini @sameo @sboeuf @rbradford what do you think?

Sounds good to me!

bonzini:
Closing since Sergio has a different plan.

PR #16, opened 2020-04-06
WIP: port virtio Queue code from Cloud Hypervisor

jiangliu:
The virtio Queue implementation from Cloud Hypervisor project provide supports of
1) indirect descriptor
2) EVENT_IDX feature
3) queue reset logic
4) descriptor chain iterator

So port this code from Cloud Hypervisor project.
The first two patches have been submitted by #15 , so marking this PR as WIP.

### DIFF
--- a/coverage_config_aarch64.json
+++ b/coverage_config_aarch64.json
@@ -1,5 +1,5 @@
 {
-  "coverage_score": 79.9,
+  "coverage_score": 78.5,
   "exclude_path": "",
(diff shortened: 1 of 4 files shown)


bonzini:
Removing the `mem` field is wrong; it's the whole reason why we added the `GuestAddressSpace` indirection. Can you please drop the first three patches?

bonzini:
Closing since Sergio has a different plan.

PR #17, opened 2020-04-15
Make DescriptorChain an Iterator

bonzini:
Make `Descriptor` a public API strict, and turn `DescriptorChain` into an Iterator that returns them.  This is more efficient because it avoids cloning of the `GuestMemory` whenever a descriptor is created, and it is a better API that prevents the list from being accessed more than once. Consumers can still store the Descriptors in a vec if desirable.

### DIFF
--- a/src/queue.rs
+++ b/src/queue.rs
@@ -46,11 +46,45 @@ const VIRTQ_AVAIL_RING_META_SIZE: usize = VIRTQ_AVAIL_RING_HEADER_SIZE + 2;
 /// A virtio descriptor constraints with C representation
 #[repr(C)]
 #[derive(Default, Clone, Copy)]
-struct Descriptor {
+pub struct Descriptor {
     addr: u64,
+
+    /// Length of device specific data
     len: u32,
+
+    /// Includes next, write, and indirect bits
     flags: u16,
+
+    /// Index into the descriptor table of the next descriptor if flags has
+    /// the next bit set
     next: u16,
+}
+
+#[allow(clippy::len_without_is_empty)]
+impl Descriptor {
+    /// Return the guest physical address of descriptor buffer
+    pub fn addr(&self) -> GuestAddress {
+        GuestAddress(self.addr)
+    }
+
+    /// Return the length of descriptor buffer
+    pub fn len(&self) -> u32 {
+        self.len
+    }
+
+    /// Return the flags for this descriptor, including next, write and indirect
+    /// bits
+    pub fn flags(&self) -> u16 {
+        self.flags
+    }
+
+    /// Checks if the driver designated this as a write only descriptor.
+    ///
+    /// If this is false, this descriptor is read only.
+    /// Write only means the the emulated device can write and the driver can read.
+    pub fn is_write_only(&self) -> bool {
+        self.flags & VIRTQ_DESC_F_WRITE != 0
+    }
 }
 
 unsafe impl ByteValued for Descriptor {}
@@ -62,18 +96,8 @@ pub struct DescriptorChain<M: GuestAddressSpace> {
     queue_size: u16,
     ttl: u16, // used to prevent infinite chain cycles
 
-    /// Guest physical address of device specific data
-    pub addr: GuestAddress,
-
-    /// Length of device specific data
-    pub len: u32,
-
-    /// Includes next, write, and indirect bits
-    pub flags: u16,
-
-    /// Index into the descriptor table of the next descriptor if flags has
-    /// the next bit set
-    pub next: u16,
+    /// This particular descriptor
+    pub desc: Descriptor,
 }
 
 impl<M: GuestAddressSpace> DescriptorChain<M> {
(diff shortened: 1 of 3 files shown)


src/queue.rs
```diff
@@ -126,43 +135,51 @@ impl<M: GuestAddressSpace> DescriptorChain<M> {
...
+        } else {
+            self.index = self.desc.next;
+            let desc_table_size = size_of::<Descriptor>() * self.queue_size as usize;
+            let slice = self.mem.get_slice(self.desc_table, desc_table_size).ok()?;
+            self.desc = slice
+                .get_array_ref(0, self.queue_size as usize)
+                .ok()?
+                .load(self.index as usize);
+            self.ttl = self.ttl - 1;
+        }
```
slp, line 181: Nitpicking, but clippy complains about this:
```
warning: manual implementation of an assign operation
   --> src/queue.rs:180:13
    |
180 |             self.ttl = self.ttl - 1;
    |             ^^^^^^^^^^^^^^^^^^^^^^^ help: replace it with: `self.ttl -= 1`
    |
    = note: `#[warn(clippy::assign_op_pattern)]` on by default
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#assign_op_pattern
```

bonzini: Fixed.

bonzini:
@sameo can you give an approving review?

andreeaflorescu:
@bonzini Alex will also take a look at this one today. Can we wait a bit to check with him as well?

bonzini:
@andreeaflorescu Sure!

andreeaflorescu:
Thanks!

jiangliu:
> Thanks!

@alexandruag could you please help to review this PR? Thanks!

alexandruag:
Hi, many apologies for the delay :( Will review today.

jiangliu:
> Looks good, just left some minor comments.
> 
> Also, this question is not in the scope of the current PR, but I see that multiple generic structs (such as `DescriptorChain` itself) have type parameters like `M: GuestAddressSpace`. However, the associated logic only cares about the `M::T`/`M::M` types. What do you think about dropping the `GuestAddressSpace` bound, and simply stating that `T` derefs into a `GuestMemory` object. For example, the definition of `DescriptorChain` turns into:
> 
> ```rust
> pub struct DescriptorChain<T> {
>     mem: T,
>     desc_table: GuestAddress,
>     queue_size: u16,
>     ttl: u16,
>     index: u16,
>     desc: Descriptor,
> }
> ```
> 
> and the `impl` blocks become `impl<M: GuestMemory, T: Deref<Target=M>> DescriptorChain<T> { ... }`. Does this make sense as a future change?

There's a pending PR addressing this:)

src/queue.rs
```diff
@@ -63,20 +85,10 @@ pub struct DescriptorChain<M: GuestAddressSpace> {
...
+    index: u16,
 
-    /// Includes next, write, and indirect bits
-    pub flags: u16,
-
-    /// Index into the descriptor table of the next descriptor if flags has
-    /// the next bit set
-    pub next: u16,
+    /// This particular descriptor
+    pub desc: Descriptor,
```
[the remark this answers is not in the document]
bonzini: It doesn't need to be public.

bonzini:
> multiple generic structs (such as `DescriptorChain` itself) have type parameters like `M: GuestAddressSpace`. However, the associated logic only cares about the `M::T`/`M::M` types.

Having type parameters of type `GuestAddressSpace` is much easier to use in the case when you have to write down manually a generic type; it also hides implementation details and makes it easier to do refactorings that change the kind of `GuestAddressSpace` you're dealing with.

alexandruag:
> Having type parameters of type `GuestAddressSpace` is much easier to use in the case when you have to write down manually a generic type; it also hides implementation details and makes it easier to do refactorings that change the kind of `GuestAddressSpace` you're dealing with.

I think that using `GuestAddressSpace` as the trait bound will force ppl to be explicit about the type parameter a lot more often, because it's difficult to infer it automatically. For example, if the `DescriptorChain` impl had the bounds mentioned earlier instead, it wouldn't be necessary to explicitly specify the type when calling `checked_new`. Anyway, I guess this discussion should (maybe) continue on a future PR/issue :D

Going back to my previous comments, right now Descriptor and its fields (as well as `DescriptorChain::desc`) are part of the public interface of the crate, and I thought it's cleaner to either provide accessors for them, or drop the outside public visibility altogether. Would you rather keep things as they are? Does it make sense to do this in a future PR?

bonzini:
Yep, that can be fixed. I have pushed now.

bonzini:
> I think that using GuestAddressSpace as the trait bound will force ppl to be explicit about the type parameter a lot more often, because it's difficult to infer it automatically

I would expect the trait bound to be mostly inferred from struct fields. In the case of struct fields however you'd have to type it out fully.

PR #18, opened 2020-06-03
Enhancement to the virtio queue

jiangliu:
This patch set includes:
1) some minor bugfixes
2) port EVENT_IDX from cloud hypervisor
3) port readable/writable iterator from cloud hypervisor
4) port indirect descriptoer from cloud hypervisor
5) reduce unused dependency on epoll

This is a patch set split out from #13 , and #13 will only includes GuestMemory related changes.

### DIFF
--- a/src/queue.rs
+++ b/src/queue.rs
@@ -156,6 +156,22 @@ impl<M: GuestAddressSpace> DescriptorChain<M> {
     pub fn memory(&self) -> &M::M {
         &*self.mem
     }
+
+    /// Returns an iterator that only yields the readable descriptors in the chain.
+    pub fn readable(self) -> DescriptorChainRwIter<M> {
+        DescriptorChainRwIter {
+            chain: self,
+            writable: false,
+        }
+    }
+
+    /// Returns an iterator that only yields the writable descriptors in the chain.
+    pub fn writable(self) -> DescriptorChainRwIter<M> {
+        DescriptorChainRwIter {
+            chain: self,
+            writable: true,
+        }
+    }
 }
 
 impl<M: GuestAddressSpace> Iterator for DescriptorChain<M> {
@@ -185,6 +201,34 @@ impl<M: GuestAddressSpace> Iterator for DescriptorChain<M> {
             self.ttl -= 1;
         }
         Some(curr)
+    }
+}
+
+/// An iterator for readable or writable descriptors.
+pub struct DescriptorChainRwIter<M: GuestAddressSpace> {
+    chain: DescriptorChain<M>,
+    writable: bool,
+}
+
+impl<M: GuestAddressSpace> Iterator for DescriptorChainRwIter<M> {
+    type Item = Descriptor;
+
+    /// Returns the next descriptor in this descriptor chain, if there is one.
+    ///
+    /// Note that this is distinct from the next descriptor chain returned by
+    /// [`AvailIter`](struct.AvailIter.html), which is the head of the next
+    /// _available_ descriptor chain.
+    fn next(&mut self) -> Option<Self::Item> {
+        loop {
+            match self.chain.next() {
+                Some(v) => {
+                    if v.is_write_only() == self.writable {
+                        return Some(v);
+                    }
+                }
+                None => return None,
+            }
+        }
     }
 }
 
@@ -890,6 +934,69 @@ pub(crate) mod tests {
             c.next().unwrap();
             assert!(!c.has_next());
             assert!(c.next().is_none());
+        }
+    }
+
+    #[test]
+    fn test_descriptor_and_iterator() {
+        let m = &GuestMemoryMmap::from_ranges(&[(GuestAddress(0), 0x10000)]).unwrap();
+        let vq = VirtQueue::new(GuestAddress(0), m, 16);
+
+        let mut q = vq.create_queue(m);
+
+        // q is currently valid
+        assert!(q.is_valid());
+
+        for j in 0..7 {
+            vq.dtable(j).set(
+                0x1000 * (j + 1) as u64,
+                0x1000,
+                VIRTQ_DESC_F_NEXT,
+                (j + 1) as u16,
+            );
+        }
+
+        // the chains are (0, 1), (2, 3, 4) and (5, 6)
+        vq.dtable(1).flags().store(0);
+        vq.dtable(2)
+            .flags()
+            .store(VIRTQ_DESC_F_NEXT | VIRTQ_DESC_F_WRITE);
+        vq.dtable(4).flags().store(VIRTQ_DESC_F_WRITE);
+        vq.dtable(5)
+            .flags()
+            .store(VIRTQ_DESC_F_NEXT | VIRTQ_DESC_F_WRITE);
+        vq.dtable(6).flags().store(0);
+        vq.avail.ring(0).store(0);
+        vq.avail.ring(1).store(2);
+        vq.avail.ring(2).store(5);
+        vq.avail.idx().store(3);
+
+        let mut i = q.iter();
+
+        {
+            let c = i.next().unwrap();
+            let mut iter = c.into_iter();
+            assert!(iter.next().is_some());
+            assert!(iter.next().is_some());
+            assert!(iter.next().is_none());
+            assert!(iter.next().is_none());
+        }
+
+        {
+            let c = i.next().unwrap();
+            let mut iter = c.writable();
+            assert!(iter.next().is_some());
+            assert!(iter.next().is_some());
+            assert!(iter.next().is_none());
+            assert!(iter.next().is_none());
+        }
+
+        {
+            let c = i.next().unwrap();
+            let mut iter = c.readable();
+            assert!(iter.next().is_some());
+            assert!(iter.next().is_none());
+            assert!(iter.next().is_none());
         }
     }
 
--- a/src/queue.rs
+++ b/src/queue.rs
@@ -14,7 +14,7 @@ use std::cmp::min;
 use std::fmt::{self, Display};
 use std::mem::size_of;
 use std::num::Wrapping;
-use std::sync::atomic::{fence, Ordering};
+use std::sync::atomic::{fence, AtomicU16, Ordering};
 
 use vm_memory::{
     Address, ByteValued, Bytes, GuestAddress, GuestAddressSpace, GuestMemory, GuestUsize,
@@ -32,6 +32,8 @@ const VIRTQ_USED_RING_HEADER_SIZE: usize = 4;
 // The total size of the used ring is:
 // VIRTQ_USED_RING_HMETA_SIZE + VIRTQ_USED_ELEMENT_SIZE * queue_size
 const VIRTQ_USED_RING_META_SIZE: usize = VIRTQ_USED_RING_HEADER_SIZE + 2;
+// Used flags
+const VIRTQ_USED_F_NO_NOTIFY: u16 = 0x1;
 
 const VIRTQ_AVAIL_ELEMENT_SIZE: usize = 2;
 // Avail ring header: flags(u16) + idx(u16)
@@ -59,6 +61,8 @@ pub enum Error {
     InvalidIndirectDescriptor,
     /// Invalid descriptor chain.
     InvalidChain,
+    /// Invalid descriptor index.
+    InvalidDescriptorIndex,
 }
 
 impl Display for Error {
@@ -69,6 +73,7 @@ impl Display for Error {
             GuestMemoryError => write!(f, "error accessing guest memory"),
             InvalidChain => write!(f, "invalid descriptor chain"),
             InvalidIndirectDescriptor => write!(f, "invalid indirect descriptor"),
+            InvalidDescriptorIndex => write!(f, "invalid descriptor index"),
         }
     }
 }
@@ -430,6 +435,15 @@ pub struct Queue<M: GuestAddressSpace> {
     next_avail: Wrapping<u16>,
     next_used: Wrapping<u16>,
 
+    /// Notification from driver is enabled.
+    event_notification_enabled: bool,
+
+    /// VIRTIO_F_RING_EVENT_IDX negotiated
+    event_idx_enabled: bool,
+
+    /// The last used value when using EVENT_IDX
+    signalled_used: Option<Wrapping<u16>>,
+
     /// The queue size in elements the driver selected
     pub size: u16,
 
@@ -459,6 +473,9 @@ impl<M: GuestAddressSpace> Queue<M> {
             used_ring: GuestAddress(0),
             next_avail: Wrapping(0),
             next_used: Wrapping(0),
+            event_notification_enabled: true,
+            event_idx_enabled: false,
+            signalled_used: None,
         }
     }
 
@@ -482,6 +499,16 @@ impl<M: GuestAddressSpace> Queue<M> {
         self.used_ring = GuestAddress(0);
         self.next_avail = Wrapping(0);
         self.next_used = Wrapping(0);
+        self.signalled_used = None;
+        self.event_notification_enabled = false;
+        self.event_idx_enabled = false;
+    }
+
+    /// Enable/disable the VIRTIO_F_RING_EVENT_IDX feature.
+    pub fn set_event_idx(&mut self, enabled: bool) {
+        /* Also reset the last signalled event */
+        self.signalled_used = None;
+        self.event_idx_enabled = enabled;
     }
 
     /// Check if the virtio queue configuration is valid.
@@ -577,13 +604,13 @@ impl<M: GuestAddressSpace> Queue<M> {
     }
 
     /// Puts an available descriptor head into the used ring for use by the guest.
-    pub fn add_used(&mut self, desc_index: u16, len: u32) {
+    pub fn add_used(&mut self, desc_index: u16, len: u32) -> Result<u16, Error> {
         if desc_index >= self.actual_size() {
             error!(
                 "attempted to add out of bounds descriptor to used ring: {}",
                 desc_index
             );
-            return;
+            return Err(Error::InvalidDescriptorIndex);
         }
 
         let snapshot = self.mem.memory();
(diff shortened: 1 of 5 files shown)


src/queue.rs
```diff
@@ -320,6 +320,12 @@ impl<M: GuestAddressSpace> Queue<M> {
         min(self.size, self.max_size)
     }
 
+    /// Reset the queue to a state that is acceptable for a device reset
```
bonzini, line 408: Wrong commit message?

src/queue.rs
```diff
@@ -142,7 +174,8 @@ impl<M: GuestAddressSpace> DescriptorChain<M> {
         }
     }
 
-    fn checked_new(
+    /// Create a new DescriptorChain instance.
+    pub fn checked_new(
```
bonzini, line 173: Why does this need to be `pub`? Can it be `pub(crate)` instead?

src/queue.rs
```diff
@@ -151,6 +184,52 @@ impl<M: GuestAddressSpace> DescriptorChain<M> {
         Self::read_new(mem, dtable_addr, queue_size, queue_size, index)
     }
 
+    /// Create a `DescriptorChain` from the indirect target descriptor table.
+    pub fn new_from_indirect(&self) -> Result<DescriptorChain<M>, Error> {
```
bonzini, line 183: The commit message says `Queue::new_from_indirect`, so that should be fixed.

Would it be possible to add an `Option<DescriptorChain<M>>` field for the indirect descriptor, and add something like

```rust
    if let Some(ref mut indirect) = self.curr_indirect {
        return Some(indirect.next())
    } 
    if self.desc.is_indirect() {
        let indirect_chain = Self::new_from_indirect().ok()?;
        let first = indirect_chain.next();
        self.curr_indirect = Some(indirect_chain);
        return Some(first)
    }

    let curr = self.desc;
    // current code
```

? It can be done in a separate pull request too.

src/queue.rs
```diff
@@ -512,14 +512,63 @@ impl<M: GuestAddressSpace> Queue<M> {
...
+        let index_addr = match mem.checked_offset(self.avail_ring, 2) {
+            Some(ret) => ret,
+            None => {
+                // TODO log address
+                warn!("Invalid offset");
+                return;
+            }
+        };
+        // Note that last_index has no invalid values
+        let last_index: u16 = match mem.read_obj::<u16>(index_addr) {
```
bonzini, line 518: Please use the `integer-atomics` feature of vm-memory instead, and `get_atomic_ref::<u16>`. You can then use normal Rust atomic loads and stores.

src/queue.rs
```diff
@@ -512,14 +512,63 @@ impl<M: GuestAddressSpace> Queue<M> {
...
+            Ok(ret) => ret,
+            Err(_) => return,
+        };
+
+        match mem.checked_offset(self.used_ring, (4 + self.actual_size() * 8) as usize) {
+            Some(a) => {
+                mem.write_obj(last_index, a).unwrap();
```
bonzini: Same here. Note that `get_atomic_ref` takes care of range checking, so you can just use it instead of `checked_offset`.

src/queue.rs
```diff
@@ -512,14 +512,63 @@ impl<M: GuestAddressSpace> Queue<M> {
...
+            }
+            None => warn!("Can't update avail_event"),
+        }
+
+        // This fence ensures the guest sees the value we've just written.
+        fence(Ordering::Release);
```
bonzini, line 540: This fence is not needed. However, you need a SeqCst fence in the consumer, before it checks the `avail_idx`.

The way QEMU does it is to have a different API, `enable_notification(bool)`. The API does the same thing as your `update_avail_event` if `VIRTIO_RING_F_EVENT_IDX` is not available, but it could also takes care of setting or clearing `VRING_USED_F_NO_NOTIFY` (depending on the bool argument) if event indices are not in use.

With this API, the SeqCst fence is only needed if the bool argument is true.

src/queue.rs
```diff
@@ -145,18 +145,13 @@ impl<M: GuestAddressSpace> DescriptorChain<M> {
         }
 
         let desc_size = size_of::<Descriptor>();
-        let desc_head = match mem.checked_offset(desc_table, (index as usize) * desc_size) {
+        let desc_head = match desc_table.checked_add((index as u64) * (desc_size as u64)) {
```
bonzini, line 148: Whenever the code is introduced earlier in the series, can you squash it there?

src/queue.rs
```diff
@@ -112,12 +112,21 @@ impl<M: GuestAddressSpace> DescriptorChain<M> {
             return None;
         }
 
-        let desc_table_size = size_of::<Descriptor>() * queue_size as usize;
-        let slice = mem.get_slice(desc_table, desc_table_size).ok()?;
```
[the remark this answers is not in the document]
bonzini: I agree that we should have a testcase if there is a bug here.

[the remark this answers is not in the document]
jiangliu: The issue is that, when the whole descriptor table crosses region boundary, it's  legal and should work. With current implementation it fails because the get_slice() works for a single region.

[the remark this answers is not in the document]
bonzini: @sameo You're confusing `GuestRegion::checked_offset` with `GuestMemory::checked_offset`. But we need a testcase.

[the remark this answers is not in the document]
jiangliu: Add one test case to cover this case:)

src/queue.rs
jiangliu: Sorry, fixed:)

src/queue.rs
jiangliu: Done, please help to review the memory ordering:)

src/queue.rs
jiangliu: fixed

src/queue.rs
jiangliu: fixed

src/queue.rs
```diff
@@ -130,6 +130,7 @@ pub struct DescriptorChain<M: GuestAddressSpace> {
 
     /// The current descriptor
     desc: Descriptor,
+    curr_indirect: Option<Box<DescriptorChain<M>>>,
```
bonzini, line 133: Why does this need `Box`?

src/queue.rs
```diff
@@ -548,6 +563,74 @@ impl<M: GuestAddressSpace> Queue<M> {
...
+                    warn!("Can't update avail_event");
+                }
+            }
+            Err(e) => warn!("Invalid offset, {}", e),
+        }
+    }
+
+    /// Return the value present in the used_event field of the avail ring.
+    fn get_used_event(&self, mem: &M::M) -> Option<Wrapping<u16>> {
+        // Following code is safe because we have validated the queue. And the `used_index` is a
```
bonzini, line 594: The code is safe not because the queue was validated, but because all accesses are wrapped with `GuestMemory`.

src/queue.rs
```diff
@@ -548,6 +563,74 @@ impl<M: GuestAddressSpace> Queue<M> {
...
+        let used_event_addr = self
+            .avail_ring
+            .unchecked_add((4 + self.actual_size() * 2) as u64);
+        // This fence ensures we're seeing the latest update from the guest.
+        mem.get_slice(used_event_addr, size_of::<u16>())
+            .map(|s| {
+                Wrapping(
+                    s.get_atomic_ref::<AtomicU16>(0)
+                        .unwrap()
+                        .load(Ordering::SeqCst),
```
bonzini, line 606: Never use `SeqCst`, it's really hard to use and can lead to extremely complicated bugs. See [here](https://patchew.org/QEMU/20200407140746.8041-1-pbonzini@redhat.com/20200407140746.8041-6-pbonzini@redhat.com/) for an example.

`SeqCst` fences are okay though.

src/queue.rs
```diff
@@ -548,6 +563,74 @@ impl<M: GuestAddressSpace> Queue<M> {
...
+                } else {
+                    warn!("Can't update avail_event");
+                }
+            }
+            Err(e) => warn!("Invalid offset, {}", e),
+        }
+    }
+
+    /// Return the value present in the used_event field of the avail ring.
+    fn get_used_event(&self, mem: &M::M) -> Option<Wrapping<u16>> {
```
bonzini, line 593: The `mem` argument should not be necessary in rust-vmm/vm-virtio.

src/queue.rs
```diff
@@ -548,6 +563,74 @@ impl<M: GuestAddressSpace> Queue<M> {
...
+                    s.get_atomic_ref::<AtomicU16>(0)
+                        .unwrap()
+                        .load(Ordering::SeqCst),
+                )
+            })
+            .ok()
+    }
+
+    /// Check whether a notification to the guest is needed.
+    pub fn needs_notification(&mut self, mem: &M::M, used_idx: Wrapping<u16>) -> bool {
```
bonzini, line 613: The `mem` argument should not be necessary here, either.

src/queue.rs
jiangliu: Otherwise it will fail due to recursive type definition.
```
DescriptorChain {
    curr_indirect: Option<Box<DescriptorChain<M>>>,
}
```

bonzini: Doh, of course.

src/queue.rs
jiangliu: Thanks for the reference, it's very helpful. Seems that `Ordering::Relaxed` is safe for the EVENT_IDX related operations. 
Yeap, the `SeqCst` is really tricky, a long description in C11 stand:(

src/queue.rs
jiangliu: We have used several unchecked_add() below, which depends on queue validation:)

src/queue.rs
bonzini: I suggest you use an `enable_notification` method similar to QEMU's.

slp:
What's the status of this PR? I see @jiangliu addressed the comments from @bonzini, but looks like the commits haven't been updated. Are we missing a force push here?

alexandruag:
By no means should this delay/block anything, but just wanted to say I've started having a better look at the PR as well and I think it we be really nice if we can wrap up some of the ongoing discussions (I'm including other PRs/issues here also), and head towards something like a first version of the queue interfaces/implementations.

slp:
> By no means should this delay/block anything, but just wanted to say I've started having a better look at the PR as well and I think it we be really nice if we can wrap up some of the ongoing discussions (I'm including other PRs/issues here also), and head towards something like a first version of the queue interfaces/implementations.

That'd be great. I'm actually trying to put together a common framework for writing vhost-user devices so we can start packaging some implementations in a clean way to ease their way into some distributions. I think we're quite near there, but `vm-virtio::Queue` is missing some functionality that's either critical or very useful for some devices. Indirect descriptors and EVENT_IDX are probably the most important ones (and are covered by this PR). Support for `packed` queues would also be a nice thing to have.

jiangliu:
> What's the status of this PR? I see @jiangliu addressed the comments from @bonzini, but looks like the commits haven't been updated. Are we missing a force push here?

I have post the latest code to https://github.com/cloud-hypervisor/vm-virtio/tree/dragonball 
and will repost the latest version.
One ongoing discussion is how to deal with the GuestMemory object.

src/queue.rs
```diff
@@ -517,13 +532,13 @@ impl<M: GuestAddressSpace> Queue<M> {
...
     /// Puts an available descriptor head into the used ring for use by the guest.
-    pub fn add_used(&mut self, desc_index: u16, len: u32) {
+    pub fn add_used(&mut self, desc_index: u16, len: u32) -> Option<u16> {
         if desc_index >= self.actual_size() {
             error!(
                 "attempted to add out of bounds descriptor to used ring: {}",
                 desc_index
             );
-            return;
+            return None;
```
alexandruag, line 535: Returning an `Option` from this method doesn't feel quite right, since if we're ever in a position to return `None`, then we've encountered a serious error. I'm not sure what's the best way to handle various `Queue` error conditions (this look like another discussion to be had post this PR), but for now I think this method should either return a `Result<u16>`, or directly  an `u16` that's equal to `self.next_used.0 - 1` (the standard says that, with `VIRTIO_F_EVENT_IDX` negotiated, a notification must be sent when the `used_idx` we've just written to was equal to `used_event`).

src/queue.rs
```diff
@@ -548,6 +563,78 @@ impl<M: GuestAddressSpace> Queue<M> {
         snapshot
             .write_obj(self.next_used.0 as u16, used_ring.unchecked_add(2))
             .unwrap();
+
+        Some(self.next_used.0)
+    }
+
+    /// Update avail_event on the used ring with the last index in the avail ring.
+    pub fn update_avail_event(&mut self) {
```
alexandruag, line 561: I think this method should be split into an `avail_index()` that returns the available ring index, and then a `set_avail_event()` that writes a specified value to the `avail_event` location in the used ring.

src/queue.rs
```diff
@@ -548,6 +563,78 @@ impl<M: GuestAddressSpace> Queue<M> {
...
+        }) {
+            Ok(index) => {
+                let offset = (4 + self.actual_size() * 8) as u64;
+                let avail_event_addr = self.used_ring.unchecked_add(offset);
+                if let Ok(avail_event_slice) = mem.get_slice(avail_event_addr, size_of::<u16>()) {
+                    // This fence ensures the guest sees the value we've just written.
+                    avail_event_slice
+                        .get_atomic_ref::<AtomicU16>(0)
+                        .unwrap()
+                        .store(index, Ordering::Relaxed);
```
alexandruag, line 581: AFAIK, `Relaxed` does not generate an appropriate memory fence here. I think we should use `Ordering::Release`.

src/queue.rs
```diff
@@ -548,6 +563,78 @@ impl<M: GuestAddressSpace> Queue<M> {
...
+                } else {
+                    warn!("Can't update avail_event");
+                }
+            }
+            Err(e) => warn!("Invalid offset, {}", e),
+        }
+    }
+
+    /// Return the value present in the used_event field of the avail ring.
+    fn get_used_event(&self) -> Option<Wrapping<u16>> {
```
alexandruag, line 591: The usual naming convention is to skip the `get_` prefix for accessors (`set_` is still used tho for methods that write a value). Also, we should return a `Result` instead of an `Option`, because `None` is not a valid outcome during normal operation.

src/queue.rs
```diff
@@ -548,6 +563,78 @@ impl<M: GuestAddressSpace> Queue<M> {
...
+        let used_event_addr = self
+            .avail_ring
+            .unchecked_add((4 + self.actual_size() * 2) as u64);
+        // This fence ensures we're seeing the latest update from the guest.
+        mem.get_slice(used_event_addr, size_of::<u16>())
+            .map(|s| {
+                Wrapping(
+                    s.get_atomic_ref::<AtomicU16>(0)
+                        .unwrap()
+                        .load(Ordering::Relaxed),
```
alexandruag, line 606: I think we should use `Ordering::Acquire` here.

src/queue.rs
```diff
@@ -548,6 +563,78 @@ impl<M: GuestAddressSpace> Queue<M> {
...
+            .ok()
+    }
+
+    /// Check whether a notification to the guest is needed.
+    pub fn needs_notification(&mut self, used_idx: Wrapping<u16>) -> bool {
+        let mut notify = true;
+
+        // The VRING_AVAIL_F_NO_INTERRUPT flag isn't supported yet.
+        if self.event_idx {
+            if let Some(old_idx) = self.signalled_used.replace(used_idx) {
```
alexandruag, line 618: Hmm, why do we need to keep `self.signalled_used` around? From what I understand, whether we should send a notification or not depends entirely on the values of `used_idx` and `used_event` (the standard specifies a notification should be sent when essentially `used_idx == used_event + 1`, although some device implementations appear to send the notification when the more relaxed `used_idx > used_event` is true).

alexandruag:
Hi everyone! While there are still some comments to address and alternatives to discuss, it seems best to move forward with everything that's still pending on parallel threads. What do folks think about merging this PR as is, and then adding fixes and improvement proposals on top?

sameo:
> Hi everyone! While there are still some comments to address and alternatives to discuss, it seems best to move forward with everything that's still pending on parallel threads. What do folks think about merging this PR as is, and then adding fixes and improvement proposals on top?

Definitely agree, thanks @alexandruag

sameo (approved):

alexandruag:
@jiangliu @bonzini, does this sound good to you?

sameo (dismissed):

jiangliu:
> @jiangliu @bonzini, does this sound good to you?

There are some fixes for bugs found when integrating the vm-virtio with viritofs device driver.
So updated to the latest version from https://github.com/cloud-hypervisor/vm-virtio/tree/dragonball

alexandruag (dismissed):

src/queue.rs
```diff
@@ -614,6 +635,138 @@ impl<M: GuestAddressSpace> Queue<M> {
...
+        if self.event_notification_enabled {
+            if self.event_idx_enabled {
+                self.update_avail_event();
+            } else {
+                self.update_used_flag(0, VIRTQ_USED_F_NO_NOTIFY);
+            }
+
+            // This fence ensures that we observe the latest of virtq_avail once we publish
+            // virtq_used.avail_event/virtq_used.flags.
+            fence(Ordering::AcqRel);
```
bonzini, line 694: This needs to be SeqCst not AcqRel. AcqRel does not order previous stores against future loads.

src/queue.rs
bonzini: It's okay, the fence (which must be SeqCst) is in the caller.

src/queue.rs
bonzini: The reason for the more complicated expression is that it allows putting more than one descriptor in the ring and only do a final check for whether to notify. As far as I understand, `signalled_used` is needed to properly handle wrapping arithmetic.

src/queue.rs
bonzini: In fact it's not even enough: this function or its caller must have a SeqCst fence at the beginning so that the used ring is filled in before the used event is written.

Please add the fence together with a comment like

    // Complete all the writes in add_used() before reading the event.

bonzini (changes requested):
Still a little work to do on the fences.

bonzini:
@jiangliu since you need a few changes on the EVENT_IDX part please also rename `get_used_event` to `used_event` as suggested by Alex.

src/queue.rs
```diff
@@ -432,9 +622,149 @@ impl<M: GuestAddressSpace> Queue<M> {
...
         // We are guaranteed to be within the used ring, this write can't fail.
-        snapshot
-            .write_obj(self.next_used.0 as u16, used_ring.unchecked_add(2))
+        mem.write_obj(self.next_used.0, used_ring.unchecked_add(2))
             .unwrap();
+        if !self.event_idx_enabled {
+            // Ensure visibility of virtq_used.idx before sending notification to guest
+            // when the EVENT_IDX feature is disabled, otherwise used_event() will ensure
+            // visibility of virtq_used.idx.
+            fence(Ordering::Release);
```
bonzini, line 631: This is not needed; please remove. Sending a notification (interrupt) is itself a "release" operation and is ordered after previous stores. Just like `used_event()` contains a fence, whatever sends the interrupt (for example a KVM ioctl) will take care of it.

The code is correct without this fence. Note that

```rust
    fence(Ordering::Release);
    fence(Ordering::Acquire);
```

is not the same as `fence(Ordering::SeqCst);`.

bonzini (changes requested):
The new fence that was introduced is not necessary.

bonzini:
Latest version is okay once CI is green.

src/queue.rs
jiangliu: Thanks for explanation, removed.

bonzini (approved):

slp (approved):

alexandruag (approved):
Thanks everyone! Looks like we can merge this now. I'll write some simple performance tests to help us better understand the impact of various changes, and then create PRs with a couple of proposals. Going forward, I think it's also useful to start an issue to better understand the fence/atomic/sync aspects and decisions we're making as part of this implementation.

PR #19, opened 2020-09-13
Simple initial benchmark and descriptor handling simplifications

alexandruag:
Hi all! This PR proposes two additions. The first is a `criterion`-based benchmark that mainly exercises descriptor chain iteration (for now :D). That commit is also adding a bunch of functionality that mocks and helps populate `Queue` contents. I've later realized we also had something like that as part of the unit test functionality; going forward, it would be great to unify the auxiliary logic. The benchmark can be started using something like `cargo bench --bench main`.

The second commit proposes some simplifications related to descriptor chain handling, mostly by unifying the handling of descriptors from the initial and indirect tables. Other changes include:
- `DescriptorChain::has_next` has been removed, as its semantics were a bit unclear and didn't seem very useful (it can return `true`, but `next()` may still result in `None` due to other errors).
- `DescriptorChain::is_indirect` now returns true only after an indirect descriptor has been effectively parsed (via `next()`), and the chain starts using the indirect table. Do ppl think this method is actually useful? (in any of its incarnations)
- I removed the `DescriptorChain::is_valid` method. The `self.desc.next < self.queue_size` check is already done in other parts of the code, while the  `checked_offset` validation is not very useful since it only validates the address of the byte past the descriptor, but not the range in-between. The `GuestMemory` implementation will validate accesses anyway before performing them.

Looking forward to any reviews/comments!

### DIFF
--- a/benches/queue/mock.rs
+++ b/benches/queue/mock.rs
@@ -1,0 +1,270 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
+
+use std::marker::PhantomData;
+use std::mem::size_of;
+
+use vm_memory::{Address, ByteValued, Bytes, GuestAddress, GuestAddressSpace, GuestMemory};
+
+use vm_virtio::Queue;
+
+struct Ref<'a, M, T> {
+    mem: &'a M,
+    addr: GuestAddress,
+    phantom: PhantomData<*const T>,
+}
+
+impl<'a, M: GuestMemory, T: ByteValued> Ref<'a, M, T> {
+    fn new(mem: &'a M, addr: GuestAddress) -> Self {
+        Ref {
+            mem,
+            addr,
+            phantom: PhantomData,
+        }
+    }
+
+    fn load(&self) -> T {
+        self.mem.read_obj(self.addr).unwrap()
+    }
+
+    fn store(&self, val: T) {
+        self.mem.write_obj(val, self.addr).unwrap()
+    }
+}
+
+struct ArrayRef<'a, M, T> {
+    mem: &'a M,
+    addr: GuestAddress,
+    len: usize,
+    phantom: PhantomData<*const T>,
+}
+
+impl<'a, M: GuestMemory, T: ByteValued> ArrayRef<'a, M, T> {
+    fn new(mem: &'a M, addr: GuestAddress, len: usize) -> Self {
+        ArrayRef {
+            mem,
+            addr,
+            len,
+            phantom: PhantomData,
+        }
+    }
+
+    fn ref_at(&self, index: usize) -> Ref<'a, M, T> {
+        // TODO: add better error handling to the mock logic.
+        assert!(index < self.len);
+
+        let addr = self
+            .addr
+            .checked_add((index * size_of::<T>()) as u64)
+            .unwrap();
+        Ref::new(self.mem, addr)
+    }
+}
+
+#[repr(C)]
+#[derive(Clone, Copy, Default)]
+pub struct Descriptor {
+    pub addr: u64,
+    pub len: u32,
+    pub flags: u16,
+    pub next: u16,
+}
+
+unsafe impl ByteValued for Descriptor {}
+
+#[repr(C)]
+#[derive(Clone, Copy, Default)]
+pub struct UsedRingElement {
+    pub id: u32,
+    pub len: u32,
+}
+
+unsafe impl ByteValued for UsedRingElement {}
+
+pub struct SplitQueueRing<'a, M, T: ByteValued> {
+    flags: Ref<'a, M, u16>,
+    idx: Ref<'a, M, u16>,
+    ring: ArrayRef<'a, M, T>,
+    // `used_event` for `AvailRing`, `avail_event` for `UsedRing`.
+    event: Ref<'a, M, u16>,
+}
+
+impl<'a, M: GuestMemory, T: ByteValued> SplitQueueRing<'a, M, T> {
+    pub fn new(mem: &'a M, base: GuestAddress, len: u16) -> Self {
+        let event_addr = base
+            .checked_add(4)
+            .and_then(|a| a.checked_add((size_of::<u16>() * len as usize) as u64))
+            .unwrap();
+
+        let split_queue_ring = SplitQueueRing {
+            flags: Ref::new(mem, base),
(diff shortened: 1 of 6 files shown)


src/queue.rs
```diff
@@ -131,111 +148,38 @@ pub struct DescriptorChain<M: GuestAddressSpace> {
...
-    /// Create a new DescriptorChain instance.
-    fn checked_new(
-        mem: M::T,
-        dtable_addr: GuestAddress,
-        queue_size: u16,
-        index: u16,
-    ) -> Option<Self> {
-        Self::read_new(mem, dtable_addr, queue_size, queue_size, index)
+    /// Create a new `DescriptorChain` instance.
+    fn new(mem: M::T, desc_table: GuestAddress, queue_size: u16, index: u16) -> Self {
```
slp, line 175: Slightly unrelated, but it'd be nice to have a public function for building a `DescriptorChain`, so other crates can make use of it for their own unit tests.

slp (commented):
The simplification part looks good to me. I've tested this implementation with `vhost-user-fs`, which makes use of indirect descriptors, and haven't found any issues with it (we're still missing a couple things, but that's another story).

As for the benchmark part, I'm not a big fan of microbenchmarks when the worksheet is so small, because you can frequently get a very different scenario once you add some actual work on each iteration. That said, I'm not opposed to having them here either.

src/queue.rs
alexandruag: I think we should keep auxiliary logic separate from the "main" functionality, so the latter is smaller and thus easier to review, test, and audit. We can add methods for building descriptor chains and other useful things in a separate module (made opt-in via a feature) that can be then consumed by our tests and potentially other crates as well. For example, we used this approach in [event-manager](https://github.com/rust-vmm/event-manager) where the `test_utilities` feature gates the `src/utilities` module. How does that sound?

alexandruag (commented):
Thanks for the review! I totally acknowledge that microbenchmarks are not necessarily representative of real-world performance, but at the same time they provide useful data points and help identify potential issues early. We should build more realistic integration harnesses as well when additional components become available.

src/queue.rs
andreeaflorescu: Just wanted to call out that the terrible thing about the test_utilities in event-manager is that `cargo test` no longer works. You need to do either `cargo test --all-features` or `cargo test --features test_utilities`.

Otherwise I agree with what Alex said :))

slp: Makes sense to me. It's cleaner and I don't see any significant drawbacks. It most situations you already need to pass the `--all-features` flag anyways.

slp:
> Thanks for the review! I totally acknowledge that microbenchmarks are not necessarily representative of real-world performance, but at the same time they provide useful data points and help identify potential issues early. We should build more realistic integration harnesses as well when additional components become available.

In an ideal world, we should have a test matrix with multiple VMMs and `vhost-user` device implementations (both in Rust and in C), to be able to get meaningful, comparable numbers. But that would require dedicated bare-metal hardware, which I guess would imply the need for some kind of sponsorship.

But that's definitely outside the scope of this PR ;-) I'll be in favor of transitioning this one from RFC to formal patch once https://github.com/rust-vmm/vm-virtio/pull/21 is merged and this is rebased on it.

alexandruag:
Actually, the CI runs on bare-metal hardware (for now at least) :D

alexandruag:
Hi! I rebased this on top of #21 and dropped the RFC status. Would be great if ppl can take another look. The separation and availability of useful auxiliary we previously talked about is not included yet, and I created #23 for tracking.

Also, I was wondering, do ppl think the `DescriptorChain::is_indirect` method is useful? I would remove it otherwise, just to try and keep things as simple as possible.

slp:
> Also, I was wondering, do ppl think the `DescriptorChain::is_indirect` method is useful? I would remove it otherwise, just to try and keep things as simple as possible.

I've checked the crates I'm familiar with, and the only one I found using it is `vhost-user-fs`, but that's because it's using the old `DescriptorChain::new_from_indirect` method. Once it's migrated to `rust-vmm/vm-virtio`, it no longer needs that.

So I'm fine with dropping it. We can always add it later if necessary.

alexandruag:
Cool! Removed `DescriptorChain::is_indirect` for now.

benches/queue/mock.rs
```diff
@@ -0,0 +1,271 @@
...
+    avail_addr: GuestAddress,
+    avail: AvailRing<'a, M>,
+    used_addr: GuestAddress,
+    _used: UsedRing<'a, M>,
+    indirect_addr: GuestAddress,
+}
+
+// Move to a defs module in vm-virtio?
+const VIRTQ_DESC_F_NEXT: u16 = 0x1;
+// const VIRTQ_DESC_F_WRITE: u16 = 0x2;
```
aghecenco, line 192: nit: remove redundant comment

src/queue.rs
```diff
@@ -116,6 +116,23 @@ impl Descriptor {
...
+
+    /// Check whether this is an indirect descriptor.
+    pub fn is_indirect(&self) -> bool {
+        // TODO: The are a couple of restrictions in terms of which flags combinations are
+        // actually valid for indirect descriptors. Implement those checks as well somewhere.
+        self.flags() & VIRTQ_DESC_F_INDIRECT != 0
+    }
+
+    /// Check whether the `VIRTQ_DESC_F_NEXT` is set for the descriptor.
+    pub fn has_next(&self) -> bool {
```
aghecenco, line 132: The PR description says these were removed.

src/queue.rs
```diff
@@ -268,6 +205,31 @@ impl<M: GuestAddressSpace> DescriptorChain<M> {
...
+            return Err(Error::InvalidIndirectDescriptor);
+        }
+
+        let table_len = (desc.len as usize) / VIRTQ_DESCRIPTOR_SIZE;
+        // Check the target indirect descriptor table is correctly aligned.
+        if desc.addr().raw_value() & (VIRTQ_DESCRIPTOR_SIZE as u64 - 1) != 0
+            || (desc.len as usize) & (VIRTQ_DESCRIPTOR_SIZE - 1) != 0
+            || table_len > usize::from(std::u16::MAX)
+        {
+            return Err(Error::InvalidIndirectDescriptor);
```
aghecenco, line 222: I would differentiate between this case and the one a few lines higher with a separate `Error`.

slp (approved):

aghecenco (approved):

PR #21, opened 2020-09-14
Implement Queue::index() and Queue::has_next()

### LINKED ISSUE #20: Where to obtain the descriptor index for Queue::add_used
After commit 904662c3830b6df303d1c15108d96df395011b6c (@bonzini) removed the field `index` from `DescriptorChain`, I can't find a way to obtain the index from the last processed descriptor to pass it to `Queue::add_used`. Am I missing something or do we need that field back?

jiangliu:
Implement Queue::index() to fix #20 and fix a bug in getting the has_next flag.

### DIFF
--- a/src/queue.rs
+++ b/src/queue.rs
@@ -84,6 +84,7 @@ impl std::error::Error for Error {}
 #[repr(C)]
 #[derive(Default, Clone, Copy)]
 pub struct Descriptor {
+    /// Guest physical address of device specific data
     addr: u64,
 
     /// Length of device specific data
@@ -131,7 +132,8 @@ pub struct DescriptorChain<M: GuestAddressSpace> {
     mem: M::T,
     desc_table: GuestAddress,
     queue_size: u16,
-    ttl: u16, // used to prevent infinite chain cycles
+    ttl: u16,        // used to prevent infinite chain cycles
+    head_index: u16, // descriptor index of the chain header
 
     /// The current descriptor
     desc: Descriptor,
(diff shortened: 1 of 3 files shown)


src/queue.rs
```diff
@@ -131,7 +132,8 @@ pub struct DescriptorChain<M: GuestAddressSpace> {
     mem: M::T,
     desc_table: GuestAddress,
     queue_size: u16,
-    ttl: u16, // used to prevent infinite chain cycles
+    ttl: u16,   // used to prevent infinite chain cycles
+    index: u16, // descriptor index of the chain header
```
alexandruag, line 141: Looks a bit clearer if we rename this and the accessor method to something like `head_index`.

src/queue.rs
```diff
@@ -239,7 +247,7 @@ impl<M: GuestAddressSpace> DescriptorChain<M> {
 
     /// Checks if this descriptor chain has another descriptor chain linked after it.
     pub fn has_next(&self) -> bool {
-        self.desc.flags & VIRTQ_DESC_F_NEXT != 0 && self.ttl > 1
+        self.has_next || self.curr_indirect.is_some()
```
alexandruag, line 250: Is the `DescriptorChain::has_next` method really useful? It doesn't seem to guarantee that the following `next()` call will return something, and we can already look at the `Descriptor` object to determine if the chain is supposed to continue. I proposed removing it in #19 (would be great if you can have a look btw :D), and was wondering if there's a good reason to keep it around.

slp: I'd say the common use case for `DescriptorChain` is to simply iterate over it, so `DescriptorChain::has_next` may be redundant or even confusing, as it doesn't guarantee that `next()` will return something, as you said.

That said, if there's a use case for it (perhaps as a hint for an optimization?), I think we can keep it as long it's clearly stated in the documentation it's return value should only be used as a hint.

alexandruag: I've started favoring the approach where only things known to be necessary get included, as new features can always be added later when they become relevant. I think this is one of those approaches that really pay off in the long run. At the same time, I'm by no means familiar with all the current use cases, so we can totally keep this if folks think it's worth it.

alexandruag:
This looks good overall! @jiangliu, what are your thoughts on the details mentioned above?

alexandruag (dismissed):

slp (approved):

andreeaflorescu (dismissed):

andreeaflorescu (approved):

alexandruag (approved):

PR #26, opened 2020-10-06
Switch to latest vm-memory and tweak the notification logic

alexandruag:
This PR switches to using version `0.3.0` of `vm-memory` which adds atomic `load` and `store` support as part of the `Bytes`/`GuestMemory` interface. There are also some proposed changes around the notification logic (for example, double checking when enabling notifications to make sure we don't miss one), as well as more comments just in case they are helpful to people reading the code (there's one example around enabling notifications that should be turned into a doc comment/example for consumers of the crate, because it's not  immediately clear what the use pattern looks like).

Also, I was wondering if it makes sense to remove `DescriptorChainRwIter` and the related logic, because it looks like we can achieve the same by calling `filter()` on `DescriptorChain` when iterating. WDYT?

### DIFF
--- a/src/queue.rs
+++ b/src/queue.rs
@@ -377,9 +377,6 @@ pub struct Queue<M: GuestAddressSpace> {
     next_avail: Wrapping<u16>,
     next_used: Wrapping<u16>,
 
-    /// Notification from driver is enabled.
-    event_notification_enabled: bool,
-
     /// VIRTIO_F_RING_EVENT_IDX negotiated
     event_idx_enabled: bool,
 
@@ -415,7 +412,6 @@ impl<M: GuestAddressSpace> Queue<M> {
             used_ring: GuestAddress(0),
             next_avail: Wrapping(0),
             next_used: Wrapping(0),
-            event_notification_enabled: true,
             event_idx_enabled: false,
             signalled_used: None,
         }
@@ -442,7 +438,6 @@ impl<M: GuestAddressSpace> Queue<M> {
         self.next_avail = Wrapping(0);
         self.next_used = Wrapping(0);
         self.signalled_used = None;
-        self.event_notification_enabled = false;
         self.event_idx_enabled = false;
     }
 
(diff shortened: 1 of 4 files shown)


slp:
I've just taken a first look and I like it. I'll look at it in depth tomorrow, and will also play a bit with a couple of vhost-user devices consuming this crate.

bonzini (dismissed):
Looks good to me, though I have not done a careful review beyond the orderings and fences.

I like having Ordering arguments even though they should always be Relaxed , because it tells you "this is using atomic accesses, look for fences" (and likewise for Release "do not look for fences". I had never seen this idiom but it's very self-documenting!

src/queue.rs
```diff
@@ -249,12 +243,11 @@ impl<M: GuestAddressSpace> Iterator for DescriptorChain<M> {
...
 
-        let desc_table_size = VIRTQ_DESCRIPTOR_SIZE * self.queue_size as usize;
-        let slice = self.mem.get_slice(self.desc_table, desc_table_size).ok()?;
-        let desc = slice
-            .get_array_ref::<Descriptor>(0, self.queue_size as usize)
-            .ok()?
-            .load(self.next_index as usize);
+        let desc_addr = self
+            .desc_table
+            .unchecked_add(self.next_index as u64 * size_of::<Descriptor>() as u64);
```
andreeaflorescu, line 248: Why is this `unchecked_add` safe?

src/queue.rs
```diff
@@ -307,58 +300,63 @@ pub struct AvailIter<'b, M: GuestAddressSpace> {
...
             return None;
         }
 
-        let offset = (VIRTQ_AVAIL_RING_HEADER_SIZE as u16
-            + (self.next_index.0 % self.queue_size) * VIRTQ_AVAIL_ELEMENT_SIZE as u16)
-            as usize;
-        let avail_addr = self.avail_ring.checked_add(offset as u64)?;
-        // This index is checked below in checked_new
-        let desc_index: u16 = self
+        // This computation cannot overflow.
```
andreeaflorescu, line 316: Mind to also say why it cannot overflow? :)) Otherwise, we can just remove the comment.

src/queue.rs
```diff
@@ -307,58 +300,63 @@ pub struct AvailIter<'b, M: GuestAddressSpace> {
...
+        let offset = VIRTQ_AVAIL_RING_HEADER_SIZE
+            + (self.next_avail.0 % self.queue_size) as u64 * VIRTQ_AVAIL_ELEMENT_SIZE;
+
+        // The logic in `Queue::is_valid` ensures it's ok to use `unchecked_add` as long
```
andreeaflorescu, line 320: Hmm, this assumption has the potential to introduce time of check, time of use bugs. Are we making sure in anyway that next is only called after `queue::is_valid`?

andreeaflorescu (commented):
Just some questions. Didn't fully review it.

slp (dismissed):
Took a second look and still LGTM (apart from the issues pointed by @andreeaflorescu). I've tested the notification logic with `virtiofsd-rs` and interrupt suppression works just as expected. The API changes are reasonable. We should probably wait for @jiangliu 's feedback before merging this, though.

On the other hand, we should probably think about a versioning policy for API changes.

src/queue.rs
alexandruag: Added a comment.

src/queue.rs
alexandruag: Added a bit more details to the comment.

src/queue.rs
alexandruag: We do not currently enforce that a queue is only used after checking `is_valid`, but rather expect the device implementations to do that before activation. The standard also forbids drivers to change queue parameters while the device is "running". A warp-around cannot lead to unsafe memory accesses because the memory model will check everything. We can add more validations if that seems like a worthwhile improvement, but I'm not sure if the best place is here or as part of logic that mutates the queue configuration.

alexandruag (commented):
Thanks for the reviews! Having a policy around API (breaking) changes is definitely something we should have in mind, but I think many components will still go through some churn until reaching an initial stable state. Any ideas/guidelines to use as a starting point for that discussion would be great though.

@jiangliu, can you please have a look as well?

src/queue.rs
andreeaflorescu: The explanation from your comment should also be found in the code IMHO.

slp:
@jiangliu I'd like to merge this one to be able to update `vhost-user-backend` to use `vm-memory` 3.x. Could you please take a look?

src/queue.rs
```diff
@@ -660,46 +649,52 @@ impl<M: GuestAddressSpace> Queue<M> {
...
-                        notify = false;
-                    }
+            if let Some(prev) = self.signalled_used {
+                let used_event = self.used_event(Ordering::Relaxed)?;
+                // This check looks at `used_idx`, `used_event`, and `prev` as if they are on an
+                // axis (that wraps around). If the difference between `used_idx` and `used_used`
+                // is greater than or equal to the difference between `used_idx` and `prev`, then
+                // `prev` is closer to `used_idx` than `used_event` (and thus more recent), so
+                // we don't need to elicit another notification.
+                if used_idx - used_event >= used_idx - prev {
```
jiangliu, line 686: Why the "- Wrapping(1u16)" has been removed?

jiangliu:
> @jiangliu I'd like to merge this one to be able to update `vhost-user-backend` to use `vm-memory` 3.x. Could you please take a look?

Sorry for reply, it took some time to figure out the indirect related question:(
It looks good to me except the "- Wrapping(1u16)" doubt mentioned above.

src/queue.rs
alexandruag: It hasn't been removed entirely, it's now part of the computation at the beginning of the function (line 672) which computes the last index in the used ring where we wrote something.

jiangliu: Seems it still has changed the logic because the `prev` is self.next_used - Wrapping(1) of last round?

src/queue.rs
```diff
@@ -1322,37 +1308,43 @@ pub(crate) mod tests {
...
-        assert_eq!(q.needs_notification(Wrapping(4)), false);
-        assert_eq!(q.needs_notification(Wrapping(5)), true);
-        assert_eq!(q.needs_notification(Wrapping(6)), false);
-        assert_eq!(q.needs_notification(Wrapping(7)), false);
+
+        for i in 0..12 {
+            q.next_used = Wrapping(i);
+            let expected = i == 5 || q.signalled_used.is_none();
+            assert_eq!(q.needs_notification().unwrap(), expected);
+        }
```
jiangliu, line 1330: The spec states:
_For example, if used_event is 0, then a device using VIRTIO_F_EVENT_IDX would send a used buffer notification to the driver after the first buffer is used (and again after the 65536th buffer, etc)._

So how about adding a case for "again after the 65536th buffer"?

src/queue.rs
alexandruag: Oh, I see your point now. So essentially, the left-hand side of the inequality is the same as before (because `used_event` is the same, and `used_idx` is now defined as `self.next_used - Wrapping(1)`), but the right-hand side is different because `used_idx` is now the old `used_idx` with a `- Wrapping(1)` as well. After a bit of doodling I think you are right and the change is erroneous. We should use a `>` instead of `>=` to account for the implicit `- Wrapping(1)` so things are exactly as before.

This also has the added advantage that `needs_notification` will return true until `used_idx` actually changes, and I think we can use a value that defaults to 0 for `signalled_used` instead of an `Option`. I'll experiment a bit more with the changes and update the PR later today or tomorrow at the latest. Thanks for pointing this out!

src/queue.rs
```diff
@@ -1322,37 +1308,43 @@ pub(crate) mod tests {
...
 
         m.write_obj::<u16>(8, avail_addr.unchecked_add(4 + 16 * 2))
             .unwrap();
-        assert_eq!(q.needs_notification(Wrapping(11)), true);
-        assert_eq!(q.needs_notification(Wrapping(12)), false);
+
+        assert_eq!(q.needs_notification().unwrap(), true);
```
jiangliu, line 1335: A little hesitating whether it should be "true" or "false" here:)

src/queue.rs
jiangliu: Be careful about the ">=" vs ">" and wrapping around issue:)
All implementation of algorithm I have seen using the form "if (used_idx - used_event - Wrapping(1u16)) >= (used_idx - old_idx)".

src/queue.rs
alexandruag: added

src/queue.rs
alexandruag: Ok so after more playing around, the sad conclusion is there's no nice way of making `needs_notification` idempotent. I think a cleaner solution is something like making `add_used` return a boolean value that indicates whether a notification is needed or not. For the time being, I've changed the code here to be essentially the same as before the  PR, for ease of re-reviewing this part.

Also, turns out the regression caused by the original version of the PR was not around the handling of `- Wrapping(1)` (which actually worked out in the end), but the fact that `signalled_used` was only updated when the method would return `true` (which fit the name, but changed the logic).

src/queue.rs
alexandruag: I think it should be `false` (and it's what the method actually returns now after changing `needs_notification` back).

alexandruag (commented):
Hi! I've addressed all outstanding comments. In particular, the `needs_notification` method now effectively implements the same checks as before. We can consider other options after this PR is merged. I'll create an issue about potential future queue refinements, where we can also discuss the aspect of breaking changes to the interface going forward.

jiangliu (approved):

alexandruag:
@slp can you please have another look? The delta between the last update (following Gerry's comments) and the previous version of the PR is available [here](https://github.com/rust-vmm/vm-virtio/compare/eab0e9b32bbd3899ac53cb4ccf0096c34eebde94..4083ec5488f0aab0fbd0461a92904145f60b6ad5).

slp (approved):
LGTM.

PR #27, opened 2020-10-26
[RFC] Another queue/mem split proposal

alexandruag:
Hi! We've discussed before about removing the `M: GuestAddressSpace` member field from `Queue`, and IIRC the proposed approach was to simply pass a `mem: &GuestMemory` extra parameter to all queue methods which need to access the memory of the guest. This quick RFC PR proposes an alternative where the current configuration of the queue gets separated from the live object used to interact with the rings. The changes are added on top of #26, so please only take a look at the most recent commit.

There are now two objects: `QueueConfig` (actual name TBD) which keeps all members previously owned by `Queue` (sans the memory object), and `Queue` which is ephemeral and only contains a reference to a `M: GuestMemory` and a mutable reference to a `QueueConfig` object. `QueueConfig` is persistent, while `Queue` is only meant to be instantiated on demand when doing queue operations. A nice thing is that we can use a common config object for both split and packed queues when we starting working in that direction.

Looking forward to any feedback. If the approach makes sense, I'll fix the tests and add some more documentation as well.

### DIFF
--- a/src/queue.rs
+++ b/src/queue.rs
@@ -373,9 +373,6 @@ pub struct Queue<M: GuestAddressSpace> {
     next_avail: Wrapping<u16>,
     next_used: Wrapping<u16>,
 
-    /// Notification from driver is enabled.
-    event_notification_enabled: bool,
-
     /// VIRTIO_F_RING_EVENT_IDX negotiated
     event_idx_enabled: bool,
 
@@ -411,7 +408,6 @@ impl<M: GuestAddressSpace> Queue<M> {
             used_ring: GuestAddress(0),
             next_avail: Wrapping(0),
             next_used: Wrapping(0),
-            event_notification_enabled: true,
             event_idx_enabled: false,
             signalled_used: None,
         }
@@ -438,7 +434,6 @@ impl<M: GuestAddressSpace> Queue<M> {
         self.next_avail = Wrapping(0);
         self.next_used = Wrapping(0);
         self.signalled_used = None;
-        self.event_notification_enabled = false;
         self.event_idx_enabled = false;
     }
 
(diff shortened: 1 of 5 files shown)


bonzini:
No problem splitting `QueueConfig` from `Queue`, in fact I would split it further by dropping the `Deref` and `DerefMut` implementations that go from `Queue` to `QueueConfig`. The APIs of the two should generally not overlap, except perhaps for `go_to_previous_position` which you can add to `Queue` and make private in `QueueConfig`.

In addition, Queue should support using smart pointers for both the memory and the config fields. The memory should be accessed as a `GuestAddressSpace::T` rather than a `&'a GuestMemory`, and instead of having a mutable reference to QueueConfig you could have a `DerefMut<Target = QueueConfig>`. This would allow protecting the configuration with a `Mutex` or `RefCell`. That would be something like this:

```rust
pub struct Queue<M: GuestAddressSpace, C: DerefMut<Target = QueueConfig>>  {
    mem: M::T,
    config: C,
}
```

alexandruag:
Hi Paolo, thanks for the comments! The `Deref*` implementations are mostly for the convenience of not having to do `queue.config.op`, since all those operations also make sense on a live object. Using smart pointers is neat and I think it also unlocks some additional use cases. However, I think the memory one can remain something like `Deref<Target=GuestMemory>`, instead of being aware of `GuestAddressSpace` as well. I'll experiment some more and try to refine the proposal.

bonzini:
> convenience of not having to do queue.config.op

I understand the convenience, but I'm wondering if it's not a violation of the Queue vs QueueConfig separation. Is there a case where you have a Queue but not a QueueConfig, and you have a legitimate need to access something in QueueConfig? Lacking a case where it's useful, I'd rather avoid the Deref* implementations.

bonzini:
Thinking more about the smart pointers too:

> However, I think the memory one can remain something like `Deref<Target=GuestMemory>`

It would have to be `Clone + Deref<Target=GuestMemory>`. But that's more easily expressed as `GuestAddressSpace::T`, I think.

Also my proposal for Config is wrong, as it should be `BorrowMut<QueueConfig>` rather than `DerefMut<Target=QueueConfig>`. I'd just leave aside that part.

src/queue.rs
```diff
@@ -334,7 +333,7 @@ impl<'b, M: GuestAddressSpace> Iterator for AvailIter<'b, M> {
         *self.next_avail += Wrapping(1);
 
         Some(DescriptorChain::new(
-            self.mem.clone(),
+            self.mem,
```
bonzini, line 336: Removing this wouldn't work with the hotplug-friendly GuestAddressSpace implementations in vm-memory.

alexandruag: Specifically for the example in this commit, it works based on the memory snapshot being created and held somewhere outside for the lifetime of the `Queue` object (that only gets the resulting `&GuestMemory` reference), and it looks like the overall idea remains the same even if we turn `M` into a smart pointer. I'll think a bit more about all these things (and a couple of others encountered while experimenting with queues) and update the proposal.

bonzini: The lifetime of the Queue is one of the following: 1) from one configuration change (including memory map changes) to the next or 2) from one ring processing until when the ring is fully processed.

In either case, the request (and thus the DescriptorChain) can outlive the Queue, and therefore the clone is necessary.

bonzini: (You can use `Arc<Queue>` instead; in that case you can use `&GuestMemory` as the argument. But it seems wrong to me when the actual thing that has an unknown lifetime is the memory map).

alexandruag: Yeah, so the nuance the a `DescriptorChain` can be "stored" and outlive the `Queue` was not part of my original thinking. Will keep that in mind.

slp: To give this some context, having `DescriptorChain`s not tied to their `Queue`'s lifetime is really useful for things like servicing requests from a thread pool.

alexandruag:
Gave this a bit more thought, and it seems there are a number of things we should take into consideration going forward (not necessarily related to the original theme of the PR :D).

- It looks like having a `QueueConfig` object makes sense, especially since the configuration/state definition is largely the same for both split and packed queues (and we can switch to the queue-neutral terminology, i.e. `DriverArea` vs `AvailableRing`, etc).
- Having different underlying implementation details might actually make sense for different use-cases, so we can define some interfaces for queues as well.
- Kinda immediately related to the previous points is the question about how to approach adding support for packed queues as well. I'm currently looking at abstracting this behind a `Queue` interface. Do you think it makes sense to attempt a unifying approach?
- Last but not least, I think we might run into the issue of logging/metrics sooner rather than later. Right now, `vm-virtio` has a `log` dependency and uses the associated macros ("warn!", "error!", etc) in different places. Going forward, this might not be the right solution for all consumers of the crate, as ppl might prefer a different log crate, or different levels/messages, the ability to bump metrics, and other things.

I'll create one or more issues for the above and later close the PR, but just wanted to mention them here for now. Also thanks for the comments and examples!

bonzini:
> Kinda immediately related to the previous points is the question about how to approach adding support for packed queues as well. I'm currently looking at abstracting this behind a Queue interface. Do you think it makes sense to attempt a unifying approach?

I would just use `if`s. Quite ugly, but it's unlikely to have a third format in the future; and this is a hot path so using dynamic trait objects would be too slow.

> Going forward, this might not be the right solution for all consumers of the crate, as ppl might prefer a different log crate, or different levels/messages, the ability to bump metrics, and other things.

Metrics and tracing are a complicated topic... Is there anything in the wider Rust ecosystem that provides them with enough performance and flexibility?

src/queue.rs
jiangliu: Exactly, we ran into the same requirement when trying to adding multi-threading for virtio-fs backends.
It would be great to support Queue outlives the virtio event handler.

alexandruag:
The traits were not meant for dynamic dispatch (they would prob end up having associated types anyway, which prevents them from working as trait objects IIRC), but rather to provide `vm-virtio` consumers with the option of easily switching between queue types and/or different implementations. Using conditionals to essentially harmonize the differences is an interesting approach, either on its own or in conjunction with others. I'll try to come up with a proposal and it's really interesting what other ideas are out there as well.

Last time we had a look, there was no existing solution meeting those requirements for the logging part, but it would be interesting to check again. Meanwhile, we're considering an abstraction based on call-backs where each specific backend decides what additional logic (if any) is invoked around a pre-defined set of events (there's an example in [this](https://github.com/rust-vmm/vm-superio/pull/20) PR).

jiangliu:
> Gave this a bit more thought, and it seems there are a number of things we should take into consideration going forward (not necessarily related to the original theme of the PR :D).
> 
> * It looks like having a `QueueConfig` object makes sense, especially since the configuration/state definition is largely the same for both split and packed queues (and we can switch to the queue-neutral terminology, i.e. `DriverArea` vs `AvailableRing`, etc).
> * Having different underlying implementation details might actually make sense for different use-cases, so we can define some interfaces for queues as well.
> * Kinda immediately related to the previous points is the question about how to approach adding support for packed queues as well. I'm currently looking at abstracting this behind a `Queue` interface. Do you think it makes sense to attempt a unifying approach?

Agree with Paolo, using ifs helps us to keep the same interface without explicit performance penalty. Linux kernel supports packed queue in this way too.
> * Last but not least, I think we might run into the issue of logging/metrics sooner rather than later. Right now, `vm-virtio` has a `log` dependency and uses the associated macros ("warn!", "error!", etc) in different places. Going forward, this might not be the right solution for all consumers of the crate, as ppl might prefer a different log crate, or different levels/messages, the ability to bump metrics, and other things.
> 
It would be great to remove dependency on the log crate.

> I'll create one or more issues for the above and later close the PR, but just wanted to mention them here for now. Also thanks for the comments and examples!

alexandruag:
Closing in favor of #41. PRs/issues for the other points discussed above will follow as well.

PR #29, opened 2020-11-03
[RFC] block: abstractions for request parsing and execution

lauralt:
This RFC proposes the following abstractions for block device:

- `Request` -> which handles the request parsing from a descriptor chain and keeps the necessary information for the request execution.
- `RequestExecutor` -> which handles the request execution and wraps the block device file to which the requests are sent. The reason for having a separate abstraction for execution is the fact that there can be multiple ways to execute a given request (an example is given in the corresponding commit).

`Request` is mostly based on the cloud hypervisor implementation as it handles the case of multiple data descriptors in the chain (as opposed to firecracker, where only one data descriptor is assumed) (side note: `writeback` field is not kept since it is most likely a feature of the block device and/or a configuration option for the executor, rather than of the request object itself).

The downside to this `Request` proposal is we are still making assumptions which represent a particular case, but they are only used as part of the implementation, and do not constrain the interface of Request. For example, our implementation starts by assuming we get separate descriptors for the request header (i.e. type, sector, etc), data regions, and status. However, these assumptions are only leveraged within the parse method implementation, which is opaque to users.
In the future we might take into consideration the `Reader` and `Writer` abstractions from crosvm which just follow the only hard requirement from the virtio specification: `The driver MUST place any device-writable descriptor elements after any device-readable descriptor elements`. If we later switch to something similar to the crosvm approach, we can keep the same request interface, where the data regions are a sequence of (GuestAddress, length) pairs.

This PR misses a lot of tests, it contains some examples for request parsing and execution, but they are far from a thorough testing.
If the overall aproach makes sense I'll add more tests (also the documentation can be improved).

The last commit is a discard/write zeroes implementation tentative (more or less the one from crosvm at this moment :D) and it wasn't tested in any way yet, so it can be omitted as well for now (but any suggestion is welcome).


src/block/request.rs
```diff
@@ -0,0 +1,667 @@
...
+        }
+        if status_desc.len() < 1 {
+            return Err(Error::DescriptorLengthTooSmall);
+        }
+
+        // Check that the address of the status descriptor is valid in guest memory.
+        // We will write an u32 status here after executing the request.
+        let _ = desc_chain
+            .memory()
+            .checked_offset(status_desc.addr(), mem::size_of::<u32>())
```
jiangliu, line 262: The Virtio spec defines that the status field is a u8, so should we check "u32" or "u8"?

src/block/request.rs
```diff
@@ -49,6 +49,8 @@ const VIRTIO_BLK_T_FLUSH: u32 = 4;
 const VIRTIO_BLK_T_DISCARD: u32 = 11;
 const VIRTIO_BLK_T_WRITE_ZEROES: u32 = 13;
 
+const VIRTIO_BLK_DISCARD_WRITE_ZEROES_FLAG_UNMAP: u32 = 1;
```
aghecenco, line 52: nit: could you call this `VIRTIO_BLK_WRITE_ZEROES_FLAG_UNMAP` instead to be consistent with the [kernel name](https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/virtio_blk.h#L178)? Makes stuff easier to find in the driver code

src/block/request.rs
```diff
@@ -181,6 +189,7 @@ impl Request {
     }
 
     /// Parses a request from a given `desc_chain`.
+    /// TODO split this gigantic function in smaller pieces.
```
aghecenco, line 192: :+1:

src/block/request.rs
```diff
@@ -204,11 +256,142 @@ impl Request {
...
+        Ok(())
+    }
+
+    /// Executes `request` Request on `D` and `mem`.
+    pub fn execute<M: GuestMemory>(
+        &mut self,
+        mem: &M,
+        request: &Request,
+    ) -> result::Result<u32, ExecuteError> {
+        self.seek(SeekFrom::Start(request.sector() << SECTOR_SHIFT))
```
aghecenco, line 361: Unchecked arithmetic alert: use [`checked_shl`](https://doc.rust-lang.org/std/primitive.u32.html#method.checked_shl) instead

aghecenco (commented):
There have been previous mentions in other PRs about which virtio spec we should adhere to, and some of my comments here also touch on that - I've been looking over the virtio 1.0 and 1.1 specs reviewing this, and also through the driver code in the latest kernel. Stuff differs. Should we use this opportunity to propose a support model (i.e. which version of the virtio spec) applicable to this crate?

src/block/request.rs
```diff
@@ -0,0 +1,204 @@
...
+unsafe impl ByteValued for RequestHeader {}
+
+impl Request {
+    /// Returns the request type.
+    pub fn request_type(&self) -> RequestType {
+        self.request_type
+    }
+
+    /// Returns the (address, len) pairs where the request data is in the guest memory.
+    pub fn data(&self) -> &Vec<(GuestAddress, u32)> {
```
alexandruag, line 177: We could just return a `&[(GuestAddress, u32)]` slice here.

src/block/request.rs
```diff
@@ -204,11 +256,142 @@ impl Request {
...
+///
+/// ```rust
+/// # use std::fs::File;
+/// # use vm_memory::GuestAddress;
+/// # use vm_virtio::request::RequestExecutor;
+///
+/// let file = File::create("foo.txt").unwrap();
+/// let request_exec = RequestExecutor::new(file);
+/// ```
+pub struct RequestExecutor<D: DiskFile> {
```
alexandruag, line 284: Does it makes sense to add the executor to its own module file? I think it's cleaner to look at that way.

src/block/request.rs
```diff
@@ -22,20 +26,38 @@
...
 const VIRTIO_BLK_T_FLUSH: u32 = 4;
 const VIRTIO_BLK_T_DISCARD: u32 = 11;
 const VIRTIO_BLK_T_WRITE_ZEROES: u32 = 13;
 
+const SECTOR_SHIFT: u8 = 9;
+const SECTOR_SIZE: u64 = (0x01 as u64) << SECTOR_SHIFT;
+
+/// Trait that keeps as supertraits the ones that are necessary for virtio block
+/// request execution on a block device (v1.1 virtio specification).
+pub trait DiskFile: Read + Write + Seek + FileSync + PunchHole + WriteZeroes {}
```
alexandruag, line 59: I think a more generic name like `Backend` is more suitable, but that's def subject to opinion.

src/block/request.rs
```diff
@@ -204,11 +256,142 @@ impl Request {
...
+                }
+                RequestType::Out => {
+                    self.check_access(
+                        RequestExecutor::<D>::len_to_num_sectors(*data_len),
+                        request.sector(),
+                    )?;
+                    mem.write_to(*data_addr, self.deref_mut(), *data_len as usize)
+                        .map_err(ExecuteError::Write)?;
+                }
+                RequestType::Flush => {
```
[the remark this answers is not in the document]
alexandruag: Yeah, looks like we can pass the negotiated feature bits of the device to the executor (as an `u64`) and check the corresponding bit before performing an operation (and return an `Unsupported` when not present).

Another question is whether it's better to use the `fsync` here or the `flush` operation provided by the `Write` trait. Does anyone have a better picture about what the differences are here?

src/block/request.rs
```diff
@@ -204,11 +256,142 @@ impl Request {
...
+                    return match self.fsync() {
```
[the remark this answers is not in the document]
alexandruag: Doesn't this work inside the loop as well?

src/block/request.rs
```diff
@@ -204,11 +256,142 @@ impl Request {
...
+                        Ok(_) => Ok(0),
+                        Err(e) => Err(ExecuteError::Flush(e)),
+                    }
+                }
+                RequestType::Discard | RequestType::WriteZeroes => {}
```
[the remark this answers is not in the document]
alexandruag: I think it's better to return `Unsupported` here for now.

src/block/request.rs
```diff
@@ -204,11 +256,142 @@ impl Request {
...
+                RequestType::Unsupported(t) => return Err(ExecuteError::Unsupported(t)),
```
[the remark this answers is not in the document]
alexandruag: I think that moving this outside the for would prob still be somewhat ugly due to needing some arm at the end anyway so the match is exhaustive :-s

src/block/request.rs
alexandruag: Hmm looks like the `Discard` and `WriteZeroes` support is noticeably more complex than the rest. Does it make sense to have two PRs? We can focus first on the previous commits and then create another PR for the rest.

src/block/request.rs
```diff
@@ -204,11 +256,142 @@ impl Request {
...
+            .checked_add(sector)
+            .ok_or(ExecuteError::InvalidAccess)?;
+        if sectors_count > self.num_sectors() {
+            return Err(ExecuteError::InvalidAccess);
+        }
+        Ok(())
+    }
+
+    /// Executes `request` Request on `D` and `mem`.
+    pub fn execute<M: GuestMemory>(
```
alexandruag, line 356: From a high level, it seems more convenient if the executor writes the status of the request as well. Does that make sense? I that case we probably no longer need `Unsupported` as an error variant, but haven't really looked into the details.

alexandruag (commented):
Hi! Left some comments, and will do another pass to focus more on the final commit in particular.

bonzini (changes requested):
I think the use of Deref and DerefMut in RequestExecutor is a symptom of a wrong design. In particular, RequestExecutor actually has two parts:

1) caching the number of sectors in the backend, and converting virtio-blk's offset/count API to separate seek and read/write operations. This can be implemented in a trait `SectorAccess` with a corresponding implementation for `Read + Write + Seek + PunchHole + WriteZeroes + FileSync`:

```rust
pub trait SectorAccess: FileSync {
    // Notice that the start/offset API allows for interior mutability
    // This will be needed in order to process multiple requests
    // in parallel in the future
    pub fn read(&self, sector_start: u64, value: &mut[u8] ) -> Result<usize>;
    pub fn write(&self, sector_start: u64, value: &[u8]) -> Result<usize>;
    pub fn unmap(&self, sector_start: u64, sector_count: usize) -> Error;
    pub fn punch_hole(&self, sector_start: u64, sector_count: usize) -> Error;

    pub fn num_sectors(&self) -> u64;
    pub fn sector_size(&self) -> u32;
}

pub struct DiskFile<D: Read + Write + Seek + PunchHole + WriteZeroes + FileSync> {
    // Mutex needed to avoid interruption between seek and read
    file: Mutex<D>,
    num_sectors: u64
}

impl<D: Read + Write + Seek + PunchHole + WriteZeroes + FileSync> SectorAccess for DiskFile<D> { /* ... */ }
```

This allows `SectorAccess` implementations that do not use the Rust `std::io` traits, for example using io_uring, or other APIs that have native support for offset/count.

If you want, you can also add a file pointer field to `DiskFile`, and use it to implement Read/Write/Seek, for example:

```rust
impl Read for DiskFile {
     fn read(&mut self, buf: &mut [u8]) -> Result<usize> {
         let g = self.file.lock().unwrap();
         g.seek(SeekFrom::Start(self.pointer));
         let nread = g.read(buf)?;
         self.pointer += nread;
         Ok(nread)
    }
}

impl Seek for DiskFile {
     fn seek(&mut self, pos: SeekFrom) -> Result<u64> {
         let g = self.file.lock().unwrap();
         let oldpos = self.pointer;
         self.pointer = g.seek(pos)?;
         Ok(oldpos)
    }
}
```

2) the parsing of requests and their forwarding to a `SectorAccess`. This would be a separate class that operates on a `SectorAccess` through a reference or an Rc/Arc:

```rust
// B can be an &A, Rc<A>, Arc<A> etc.
pub struct RequestParser<A: SectorAccess, B: Deref<Target = A>> {
    backend: B,
}
```

Once you split the backend this way, you don't need anymore to access the disk through the `RequestExecutor`. Instead you can just clone the Rc/Arc and let interior mutability do its job.

alexandruag:
Hi Paolo! The `Deref` implementation for `RequestExecutor` can be removed, and the parsing of request objects (which is a generic operation in itself) is already separated from the execution part (which can be implemented in multiple ways). `RequestExecutor` (I guess a better name would have been `SimpleRequestExecutor`) represents a simple initial implementation of block request execution, while we figure out what the best approach is towards a more powerful abstraction that, like you mentioned, works with stuff like `io_uring` and others.

The simple implementation can later be superseded (or coexist) with more complex solutions. Do you think it's reasonable to start with the simplified approach and then add an advanced abstraction as well?

bonzini:
> RequestExecutor (I guess a better name would have been SimpleRequestExecutor) represents a simple initial implementation of block request execution, while we figure out what the best approach is towards a more powerful abstraction that, like you mentioned, works with stuff like io_uring and others.

What I don't like is that the parsing of the requests is tied to Read and Write, which are the "wrong" API to access files for a virtio backend (i.e. an API that has to be worked around). It happens to work because vm-memory's `Bytes` trait supports Read and Write, but it's basically impossible to turn it into a scalable implementation without a complete rewrite.

I could buy renaming `RequestExecutor` into `StdIoBackend`, I guess. I'm just not sure if it's a good idea to base the reference implementation on suboptimal traits from `std::io` instead of defining our own.

alexandruag:
The proposed request executor is more of a minimal viable implementation than a reference one (we can make this even clearer in the documentation), which can already enable use cases like Firecracker. The original intention was to have multiple solutions for block request execution, based on different sets of simplifying assumptions. We def want to add more powerful abstractions as well, but it would be great to start with this very simple one that’s immediately useful. We’re committed to evolving and improving things as we figure out what the best path forward looks like. Is that ok with you?

lauralt (commented):
Hello, thanks for the comments! I renamed `RequestExecutor` to `StdIoBackend` and updated the documentation to something that is hopefully clearer (also removed the `Deref`s). I like the `SectorAccess` proposal and played a bit with it, but it doesn't look so straightforward (at least for me), maybe we can focus on this abstraction in a future PR (when we'll define what executors we plan to have, for example).

I removed for now the last commit and will open another PR with discard, write zeroes and get device id ops if it's okay for everyone. I also added more tests and the documentation is better now (I hope).

src/block/executor.rs
```diff
@@ -0,0 +1,437 @@
...
+        }
+        if !self.has_feature(VIRTIO_BLK_F_FLUSH) && request.request_type() == RequestType::Flush {
+            return Err(Error::Unsupported(VIRTIO_BLK_T_FLUSH));
+        }
+
+        match request.request_type() {
+            RequestType::In => {
+                for (data_addr, data_len) in request.data() {
+                    self.check_access(
+                        StdIoBackend::<B>::len_to_num_sectors(*data_len + bytes_from_dev),
```
alexandruag, line 192: Is it possible to change `check_access` so the conversion from number of bytes to number of sectors happens within, so we can get rid of `len_to_num_sectors` as a separate method as well?

alexandruag (commented):
Cool, left some comments for now, and I'll have a better look at the tests during a second pass.

src/block/executor.rs
lauralt: The reason behind this split was the fact that for discard/write zeroes we receive the number of sectors in the request and it seemed cleaner to perform the check directly based on that number for those ops. If it hurts readability much, I can revert to one function.

src/block/request.rs
```diff
@@ -0,0 +1,244 @@
...
+//! implementation more generic by removing any constraint regarding the number of descriptors
+//! that each request element can occupy. Basically, the only hard requirement regarding the
+//! request descriptors is: The driver MUST place any device-writable descriptor elements after any
+//! device-readable descriptor elements (see
+//! [this](https://docs.oasis-open.org/virtio/virtio/v1.1/csprd01/virtio-v1.1-csprd01.html#x1-280004)).
+//! However, we expect the `Request` interface won't change even if we switch to the general
+//! approach.
+//!
+//! # Note
+//! Wo offer support only for virtio v1.0+
```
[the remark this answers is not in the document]
lauralt: I moved the note in README.

src/block/request.rs
```diff
@@ -0,0 +1,244 @@
...
+};
+
+const VIRTIO_BLK_T_IN: u32 = 0;
+const VIRTIO_BLK_T_OUT: u32 = 1;
+const VIRTIO_BLK_T_FLUSH: u32 = 4;
+
+/// Block request parsing errors.
+#[derive(Debug)]
+pub enum Error {
+    /// Guest gave us too few descriptors in a descriptor chain.
```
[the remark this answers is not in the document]
lauralt: done

src/block/request.rs
```diff
@@ -0,0 +1,244 @@
...
+        self.sector
+    }
+
+    /// Returns the status address.
+    pub fn status_addr(&self) -> GuestAddress {
+        self.status_addr
+    }
+
+    // Checks that a descriptor meets the minimal requirements for a valid status descriptor.
+    fn check_status_desc<M: GuestAddressSpace>(mem: &<M>::M, desc: Descriptor) -> Result<()> {
```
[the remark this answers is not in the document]
lauralt: done

src/block/request.rs
```diff
@@ -0,0 +1,244 @@
...
+        // It makes sense to check the data length from descriptor, since we're going to
+        // keep from now on only the address of the status buffer in the `Request` object
+        // so we won't be able to validate this at some other point.
+        if desc.len() < 1 {
+            return Err(Error::DescriptorLengthTooSmall);
+        }
+
+        // Check that the address of the status descriptor is valid in guest memory.
+        // We will write an u8 status here after executing the request.
+        let _ = mem
```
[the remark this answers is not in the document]
lauralt: As mentioned in another comment, I would leave these checks so that we don't have partial execution of requests before the error is thrown (for read, write) or entire request execution for the case when the status address is invalid.

src/block/request.rs
```diff
@@ -0,0 +1,244 @@
...
+    ///
+    /// The checks we're doing here are mostly the `MUST`s from the virtio specification.
+    /// We are also partially validating `The driver MUST place any device-writable descriptor
+    /// elements after any device-readable descriptor elements` rule by checking that the
+    /// expected header descriptor (the chain head) is device-readable and the expected status
+    /// descriptor (the chain tail) is device-writable.
+    ///
+    /// # Arguments
+    /// * `desc_chain` - A reference to the descriptor chain that should point to the buffers of a
+    ///                - virtio block request.
```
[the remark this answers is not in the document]
lauralt: oops, done

src/block/executor.rs
lauralt: That function is no longer needed, in the meantime I found [this commit](https://github.com/oasis-tcs/virtio-spec/commit/bef3ff7f1e5e81bf9734b2f73f0ef26447358805) which clearly states that data for IN and OUT should be a multiple of 512B, so now I'm returning an error if this condition is not met.

lauralt:
@bonzini @alexandruag @aghecenco @jiangliu I addressed your comments, can you please take another look? The documentation is now updated to clearly state that the `StdIoBackend` is just a _simple_ executor, not a reference one. Do you think there is something else we should do here or can we proceed in merging this one (and open an issue for future abstractions/executors etc.)?

bonzini:
Can you hide StdIoBackend behind a feature flag?

lauralt:
Sure, done.

src/block/stdio_executor.rs
```diff
@@ -0,0 +1,453 @@
...
+
+/// Trait that keeps as supertraits the ones that are necessary for the `StdIoBackend` abstraction
+/// used for the virtio block request execution.
+pub trait Backend: Read + Write + Seek + FileSync + PunchHole + WriteZeroes {}
+
+impl<B: Read + Write + Seek + FileSync + PunchHole + WriteZeroes> Backend for B {}
+
+/// Errors encountered during request execution.
+#[derive(Debug)]
+pub enum Error {
```
alexandruag, line 53: Can we add something like [this](https://github.com/firecracker-microvm/firecracker/blob/c1ec34482f335aacee99901de472b5b0e53468d9/src/devices/src/virtio/block/request.rs#L32) method to ease converting an error to the appropriate status value?

alexandruag (commented):
Looks good to me; left just one small comment.

src/block/stdio_executor.rs
lauralt: I wanted to give this a bit more thought in the next PR when I'll have a better idea of how to write the status too and regarding the method you've mentioned, for example, I would've liked to have some sort of `From` impl.
Anyway, I can quickly add the `status` method as is if it is really required for your PR in vmm-reference (but I expect to change it in the next PR).

alexandruag (approved):

src/block/stdio_executor.rs
alexandruag: It's not immediately required, the temporary solution we're using is fine for the time being.

lauralt:
I opened [an issue](https://github.com/rust-vmm/vm-virtio/issues/31) for the file access abstraction and also added here a commit for the `backend-stdio` feature. @bonzini can you please take another look? Thanks!

alexandruag (approved):

aghecenco (approved):

andreeaflorescu:
@bonzini we are going to merge this PR so we can also publish the next one that adds the discard/write zeroes operations. If there are any remaining concerns, we will address them in a following PR.

PR #30, opened 2020-11-06
Generic abstractions for virtio devices and the MMIO transport

alexandruag:
This PR proposes generic abstractions for virtio devices and the MMIO transport in particular (the aim is to add PCI support as well at some point). We first redefine the `VirtioDevice` trait, which contains methods required to support interactions that happen at the transport level. The following commit defines `VirtioMmioDevice`, which provides default implementations (based on the functionality from `VirtioDevice`) for reads and writes to the registers/configuration space of a virtio device that uses the MMIO transport. Finally, the last commit introduces the `VirtioConfig`, `WithVirtioConfig`, and `WithDeviceOps` helper abstractions, which are used to provide automatic implementations of the previous traits.

There are more details present in the code comments. This [commit](https://github.com/alexandruag/vm-virtio/commit/32f26359c6fa5996f62ae767115c793f43ab78af#diff-060bce9da11057f730f84bbc0e029eb859f54cc5e05b05bed48f5a8f1d4df524) contains a simple example of how the previous pieces fit together to simplify the implementation of virtio-over-MMIO functionality for a potential device. I'm working on a more complete example of an actual working device and will add it here as well once it's available. Looking forward to any feedback!


alexandruag:
[Here](https://github.com/alexandruag/vmm-reference/blob/block_rfc/src/devices/src/virtio/block/device.rs)'s a better example of how the previously mentioned abstractions can be used as part of a block device implementation.

src/devices/mmio.rs
```diff
@@ -0,0 +1,199 @@
...
+                        warn!("unknown virtio mmio register read: 0x{:x}", offset);
+                        return;
+                    }
+                };
+                // This cannot panic, because we checked that `data.len() == 4`.
+                data.copy_from_slice(v.to_le_bytes().as_ref());
+            }
+
+            // TODO: The standard specifies that configuration space size is device specific, so
+            // we might want to express that via the trait instead of hard coding the current
```
aghecenco, line 114: +1. It can have a const fn and default implementation, not a big change, and would be neater.

src/devices/mmio.rs
```diff
@@ -0,0 +1,199 @@
...
+            // arbitrary ceiling.
+            // It's ok to use `as` here because `offset` always fits into an `usize` in this case.
+            0x100..=0xfff => self.read_config(offset as usize - 0x100, data),
```
aghecenco, line 117: According to the spec: `Device-specific configuration space starts at the offset 0x100 and is accessed with byte alignment.`
Admittedly, `byte alignment` is a weird phrase but looking next door, qemu interprets this as `only byte, word or dword accesses are allowed` ([code](https://github.com/qemu/qemu/blob/c8e5c4b246584da36694a3c259a7dbb8a7e7b1f3/hw/virtio/virtio-mmio.c#L113)). I think this makes sense and we should do the same check, WDYT?

slp:
I like both the design and the implementation. :+1:

src/devices/mmio.rs
rbradford: The byte alignment here means you're not restricted to only accessing it at a word,dword or qword granularity. i.e. you can access it however you like so there is no check you can make.

alexandruag:
Thanks for the comments everyone! I'll address them and turn this into a regular PR over the following days.

src/lib.rs
```diff
@@ -14,10 +14,10 @@ extern crate log;
 extern crate vm_memory;
 extern crate vmm_sys_util;
 
-mod device;
+mod devices;
 mod queue;
 
-pub use self::device::*;
+pub use self::devices::*;
```
andreeaflorescu: I think that queues and devices are conceptually different and can be used independently (at least the queues). For this reason, what do you say about exporting the modules directly instead of `pub use`?

My general thinking when using flattened vs non-flattened exports is that flattened exports are to be used when the modules are there just to provide a structural separation, and not a logical one.

**Structural separation**: If everything would be in one module, than the module would be to big, and thus the code becomes hard to follow, hence we are splitting it in multiple modules. This artificial separation is hidden from the end user by exporting everything flattened.
**Logical separation**: separate concepts are split between logical modules.

src/devices/mmio.rs
```diff
@@ -0,0 +1,199 @@
...
+// TODO: Crosvm was using 0 here a while ago, and Firecracker started doing that as well. Should
+// we leave it like that, or should we use the VENDOR_ID value for PCI Virtio devices? It looks
+// like the standard doesn't say anything regarding an actual VENDOR_ID value for MMIO devices.
+const VENDOR_ID: u32 = 0;
+
+// Helper function that runs the provided closure to mutate the currently selected queue of
+// a `VirtioDevice`, provided the status check is successful.
+// TODO: This function and its uses will likely have to be updated when we start offering
+// packed virtqueue support as well.
+fn update_queue_field<M, D, F>(device: &mut D, f: F)
```
andreeaflorescu, line 32: I deeply dislike these kind of functions that don't belong to any object :(

andreeaflorescu (commented):
I just looked briefly over the interfaces. I'll do an in-depth review of the implementation when this is not longer an RFC.

I find the `WithDriverSelection` a bit too strange. I believe we can work on improving the readability of that code by providing a better abstraction (that is in case I understood correctly what you're trying to achieve with it).

Otherwise, the PR LGTM.

src/devices/mmio.rs
alexandruag: On second though, the effective size of the configuration space is actually variable depending on selected features, so I would keep the static upper bound and update TODO for now to think a bit more about this. Ultimately, each specific implementation of `read/write_config` takes care of validations, etc.

src/devices/virtio_config.rs
```diff
@@ -0,0 +1,248 @@
...
+/// An object that provides a common virtio device configuration representation. It is not part
+/// of the main `vm-virtio` set of interfaces, but rather can be used as a helper object in
+/// conjunction with the `WithVirtioConfig` trait (provided in the same module), to enable the
+/// automatic implementation of other traits such as `VirtioDevice` and `WithDriverSelect`.
+// Adding the `M` generic parameter that's also required by `VirtioDevice` for the time being.
+// The various members have `pub` visibility until we determine whether it makes sense to drop
+// this in favor of adding accessors.
+pub struct VirtioConfig<M: GuestAddressSpace> {
+    /// The set of features exposed by the device.
+    pub device_features: u64,
```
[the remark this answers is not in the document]
alexandruag: Yeah so like I mentioned in a previous comment, the driver can write a full 32-bit value when selecting a page, but then all the feature bits defined by the standard so far appear to fit within a `u64`. I'll use the simplification mentioned there if it sounds good.

src/devices/virtio_config.rs
```diff
@@ -0,0 +1,248 @@
...
+    // for devices that support multiple queues which might not all be configured/activated by
+    // the driver.
+    fn queues_valid(&self) -> bool {
+        self.virtio_config().queues.iter().all(Queue::is_valid)
+    }
+}
+
+/// Another helper trait, which (in conjunction with `WithVirtioConfig`) is used to provide
+/// automatic implementations of `VirtioDevice` and `WithDriverSelect`.
+pub trait WithDeviceOps {
```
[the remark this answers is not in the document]
alexandruag: Yeah so I don't know which name is best here, but at the same time I don't know if `DeviceStateUpdate` fits that well. We can think a bit more and just go with something in the end.

src/devices/virtio_config.rs
```diff
@@ -0,0 +1,248 @@
...
+    fn queue(&self, index: u16) -> Option<&Queue<M>> {
+        self.virtio_config().queues.get(usize::from(index))
+    }
+
+    fn queue_mut(&mut self, index: u16) -> Option<&mut Queue<M>> {
+        self.virtio_config_mut().queues.get_mut(usize::from(index))
+    }
+
+    fn num_feature_pages(&self) -> u32 {
+        // We use an `u64` to keep track of features within `VirtioConfig`.
```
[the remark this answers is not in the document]
alexandruag: This method might disappear, depending on the discussion around the previous comment about removing it.

alexandruag:
I've simplified `devices_features`/`driver_features` handling a bit (as per one of the previous comments) and added unit tests in a new, separate commit. I also removed the draft status from the PR. Thanks for all the comments so far and please take another look.

alexandruag:
Did another push that renames some of the helper traits in the fourth commit, uses `BorrowMut` to abstract access to a `VirtioConfig` object, and we're also setting the `event_idx` queue flag as part of feature acknowledgement for extra convenience.

src/device/mmio.rs
```diff
@@ -0,0 +1,355 @@
...
+                let v = match offset {
+                    0x0 => MMIO_MAGIC_VALUE,
+                    0x04 => MMIO_VERSION,
+                    0x08 => self.device_type(),
+                    0x0c => VENDOR_ID,
+                    0x10 => match self.device_features_select() {
+                        0 => self.device_features() as u32,
+                        1 => (self.device_features() >> 32) as u32,
+                        // No device features defined beyond the first two pages.
+                        _ => 0,
```
aghecenco, line 92: This can be tested too fairly easily if the `Dummy` device does `d.set_device_features_select(2)` before an MMIO read from offset `0x10`

src/device/mmio.rs
```diff
@@ -0,0 +1,355 @@
...
+                    0x24 => self.set_driver_features_select(v),
+
+                    // TODO: add warnings or signal some sort of event (depending on how we end up
+                    // implementing logging and metrics) for values that do not actually fit the
+                    // data type specified by the virtio standard (we simply use `as` conversion
+                    // for now).
+                    0x30 => self.set_queue_select(v as u16),
+                    0x38 => update_queue_field(self, |q| q.size = v as u16),
+                    0x44 => update_queue_field(self, |q| q.ready = v == 1),
+                    0x50 => self.queue_notify(v),
```
aghecenco, line 163: This is also rather easily testable if you override `fn queue_notify` for `Dummy` and force a write here.

src/device/mod.rs
```diff
@@ -0,0 +1,289 @@
...
+                }
+
+                // TODO: Bring this (and other feature definitions) to the vm-virtio crate proper.
+                // Using a local const temporarily until then.
+                const VIRTIO_F_RING_EVENT_IDX: u64 = 29;
+
+                // Set the appropriate configuration flag for all queues if we offered the
+                //`VIRTIO_F_RING_EVENT_IDX` feature and the driver acknowledged it.
+                if self.driver_features() & (1 << VIRTIO_F_RING_EVENT_IDX) != 0 {
+                    for i in 0..self.num_queues() {
```
aghecenco, line 146: I think there's a way to test this too with `Dummy` though I don't have the specifics.

aghecenco (approved):
I have some final nits and minor test enhancements that don't look too hard to add, but I think the PR is good to go as-is. Lack of proper documentation is my main note here, but that can wait until the crate is ready for publishing.

slp (dismissed):

slp:
LGTM. I think it can be merged once is rebased.

src/device/virtio_config.rs
```diff
@@ -0,0 +1,227 @@
...
+
+    fn read_config(&self, offset: usize, data: &mut [u8]) {
+        let config_space = &self.borrow().config_space;
+        let config_len = config_space.len();
+        if offset >= config_len {
+            error!("Failed to read from config space");
+            return;
+        }
+
+        // TODO: Are partial reads ok?
```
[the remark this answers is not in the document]
alexandruag: Sure, will add an issue immediately after this gets merged in the current form.

alexandruag (commented):
Addressed the latest comments and did a rebase.

aghecenco (approved):

src/device/mod.rs
```diff
@@ -0,0 +1,323 @@
...
+    pub const RESET: u8 = 0;
+    /// Indicates that the guest OS has found the device and recognized it as a valid
+    /// virtio device.
+    pub const ACKNOWLEDGE: u8 = 1;
+    /// Indicates that the guest OS knows how to drive the device.
+    pub const DRIVER: u8 = 2;
+    /// Indicates that something went wrong in the guest, and it has given up on the device.
+    /// This could be an internal error, or the driver didn’t like the device for some reason,
+    /// or even a fatal error during device operation.
+    pub const FAILED: u8 = 128;
```
jiangliu, line 48: Could we sort these consts by value?

slp (approved):

.buildkite/custom-pipeline.yml
```diff
@@ -0,0 +1,52 @@
...
+    commands:
+      - cargo build --release --features=backend-stdio
+    retry:
+      automatic: false
+    agents:
+      platform: x86_64.metal
+      os: linux
+    plugins:
+      - docker#v3.0.1:
+          image: "rustvmm/dev:v6"
```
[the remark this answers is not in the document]
alexandruag: Yeah it makes sense to do so; we can piggy back the update on another PR.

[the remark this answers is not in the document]
andreeaflorescu: I've opened an issue to update the rust-vmm-ci as well as it is super old: https://github.com/rust-vmm/vm-virtio/issues/34

I'll add a comment to also update the container version in the custom-pipeline.

src/device/mod.rs
alexandruag: They were initially written down in the order they are presented in the standard (i.e. in section 2.1), but we can def reorder them if it improves clarity & readability.

Issue #34, opened 2020-12-11
update rust-vmm-ci


andreeaflorescu:
This crate is tested with an old container version (v6).

andreeaflorescu:
As part of this update, we should also update the custom buildkite pipelines to use the latest container.

PR #35, opened 2020-12-14
block: add support for the rest of requests

lauralt:
- added discard, write zeroes and get device id requests
- added a method to StdIoBackend that processes the request execution result, by writing the status byte to memory and returning the number of bytes written to memory. For now, this method assumes that if writing to memory in `execute` method fails, no bytes were written to memory, which is not necessarily true. Will add in a subsequent PR/commit a logic that handles partial writing too.
+ some other fixes

Also, from a high level view over #33, looks like the entire `block` module could benefit in the future from the Reader and Writer wrappers and we can switch to using them here too, once they get merged.

### DIFF
--- a/src/block/request.rs
+++ b/src/block/request.rs
@@ -24,7 +24,7 @@
 //! approach.
 
 use std::fmt::{self, Display};
-use std::{mem, result};
+use std::result;
 
 use crate::{
     block::defs::{
@@ -173,22 +173,16 @@ impl Request {
             return Err(Error::DescriptorLengthTooSmall);
         }
 
-        // Check that the address of the status descriptor is valid in guest memory.
+        // Check that the address of the status is valid in guest memory.
         // We will write an u8 status here after executing the request.
-        let _ = mem
-            .checked_offset(desc.addr(), mem::size_of::<u8>())
-            .ok_or_else(|| {
-                Error::GuestMemory(GuestMemoryError::InvalidGuestAddress(desc.addr()))
-            })?;
+        let _ = mem.check_address(desc.addr()).ok_or_else(|| {
+            Error::GuestMemory(GuestMemoryError::InvalidGuestAddress(desc.addr()))
+        })?;
         Ok(())
     }
 
     // Checks that a descriptor meets the minimal requirements for a valid data descriptor.
-    fn check_data_desc<M: GuestMemory>(
-        mem: &M,
-        desc: Descriptor,
-        request_type: RequestType,
-    ) -> Result<()> {
+    fn check_data_desc(desc: Descriptor, request_type: RequestType) -> Result<()> {
         // We do this check only for the device-readable buffers, as opposed to
         // also check that the device doesn't want to read a device-writable buffer
         // because this one is not a MUST (the device MAY do that for debugging or
@@ -196,13 +190,6 @@ impl Request {
         if !desc.is_write_only() && request_type == RequestType::In {
             return Err(Error::UnexpectedReadOnlyDescriptor);
         }
-
-        // Check that the address of the data descriptor is valid in guest memory.
-        let _ = mem
-            .checked_offset(desc.addr(), desc.len() as usize)
-            .ok_or_else(|| {
-                Error::GuestMemory(GuestMemoryError::InvalidGuestAddress(desc.addr()))
-            })?;
         Ok(())
     }
 
@@ -243,7 +230,7 @@ impl Request {
         let mut desc = desc_chain.next().ok_or(Error::DescriptorChainTooShort)?;
 
         while desc.has_next() {
-            Request::check_data_desc::<<M>::M>(desc_chain.memory(), desc, request.request_type)?;
+            Request::check_data_desc(desc, request.request_type)?;
 
             request.data.push((desc.addr(), desc.len()));
             desc = desc_chain.next().ok_or(Error::DescriptorChainTooShort)?;
@@ -428,31 +415,6 @@ mod tests {
         assert_eq!(
             Request::parse(&mut chain).unwrap_err(),
             Error::UnexpectedReadOnlyDescriptor
-        );
-
-        // Invalid data buffer address.
-        let v = vec![
-            Descriptor::new(0x10_0000, 0x100, 0, 0),
-            Descriptor::new(0xFFF_FFF0, 0x100, VIRTQ_DESC_F_WRITE, 0),
-            Descriptor::new(0x30_0000, 0x200, VIRTQ_DESC_F_WRITE, 0),
-            Descriptor::new(0x40_0000, 0x100, VIRTQ_DESC_F_WRITE, 0),
-        ];
-        let req_header = RequestHeader {
-            request_type: VIRTIO_BLK_T_OUT,
-            _reserved: 0,
-            sector: 2,
-        };
-        mem.write_obj::<RequestHeader>(req_header, GuestAddress(0x10_0000))
-            .unwrap();
-
-        let mut chain = build_desc_chain(&mem, &v[..4]);
-
-        // The first data descriptor would cause a write beyond memory capacity.
-        assert_eq!(
-            Request::parse(&mut chain).unwrap_err(),
-            Error::GuestMemory(GuestMemoryError::InvalidGuestAddress(GuestAddress(
-                0xFFF_FFF0,
-            )))
         );
 
         // Invalid status address.
--- a/src/block/request.rs
+++ b/src/block/request.rs
@@ -27,7 +27,10 @@ use std::fmt::{self, Display};
 use std::{mem, result};
 
 use crate::{
-    block::defs::{VIRTIO_BLK_T_FLUSH, VIRTIO_BLK_T_IN, VIRTIO_BLK_T_OUT},
+    block::defs::{
+        VIRTIO_BLK_T_DISCARD, VIRTIO_BLK_T_FLUSH, VIRTIO_BLK_T_IN, VIRTIO_BLK_T_OUT,
+        VIRTIO_BLK_T_WRITE_ZEROES,
+    },
     queue::DescriptorChain,
     Descriptor,
 };
@@ -79,6 +82,10 @@ pub enum RequestType {
     Out,
     /// Flush request.
     Flush,
+    /// Discard request.
+    Discard,
+    /// Write zeroes request.
+    WriteZeroes,
     /// Unknown request.
     Unsupported(u32),
 }
@@ -89,6 +96,8 @@ impl From<u32> for RequestType {
             VIRTIO_BLK_T_IN => RequestType::In,
             VIRTIO_BLK_T_OUT => RequestType::Out,
             VIRTIO_BLK_T_FLUSH => RequestType::Flush,
+            VIRTIO_BLK_T_DISCARD => RequestType::Discard,
+            VIRTIO_BLK_T_WRITE_ZEROES => RequestType::WriteZeroes,
             t => RequestType::Unsupported(t),
         }
     }
--- a/src/block/request.rs
+++ b/src/block/request.rs
@@ -28,8 +28,8 @@ use std::{mem, result};
 
 use crate::{
     block::defs::{
-        VIRTIO_BLK_T_DISCARD, VIRTIO_BLK_T_FLUSH, VIRTIO_BLK_T_IN, VIRTIO_BLK_T_OUT,
-        VIRTIO_BLK_T_WRITE_ZEROES,
+        VIRTIO_BLK_T_DISCARD, VIRTIO_BLK_T_FLUSH, VIRTIO_BLK_T_GET_ID, VIRTIO_BLK_T_IN,
+        VIRTIO_BLK_T_OUT, VIRTIO_BLK_T_WRITE_ZEROES,
     },
     queue::DescriptorChain,
     Descriptor,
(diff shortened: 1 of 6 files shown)


.buildkite/custom-pipeline.yml
```diff
@@ -0,0 +1,52 @@
...
+    commands:
+      - cargo build --release --features=backend-stdio
+    retry:
+      automatic: false
+    agents:
+      platform: x86_64.metal
+      os: linux
+    plugins:
+      - docker#v3.0.1:
+          image: "rustvmm/dev:v6"
```
andreeaflorescu, line 12: Do you mind also fixing this: #34? We're using a really old container :(

src/block/defines.rs
```diff
@@ -0,0 +1,22 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
andreeaflorescu, line 1: I find it preferably to have the defines in corresponding files (as it was before). It is easier to find them if they're defined closer to where they're used or in the modules that they logically belong to.

src/block/defines.rs
```diff
@@ -9,12 +9,24 @@ pub const VIRTIO_BLK_T_IN: u32 = 0;
...
 
 // Feature bits.
 /// Read-only device.
 pub const VIRTIO_BLK_F_RO: u64 = 5;
 /// Flush command supported.
 pub const VIRTIO_BLK_F_FLUSH: u64 = 9;
+/// Discard command supported.
+pub const VIRTIO_BLK_F_DISCARD: u64 = 13;
+/// Write zeroes command supported.
+pub const VIRTIO_BLK_F_WRITE_ZEROES: u64 = 14;
```
andreeaflorescu, line 27: Looks like these are defined twice.

src/block/defines.rs
```diff
@@ -9,12 +9,24 @@ pub const VIRTIO_BLK_T_IN: u32 = 0;
...
+
+/// If this bit is set, the targeted range should be unmapped
+/// (only valid for write zeroes command).
+pub const VIRTIO_BLK_WRITE_ZEROES_FLAG_UNMAP: u32 = 1;
```
andreeaflorescu, line 31: This should be called somehow such that it is clear that it represents a bit. How about `WRITE_ZEROES_UNMAP_BIT`? I don't think we need to prefix it with VIRITO_BLK because it does not map 1:1 to the defines in the virtio spec.

src/block/stdio_executor.rs
```diff
@@ -25,30 +25,53 @@
...
-pub trait Backend: Read + Write + Seek + FileSync + PunchHole + WriteZeroes {}
+pub trait Backend: Read + Write + Seek + FileSync + PunchHole + WriteZeroesAt {}
+
+impl<B: Read + Write + Seek + FileSync + PunchHole + WriteZeroesAt> Backend for B {}
+
+/// One or more `DiscardWriteZeroesData` structs are used to describe the data for
+/// discard or write zeroes command.
+#[derive(Copy, Clone, Debug, Default)]
+#[repr(C)]
+struct DiscardWriteZeroesData {
```
andreeaflorescu, line 54: nit: we can drop `Data` from the name so that it maps to the virtio spec name better.

Suggestion: how about we move the definition of `VIRTIO_BLK_WRITE_ZEROES_FLAG_UNMAP` to the implementation of this structure? This way we can shorten the name quite a bit and just call it `UNMAP_BIT`, and the definition is also put in context.

andreeaflorescu (commented):
WIP. Mostly nits so far.

src/block/defines.rs
andreeaflorescu: Nvm, they're different.

.buildkite/custom-pipeline.yml
lauralt: Sure, opened #37 for that.

src/block/stdio_executor.rs
```diff
@@ -202,28 +241,118 @@ impl<B: Backend> StdIoBackend<B> {
...
+        if request_type == RequestType::Discard {
+            // Since Discard is just a hint and some filesystems may not implement
+            // FALLOC_FL_PUNCH_HOLE, ignore punch_hole() errors.
+            let _ = self.inner.punch_hole(offset, length);
+        } else {
+            // If unmap is set, try at first to punch a hole, if it fails, fall back to just
+            // writing zeroes.
+            // After a write zeroes command is completed, reads of the specified ranges of sectors
+            // MUST return zeroes, independent of unmap value.
+            if flags & VIRTIO_BLK_WRITE_ZEROES_FLAG_UNMAP != 0 {
```
alexandruag, line 416: Hmm, can we replace this `if/else` statement with something like:

```rust
if flags & VIRTIO_BLK_WRITE_ZEROES_FLAG_UNMAP == 0 
|| self.inner.punch_hole(offset, length).is_err() {
    self.inner
        .write_all_zeroes_at(offset, length as usize)      
        .map_err(Error::DiscardWriteZeroes)?;    
}
Ok(0)
```
Kinda wanted to avoid repeating the `write_all_zeroes` call, but pls ignore if it actually reduces clarity.

src/block/stdio_executor.rs
```diff
@@ -202,28 +241,118 @@ impl<B: Backend> StdIoBackend<B> {
...
+                if self.inner.punch_hole(offset, length).is_err() {
```
alexandruag, line 417: Hmm, according to the standard (version 1.1, section 5.2.6.2), there's some interaction between discarding data on `WRITE_ZEROES` and the `write_zeroes_may_unmap` field in the configuration space. Does it make sense to add a configuration option to the executor which determines whether we attempt a discard first, or simply write zeroes every time regardless?

src/block/defines.rs
```diff
@@ -9,6 +9,8 @@ pub const VIRTIO_BLK_T_IN: u32 = 0;
 pub const VIRTIO_BLK_T_OUT: u32 = 1;
 /// Flush request.
 pub const VIRTIO_BLK_T_FLUSH: u32 = 4;
+/// Get device ID request.
+pub const VIRTIO_BLK_T_GET_ID: u32 = 8;
```
alexandruag, line 13: Hmm, just wondering, where do the added values come from? (it looks like they're not in the 1.1 standard). Are they about to be to added to a subsequent version? We could also say a few words about this to improve clarity.

src/block/stdio_executor.rs
```diff
@@ -140,7 +143,8 @@ impl<B: Backend> StdIoBackend<B> {
     /// # Arguments
     /// * `inner` - The block device backend.
     /// * `features` - The features that were negotiated between driver and device.
-    pub fn new(mut inner: B, features: u64) -> Result<Self> {
+    /// * `device_id` - The block device id.
+    pub fn new(mut inner: B, features: u64, device_id: [u8; VIRTIO_BLK_ID_BYTES]) -> Result<Self> {
```
alexandruag, line 188: Does the`GET_ID` operation always have to be supported? If not, does it make sense to have the `device_id` field as an `Option<...>` and return unsupported when it's `None`? In that case, we could also add a separate constructor (something like `with_device_id`), and keep the old `new` signature.

src/block/request.rs
```diff
@@ -154,8 +154,10 @@ impl Request {
     }
 
     /// Returns the total length of request data.
-    pub fn total_data_len(&self) -> u32 {
-        self.data.iter().map(|&x| x.1).sum()
+    pub fn total_data_len(&self) -> u64 {
+        // The maximum queue size is 32768 (2^15), which is the maximum  possible descriptor chain
+        // length and since data length is an u32, this sum can not overflow an u64.
+        self.data.iter().map(|&x| x.1 as u64).sum()
```
alexandruag, line 160: Just wondering, using does using `|x|` instead of `|&x|` work here?

src/block/stdio_executor.rs
```diff
@@ -233,16 +241,16 @@ impl<B: Backend> StdIoBackend<B> {
                 for (data_addr, data_len) in request.data() {
                     mem.read_exact_from(*data_addr, &mut self.inner, *data_len as usize)
                         .map_err(Error::Read)?;
-                    bytes_from_dev += data_len;
+                    bytes_from_dev = bytes_from_dev
+                        .checked_add(*data_len)
+                        .ok_or_else(|| Error::Overflow)?;
```
alexandruag, line 307: I think we can use `.ok_or(Error::Overflow)` directly here instead of `ok_or_else`.

src/block/stdio_executor.rs
```diff
@@ -117,6 +137,21 @@ impl Display for Error {
     }
 }
 
+/// Errors encountered while processing a request execution result.
+#[derive(Debug)]
+pub enum ProcessReqError {
```
alexandruag, line 142: Can we use the `Error` type for `process_request` errors instead of defining a new one? The variants seem to be already present (and we could have added new ones if they weren't).

alexandruag (commented):
Did an initial pass and left some comments, but looks good overall! I'll have another go to take a better look at the tests as well.

src/block/defines.rs
lauralt: It's not the position, but the least significant bit set. I'll make the comment clearer.

src/block/stdio_executor.rs
lauralt: done

src/block/defines.rs
lauralt: Moved the definition to `DiscardWriteZeroes` and hopefully, updated with a clearer comment.

src/block/stdio_executor.rs
lauralt: Hmmm, that would mean we would have to associate a status with that variant and I think this makes sense only for errors thrown during request execution (not, for example, for an error that is thrown if we fail at writing the status to memory). I would keep a separate enum here, even though it happens that the variants are identical to some of `Error`. Does it make sense?

src/block/stdio_executor.rs
lauralt: I'll think a bit more about this, the specification is not the clearest regarding this aspect. Also if we decide to add/use this field, I guess we should also add the other ones related to discard, write zeroes, e.g. `max_discard_sectors`.

src/block/stdio_executor.rs
```diff
@@ -202,28 +250,104 @@ impl<B: Backend> StdIoBackend<B> {
...
             RequestType::Flush => return self.inner.fsync().map(|_| 0).map_err(Error::Flush),
+            RequestType::Discard | RequestType::WriteZeroes => {
+                if total_len % DiscardWriteZeroes::LEN != 0 {
+                    return Err(Error::InvalidDataLength);
+                }
+                for (data_addr, data_len) in request.data() {
+                    let mut available_bytes = *data_len as u64;
+                    let mut crt_addr = *data_addr;
+                    
+                    while available_bytes >= DiscardWriteZeroes::LEN {
```
alexandruag, line 368: Until we add support to easily read data across consecutive descriptors, it looks like we might end up interpreting things a bit erroneously here if a `DiscardWriteZeros` object starts at the end of one descriptor and spills over the next. How about we replace the `if total_len % DiscardWriteZeroes::LEN != 0` block above with a similar check for every `data_len` in this `for` loop? If `data_len` is always a multiple of the object len (which AFAIK happens for all driver implementations we're aware of), then we know for sure all `DiscardWriteZeros` structs found in a descriptor are complete.

lauralt: :+1: makes sense, done

alexandruag (approved):
Looks good, just left a minor comment.

src/block/stdio_executor.rs
```diff
@@ -136,9 +210,19 @@ impl<B: Backend> StdIoBackend<B> {
...
             num_sectors: disk_size >> SECTOR_SHIFT,
             features,
+            device_id: None,
         })
     }
 
+    /// Sets the `device_id`.
+    ///
+    /// # Arguments
+    /// * `device_id` - The block device id.
```
Mic92, line 220: Might be helpful?

```suggestion
    /// * `device_id` - The block device id. On Linux guests one can read this information from /sys/block/<device>/serial
```

src/block/stdio_executor.rs
```diff
@@ -192,45 +318,188 @@ impl<B: Backend> StdIoBackend<B> {
...
+                // requests.
+                if total_len != VIRTIO_BLK_ID_BYTES as u64 {
+                    return Err(Error::InvalidDataLength);
+                }
+                for (data_addr, data_len) in request.data() {
+                    // The device_id accesses are safe because we checked that the total data length
+                    // is VIRTIO_BLK_ID_BYTES, which is the size of the id as well.
+                    mem.read_exact_from(
+                        *data_addr,
+                        &mut &device_id[bytes_to_mem as usize..(*data_len + bytes_to_mem) as usize],
```
Mic92, line 369: I used a cursor here but a slice range should be also fine: https://github.com/rust-vmm/vm-virtio/pull/56/files#diff-ae9f3a857e7d373fd75ef4481f3d920ae3416de07e3a67c43fe8241443cf5fdfR220
The error handling looks more complex. I have not looked into it if that is necessary but probably you di.

lauralt: I am exaggerating a bit here indeed. It is probably impossible to have a partial read when the max size of access is 20 bytes, so we could probably remove the match below. My preference would be to keep the check though.

alexandruag (approved):

andreeaflorescu (approved):

PR #38, opened 2020-12-29
split Queue into receiver and sender to support multi-thread environment

jiangliu:
The virtio queue is duplex comunication channel, it receives available descriptors from the driver and sends used descriptor to the driver. So the queue could be split into a receive endpoint and a sender endpoint.
 
With split queue, the receiver endpoint receives available descriptors from driver by accessing the descriptor table and available ring. It also updates device event suppression status when receiving available descriptors. For the sender endpoint, it puts used descriptors into the used ring. It also queries driver event suppression status when sending used descriptors.
    
For packed queue, it includes a descriptor table, a driver event suppression block and a device event suppression block. So packed queue could be abstracted in the same way as split queue.
    
Current queue implementation is optimzed for single thread environment, there a single thread will receive available descriptors, handle these descriptors and then send those used descriptors in order. This mode will limit the performance of high IO virtio backends, such as virtiofs.
    
So split the queue into receiver and sender. Due to implementation complex, the receiver still only supports single thread environment. For the sender endpoint, both a single thread version and a multiple thread version are provided. For the multi-thread version, it could be send between threads, and the sender and the receiver could work concurrently.

This is also a preparation for support rust async/await and packed queue.

### DIFF
--- a/benches/queue/mod.rs
+++ b/benches/queue/mod.rs
@@ -5,7 +5,7 @@ mod mock;
 
 use criterion::{black_box, BatchSize, Criterion};
 use vm_memory::{GuestAddress, GuestAddressSpace, GuestMemoryAtomic, GuestMemoryMmap};
-use vm_virtio::Queue;
+use vm_virtio::{Queue, QueueSinkerT, QueueSourceT};
 
 use mock::MockSplitQueue;
 
--- a/coverage_config_x86_64.json
+++ b/coverage_config_x86_64.json
@@ -1,5 +1,5 @@
 {
-  "coverage_score": 88.8,
+  "coverage_score": 89.2,
   "exclude_path": "",
   "crate_features": "backend-stdio"
 }
--- a/src/block/request.rs
+++ b/src/block/request.rs
@@ -253,7 +253,7 @@ mod tests {
     use vm_memory::{Address, GuestMemoryMmap};
 
     use crate::queue::tests::VirtQueue;
-    use crate::{VIRTQ_DESC_F_NEXT, VIRTQ_DESC_F_WRITE};
+    use crate::{QueueSourceT, VIRTQ_DESC_F_NEXT, VIRTQ_DESC_F_WRITE};
 
     impl PartialEq for Error {
         fn eq(&self, other: &Self) -> bool {
--- a/src/device/mmio.rs
+++ b/src/device/mmio.rs
@@ -12,7 +12,7 @@ use std::sync::atomic::Ordering;
 use vm_memory::{GuestAddress, GuestAddressSpace};
 
 use crate::device::{status, WithDriverSelect};
-use crate::Queue;
+use crate::{Queue, QueueConfigT};
 
 // Required by the Virtio MMIO device register layout at offset 0 from base. Turns out this
 // is actually the ASCII sequence for "virt" (in little endian ordering).
--- a/src/device/mod.rs
+++ b/src/device/mod.rs
@@ -17,7 +17,7 @@ use std::result;
 use std::sync::atomic::AtomicU8;
 use std::sync::Arc;
 
-use crate::Queue;
+use crate::{Queue, QueueConfigT};
 
 pub use mmio::VirtioMmioDevice;
 pub use virtio_config::{VirtioConfig, VirtioDeviceActions, VirtioDeviceType};
--- a/src/device/virtio_config.rs
+++ b/src/device/virtio_config.rs
@@ -11,7 +11,7 @@ use std::sync::Arc;
 use vm_memory::GuestAddressSpace;
 
 use crate::device::{VirtioDevice, WithDriverSelect};
-use crate::Queue;
+use crate::{Queue, QueueConfigT};
 
 /// An object that provides a common virtio device configuration representation. It is not part
 /// of the main `vm-virtio` set of interfaces, but rather can be used as a helper object in
--- a/src/queue.rs
+++ b/src/queue.rs
@@ -77,38 +77,42 @@ impl Display for Error {
 
 impl std::error::Error for Error {}
 
-/// A virtio descriptor constraints with C representation
+/// A split virtio descriptor to encapsulate a driver data buffer.
+///
+/// The descriptor table refers to the buffers the driver is using for the device. The `addr` is
+/// a physical address, and the buffers can be chained via next. Each descriptor describes a
+/// buffer which is read-only for the device (“device-readable”) or write-only for the device
+/// (“device-writable”), but a chain of descriptors can contain both device-readable and
+/// device-writable buffers.
 #[repr(C)]
 #[derive(Default, Clone, Copy)]
 pub struct Descriptor {
-    /// Guest physical address of device specific data
+    /// Guest physical address of the data buffer.
     addr: u64,
 
-    /// Length of device specific data
+    /// Length of the data buffer.
     len: u32,
 
-    /// Includes next, write, and indirect bits
+    /// Data buffer flags, including the next, write, and indirect bits.
     flags: u16,
 
-    /// Index into the descriptor table of the next descriptor if flags has
-    /// the next bit set
+    /// Index into the descriptor table of the next descriptor if flags has the next bit set.
     next: u16,
 }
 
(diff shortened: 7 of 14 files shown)


alexandruag:
Hi Gerry! I've started looking at this, and I was wondering if you have a quick example or high level description of how multiple threads can process a queue in parallel using the new functionality (i.e. what happens after the driver sends a notification, how is work distributed between workers, and how do the descriptor chains end up in the used ring). This would complement the PR nicely, and helps thinking about this class of use cases.

jiangliu:
> Hi Gerry! I've started looking at this, and I was wondering if you have a quick example or high level description of how multiple threads can process a queue in parallel using the new functionality (i.e. what happens after the driver sends a notification, how is work distributed between workers, and how do the descriptor chains end up in the used ring). This would complement the PR nicely, and helps thinking about this class of use cases.

We started to enable multi-threaded support of virtio queue, but it turns out to be a much more fundamental refactor.
   Several traits are introduced to mudularize the queue implementation:
    - QueueSource: interfaces to consume available descriptors from Queue
    - QueueSourceEndpoint: internal trait to implement QueueSource
    - QueueSink: interfaces to put used descriptors into Queue
    - QueueSinkEndpoint: internal trait to implement QueueSink
    - QueueConfig: configure and get configuraiton about Queue
    - QueueT: QueueConfig + QueueSource + QueneSink
    
    With this changes, we could hide `GuestAddressSpace` from device and
    block subsystems, which now takes a `Q: QueueT` generic parameter
    instead of `M: GuestAddressSpace`.

I found the PR has heavy conflicts with #33, and @andreeaflorescu is proposing to convert vm-virtio to a workspace #44 ,
so I have merged the main idea from #33 to split the queue.rs into submodules.  It may also interfere with #41 because both introduce VirtioConfig:)

alexandruag:
Hi Gerry, apologies for taking longer than expected to wrap up my initial review. Also, thanks for the detailed tests and comments; they really helped understand the model you're building better. Given the complexity of the changes, it looks like a good next step is to see if we can start with some simplifications, and add things on top if it turns out they're not actually sufficient (they also help me validate I actually got some of these things right :D).

With the current state of upstream `vm-virtio`, the most straightforward way of interacting with the same queue from multiple threads is to wrap it in an `Arc<Mutex<>>` and pass that around. This is probably not the most flexible option, as both the "available" and "used" parts have to contend for taking the same lock. For example, in the `test_multi_thread_source_sink` test you've added, there's a `Mutex` for the "available" part, and another one for "used" to avoid that kind of contention I would assume. A interesting question in this context is whether it makes sense to simply use two clones of the same `Queue` instead of actually splitting `Queue` into separate sink and source structures. Going back to your test, instead of:

```rust
let source = Arc::new(Mutex::new(AvailRing::new(m.clone(), vq.size())));
let sink = Arc::new(Mutex::new(UsedRing::new(m.clone(), vq.size())));
``` 

we'd have something like


```rust
let queue = // fully set up new queue ...
let source = Arc::new(Mutex::new(queue.clone());
let sink = Arc::new(Mutex::new(queue));

```

From a high level, it looks like this should work, as the sink and source are orthogonal in terms of functionality. If so, a related question is whether `next_avail` needs to become a `Position` (vs remaining a `Wrapping<u16>`), because it also seems like we can make the sink functionality independent of that field.

Using this approach means that the actual state of the `Queue` gets split between the two clones, but ultimately the same thing happens when dividing the `Queue` into separate source and sink objects in this proposal. I think one potential concern can be that, when passing the clones around, they support the full functionality of the Queue (and not just sink or just source). However, if this is an issue, we can define the sink and source traits, implement them for the entire `Queue` object, and use each clone behind an appropriate trait bound, as opposed to the full current interface.

I know the above questions are quite high level and omit some details, but I was really curious to get your thoughts around starting with a simplified approach, before moving forward with the discussion.

jiangliu:
> Hi Gerry, apologies for taking longer than expected to wrap up my initial review. Also, thanks for the detailed tests and comments; they really helped understand the model you're building better. Given the complexity of the changes, it looks like a good next step is to see if we can start with some simplifications, and add things on top if it turns out they're not actually sufficient (they also help me validate I actually got some of these things right :D).
> 
> With the current state of upstream `vm-virtio`, the most straightforward way of interacting with the same queue from multiple threads is to wrap it in an `Arc<Mutex<>>` and pass that around. This is probably not the most flexible option, as both the "available" and "used" parts have to contend for taking the same lock. For example, in the `test_multi_thread_source_sink` test you've added, there's a `Mutex` for the "available" part, and another one for "used" to avoid that kind of contention I would assume. A interesting question in this context is whether it makes sense to simply use two clones of the same `Queue` instead of actually splitting `Queue` into separate sink and source structures. Going back to your test, instead of:
> 
> ```rust
> let source = Arc::new(Mutex::new(AvailRing::new(m.clone(), vq.size())));
> let sink = Arc::new(Mutex::new(UsedRing::new(m.clone(), vq.size())));
> ```
> 
> we'd have something like
> 
> ```rust
> let queue = // fully set up new queue ...
> let source = Arc::new(Mutex::new(queue.clone());
> let sink = Arc::new(Mutex::new(queue));
> ```
> 

This is really an interesting observersion:)
Previously I think  struct UsedRing needs to share AvailRing::next_avail, but after a double checking it turns out to be unnecessary.
But when implementing packed queue, some tricky is needed to work out this way.

jiangliu:
> Hi Gerry, apologies for taking longer than expected to wrap up my initial review. Also, thanks for the detailed tests and comments; they really helped understand the model you're building better. Given the complexity of the changes, it looks like a good next step is to see if we can start with some simplifications, and add things on top if it turns out they're not actually sufficient (they also help me validate I actually got some of these things right :D).
> 
> With the current state of upstream `vm-virtio`, the most straightforward way of interacting with the same queue from multiple threads is to wrap it in an `Arc<Mutex<>>` and pass that around. This is probably not the most flexible option, as both the "available" and "used" parts have to contend for taking the same lock. For example, in the `test_multi_thread_source_sink` test you've added, there's a `Mutex` for the "available" part, and another one for "used" to avoid that kind of contention I would assume. A interesting question in this context is whether it makes sense to simply use two clones of the same `Queue` instead of actually splitting `Queue` into separate sink and source structures. Going back to your test, instead of:
> 
> ```rust
> let source = Arc::new(Mutex::new(AvailRing::new(m.clone(), vq.size())));
> let sink = Arc::new(Mutex::new(UsedRing::new(m.clone(), vq.size())));
> ```
> 
> we'd have something like
> 
> ```rust
> let queue = // fully set up new queue ...
> let source = Arc::new(Mutex::new(queue.clone());
> let sink = Arc::new(Mutex::new(queue));
> ```
> 
> From a high level, it looks like this should work, as the sink and source are orthogonal in terms of functionality. If so, a related question is whether `next_avail` needs to become a `Position` (vs remaining a `Wrapping<u16>`), because it also seems like we can make the sink functionality independent of that field.
> 
> Using this approach means that the actual state of the `Queue` gets split between the two clones, but ultimately the same thing happens when dividing the `Queue` into separate source and sink objects in this proposal. I think one potential concern can be that, when passing the clones around, they support the full functionality of the Queue (and not just sink or just source). However, if this is an issue, we can define the sink and source traits, implement them for the entire `Queue` object, and use each clone behind an appropriate trait bound, as opposed to the full current interface.
> 
> I know the above questions are quite high level and omit some details, but I was really curious to get your thoughts around starting with a simplified approach, before moving forward with the discussion.

I have tried to follow the way suggested above, but failed.
A queue conceptually includes three parts: configuration, source and sink.
```
pub trait QueueT: QueueConfig + QueueSource + QueueSink
```
If we clone the queue, one for source, and another for sink, it will cause trouble to the configuration part.

And I have heavily refactored the abstraction, now it's much more clear. Now the main interface traits and the relationship among them is as below:
```
QueueT: QueueConfig + QueueSource + QueueSink;
QueueConfig: QueueSourceConfig + QueueSinkConfig
```

jiangliu:
@alexandruag Hi, I have tried another round to follow the way suggested by you, and succeeded this time. As expected, the final change is much more clear and simple. Thanks for you suggestions:)

Issue #6, opened 2019-06-05
Support of Packed Virtque defined in virtio spec 1.1


jiangliu:
The virtio spec 1.1 defines a virtque format named "Packed Virtque", which may help to improve virtio device performance.

stefano-garzarella:
Done in #310

PR #39, opened 2020-12-30
RFC: extend Queue to support packed queue

jiangliu:
Fixes： #6 
This patch set is based on #38 , and aims to implement an PoC by extending current Queue implementation to support Packed Queue. Current design shares most part with split queue, and use flag to control support of packed. 

Currently it only implements basic support of packed to receive available descriptors and send used descriptors. More effort is needed to support device/driver event suppress mechanism.

### DIFF
--- a/benches/queue/mock.rs
+++ b/benches/queue/mock.rs
@@ -260,11 +260,11 @@ impl<'a, M: GuestMemory> MockSplitQueue<'a, M> {
         self.update_avail_idx(head_idx);
     }
 
-    pub fn create_queue<A: GuestAddressSpace>(&self, a: A) -> Queue<A> {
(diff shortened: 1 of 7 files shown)


serban300:
Hi @jiangliu ! I was wondering if you had the chance to run any performance tests on the packed queues implementation and if you noticed any significant improvement compared to split queues.

jiangliu:
> Hi @jiangliu ! I was wondering if you had the chance to run any performance tests on the packed queues implementation and if you noticed any significant improvement compared to split queues.

I have no performance data yet.
To be honest, after implementing the packed queue, I feel it's more friendly to hardware implemented virtio devices. For software emulated virtio device, I feel the packed queue may not help too much. So I'm a little hesitate whether we should implement packed queue.

PR #41, opened 2021-01-03
[RFC] Additional queue interface tweaks

alexandruag:
This PR introduces two sets of changes (besides renaming some fields to correspond with the Virtio 1.1 standard terminology):

- Based on the first iteration and discussions in #27, we add a new `QueueConfig` object that holds the configuration of a queue (sans the memory object), and then `Queue` itself has a new generic type parameter `C` that allows borrowing a `QueueConfig`. For example, we can now remove the `M` dependency from `VirtioDevice` and related abstractions, as well as disconnect the queue configuration from the memory object in general (and recombine them later, or simply create ephemeral `Queue`s that are built using references) when necessary. The extra type parameter can be hidden with aliases; for example `type Queue<M> = vm_virtio::Queue<M, QueueConfig>` is equivalent to the old definition.

- The last commit introduces several tweaks to the `Queue` interface. First, methods that don't necessarily have to return an error (such as the initial incarnation of `needs_notification`) are no longer using `Result` as their return types. I've been experimenting in a couple of settings, and things look clearer without extra error handling. Second, there's a new `Queue::pop_descriptor_chain` method. The `AvailIter` object has to mutably borrow the associated `Queue`, which means we cannot call other `Queue` methods in a block such as `for chain in queue.iter() { /* ... */ queue.add_used(...) }`. The alternative is to pop and process one descriptor chain at a time.

The changes in the RFC are on top of an older master branch. I'll rebase them and update the `VirtioDevice` interfaces as well if things look good so far.

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -46,7 +46,7 @@ pub trait VirtioDevice<M: GuestAddressSpace>: Send {
         mem: M,
         interrupt_evt: EventFd,
         status: Arc<AtomicUsize>,
-        queues: Vec<Queue<M>>,
+        queues: QueueConfig,
         queue_evts: Vec<EventFd>,
     ) -> ActivateResult;
 
--- a/src/queue.rs
+++ b/src/queue.rs
@@ -10,6 +10,7 @@
 //
 // SPDX-License-Identifier: Apache-2.0 AND BSD-3-Clause
 
+use std::borrow::{Borrow, BorrowMut};
 use std::cmp::min;
 use std::fmt::{self, Display};
 use std::mem::size_of;
@@ -366,11 +367,9 @@ impl VirtqUsedElem {
 
 unsafe impl ByteValued for VirtqUsedElem {}
 
-#[derive(Clone)]
 /// A virtio queue's parameters.
-pub struct Queue<M: GuestAddressSpace> {
-    mem: M,
-
+#[derive(Clone, Default)]
+pub struct QueueConfig {
     /// The maximal size in elements offered by the device
     max_size: u16,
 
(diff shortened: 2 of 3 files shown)


glitzflitz:
@alexandruag Can you add a function to check whether queue is empty or not? You need it for checking whether eventq is empty for virtio scsi

alexandruag:
> @alexandruag Can you add a function to check whether queue is empty or not? You need it for checking whether eventq is empty for virtio scsi

Hi! You mean checking whether one or more descriptor chains are available without also popping the next one and incrementing the `next_avail` field? Does the use case you mention require knowledge of available descriptor chains without also handling any of them? That can def be added. We're currently trying to help reviewing the other ongoing PRs in `vm-virtio` and `vm-memory`, but will circle back here soon.

glitzflitz:
@alexandruag For the use case of eventq in virtio scsi one just needs to check whether there are any descriptor chains in eventq so I guess checking if `pop_descriptor_chain` returns `None` would work

alexandruag:
@glitzflitz That does work, but it also advances the available ring one position (if a descriptor chain is available), so if you only want to detect availability but not do anything with the chain at that moment, you'd have to call `go_to_previous_position` as well if `pop_descriptor_chain` returned `Some(chain)`.

jiangliu:
> @glitzflitz That does work, but it also advances the available ring one position (if a descriptor chain is available), so if you only want to detect availability but not do anything with the chain at that moment, you'd have to call `go_to_previous_position` as well if `pop_descriptor_chain` returned `Some(chain)`.

Maybe some sort of "peek" operation is desired. It's hard to implement go_to_previous_position() for packed queue.

glitzflitz:
> Maybe some sort of "peek" operation is desired. It's hard to implement go_to_previous_position() for packed queue.

Yes the "peek" was term I was looking for. @alexandruag I think peek based functions would be great to have

src/queue.rs
```diff
@@ -510,88 +513,51 @@ impl<M: GuestAddressSpace> Queue<M> {
...
             .memory()
-            .store(val, self.device_area, order)
+            .store(val, self.config().device_area, order)
             .map_err(Error::GuestMemory)
     }
 
     // Write the appropriate values to enable or disable notifications from the driver. Every
     // access in this method uses `Relaxed` ordering because a fence is added by the caller
     // when appropriate.
-    fn set_notification(&mut self, enable: bool) -> Result<(), Error> {
```
bonzini, line 581: I think these were `mut` by design. In principle they don't need mutable access to `self` because the fields themselves employ interior mutability; however, we don't want multiple threads to muck with a single queue without external protection. The usage of atomic primitives works fine if there is parallel access from the guest, but not if other threads in the VMM call these methods concurrently.

src/queue.rs
```diff
@@ -510,88 +513,51 @@ impl<M: GuestAddressSpace> Queue<M> {
...
-            self.next_used.0,
-            self.device_area.unchecked_add(2),
-            Ordering::Release,
-        )
-        .map_err(Error::GuestMemory)
-    }
-
     // Helper method that writes `val` to the `avail_event` field of the used ring, using
     // the provided ordering.
     fn set_avail_event(&self, val: u16, order: Ordering) -> Result<(), Error> {
```
bonzini, line 526: This in fact could/should probably be `mut` too.

src/device.rs
```diff
@@ -46,7 +46,7 @@ pub trait VirtioDevice<M: GuestAddressSpace>: Send {
         mem: M,
         interrupt_evt: EventFd,
         status: Arc<AtomicUsize>,
-        queues: Vec<Queue<M>>,
+        queues: QueueConfig,
```
bonzini, line 49: Shouldn't this be `Vec<QueueConfig>`?

src/queue.rs
```diff
@@ -399,70 +398,74 @@ pub struct Queue<M: GuestAddressSpace> {
...
+    pub fn new(mem: M, max_size: u16) -> Queue<M, QueueConfig> {
         Queue {
             mem,
-            max_size,
-            size: max_size,
-            ready: false,
-            desc_area: GuestAddress(0),
-            driver_area: GuestAddress(0),
-            device_area: GuestAddress(0),
-            next_avail: Wrapping(0),
```
bonzini, line 413: I am not sure that `ready`/`next_avail`/`next_used`/`event_idx_enabled`/`signalled_used` should be part of the `QueueConfig`. They are runtime state that should only be setup at `activate()` time; therefore it may belong in the `Queue`s that `activate()` creates.

src/device.rs
```diff
@@ -46,7 +46,7 @@ pub trait VirtioDevice<M: GuestAddressSpace>: Send {
         mem: M,
```
bonzini, line 46: It may make sense to remove the `mem` field from here, with the idea of passing it at construction time instead. `activate()` would forward or clone `self.mem` into the `Queue<M, C>` that it creates.

That would have the disadvantage of being an incompatible change, but it would be nice (and it would make sense) to remove the `GuestAddressSpace` argument from `VirtioDevice`. The `GuestAddressSpace` argument would still remain in the structs that implement the trait, e.g. `VirtioBlock`.

bonzini (changes requested):
The only "blocker" change is for the `s/mut self/self/` changes in `Queue`, which don't seem correct to me.

src/device.rs
alexandruag: The `VirtioDevice` interface (and implicitly, the signature of `activate`) has changed, but this RFC was not rebased on top of the latest version at the time.

src/queue.rs
alexandruag: The config struct (`QueueState` might have been a better name) attempts to capture all the characteristics of a `Queue` object that are independent of the memory handle. Further separating this into pre- and post-activate configuration is def possible, but ultimately we still need a way to reconstruct a `Queue` from the latter (for example, when restoring previously saved state), and it seemed more straightforward to keep everything together.

src/queue.rs
alexandruag: That makes sense (and it became even more apparent after Gerry started the discussion about handling queues via multiple threads). I'll make the necessary adjustments after thinking a bit more about the multi-thread angle as well.

alexandruag (commented):
Thanks for the review Paolo! I'll update the proposal and ping the reviewers afterwards.

src/queue.rs
bonzini: > ultimately we still need a way to reconstruct a Queue from the latter

Ok, that makes sense then. I thought it was just for the configuration as done by the guest, because it is passed to `activate`. Renaming to QueueState would then be a good idea.

PR #54, opened 2021-03-29
block: Use GuestMemory::checked_range() for validation

rbradford:
Using GuestMemory::checked_offset() will work for the majority of the
time however if the guest chooses to locate the data at the very end of
of the guest memory it will fail as the offset + offset will point to
the next byte (and thus the 0th byte outside the memory.)
GuestMemory::check_range() however takes an offset and length so will
handle this correctly.

Signed-off-by: Rob Bradford <robert.bradford@intel.com>

### DIFF
--- a/src/block/request.rs
+++ b/src/block/request.rs
@@ -164,11 +164,11 @@ impl Request {
 
         // Check that the address of the status descriptor is valid in guest memory.
         // We will write an u8 status here after executing the request.
-        let _ = mem
-            .checked_offset(desc.addr(), mem::size_of::<u8>())
-            .ok_or_else(|| {
-                Error::GuestMemory(GuestMemoryError::InvalidGuestAddress(desc.addr()))
-            })?;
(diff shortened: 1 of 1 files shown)


rbradford:
We just fixed a similar issue in cloud-hypervisor/cloud-hypervisor#2450 so I thought it helpful to check other code bases for similar issues.

lauralt:
Hi, the issue was already fixed in this commit: https://github.com/rust-vmm/vm-virtio/pull/35/commits/6f7f256e7ed32269ce7f3f4ad247de6435025e60 from https://github.com/rust-vmm/vm-virtio/pull/35 (I forgot to open an issue for visibility, sorry for that!). We've decided to not use `check_range()` for data descriptors since you can have holes in the range anyway, so the check is not necessary useful; I added an explanation in the commit message as well. Do you see things differently here?

rbradford:
Ok.

PR #56, opened 2021-05-24
implement get id block requests

Mic92:
If you agree to the interface, I can also add some tests.


alexandruag:
Hi! Looks like #35 (this [commit](https://github.com/rust-vmm/vm-virtio/pull/35/commits/92ca17f094267bc6d34cef0ad2b740e2783d3a5a) in particular) also adds support for the GET ID operation. Can you please have a look to see if it covers the functionality implemented in your PR as well?

Mic92:
I think the other PR should be also fine. I close this one, since I left my comments in the other PR.

PR #58, opened 2021-06-03
Convert repository to a cargo workspace

### LINKED ISSUE #44: Convert vm-virtio to a workspace
If we convert this repository to a workspace, then we can have all vm-virtio components published from the repositories. This allows us to test the interoperability of the various components (virtio queues, implementation of virtio-block/net/vsock/others), while still being able to publish them as individual crates. It should also make it easier to change the interfaces (up to the point where we actually get them settled in) because we can add all required changes in one PR.

krady21:
Resolves #44.

## Describe your changes:
* Moved `src/block/` to `crates/devices/virtio-blk` crate
* Moved `src/device` to `crates/virtio-device` crate
* Moved `src/queue.rs` to `crates/virtio-queue` crate

```
vm-virtio
+- crates
|   +- virtio-queue
|   +- virtio-device
|   +- devices
|   |   +- virtio-blk
|   |   +- virtio-net
```

I chose `crates` over `src` as it better suggests that the repository is organized as a workspace, but I can switch back to /src or a different name (virtio maybe?). 

Also which ones would you prefer?
`virtio-blk`, `virtio-queue`, etc or `vm-block`, `vm-queue`, etc
`devices/virtio-blk` or `device-types/virtio-blk`

## Help wanted:
At the moment, the only problem comes from the code that does some mocking in [queue.rs](https://github.com/rust-vmm/vm-virtio/blob/f63b67d90188186d116dba739c5db05916db296e/src/queue.rs#L721-L950) and is also imported in the [block device tests](https://github.com/rust-vmm/vm-virtio/blob/f63b67d90188186d116dba739c5db05916db296e/src/block/request.rs#L253). Block device tests aren't able to use `Descriptor::new()` because it was implemented inside the `#[cfg(tests)] pub mod tests` of a different crate (virtio-queue). I've looked into moving the mocking code into a different crate as suggested by @andreeaflorescu, but I got stuck when I couldn't implement `Descriptor` outside the crate where it was defined. My suggestion is to move the mocking code outside the tests module, until we come up with a better solution.


jiangliu:
How about putting those sub-crates under the vm-virtio root directory? Such as:
```
[workspace]
members = [
    "virtio-queue",
    "virtio-device",
    "devices/*",
]
```

krady21:
> How about putting those sub-crates under the vm-virtio root directory? Such as:
> 
> ```
> [workspace]
> members = [
>     "virtio-queue",
>     "virtio-device",
>     "devices/*",
> ]
> ```

That sounds good too!

andreeaflorescu:
@jiangliu I was thinking to have the code under the `src` folder instead:

```toml
[workspace]
members = [
    "src/virtio-queue",
    "src/virtio-device",
    "src/devices/*",
]
````

The reason is that if we are going to have multiple members in the workspace, it is going clutter the repository. We did this change for example in Firecracker, and it looks pretty tidy without all the rust-related files in root.

Since we do not yet have that many independent crates we can also just leave it as you suggested for now, and only move it under `src` if needed.

**Comment by krady21** [comment]
I moved the mocking code to a test_utils module as suggested. Everything should work now.

File: crates/virtio-queue/src/lib.rs [hunk 1]
```diff
@@ -9,6 +9,7 @@
 // Copyright (C) 2020 Alibaba Cloud. All rights reserved.
 //
 // SPDX-License-Identifier: Apache-2.0 AND BSD-3-Clause
+#![feature(stmt_expr_attributes)]
```
krady21, line 12: This allows me to use #[allow(deref_nullptr)] inside the macro definition, but is experimental. I added this to avoid getting warnings everytime offset_of! is used. I could delete this, if you consider it to be too dangerous. In any case, this shows that we should probably look for a rework of the offset_of! macro.

coverage_config_x86_64.json.sample
```diff
@@ -0,0 +1,5 @@
+{
+  "coverage_score": 90,
+  "exclude_path": "my_autogenerated_file.rs,second_file.rs",
```
andreeaflorescu, line 3: Hmm, these need to be updated.

krady21:
This PR is ready for review. The tests that are currently [failing](https://buildkite.com/rust-vmm/vm-virtio-ci/builds/198#ef8b80d0-e372-4991-a432-b02ccc8dc888) will pass as soon as the rust toolchain version is updated. I've ran them in the rust-vmm container with the latests rust stable (1.52.1).

andreeaflorescu:
> This PR is ready for review. The tests that are currently [failing](https://buildkite.com/rust-vmm/vm-virtio-ci/builds/198#ef8b80d0-e372-4991-a432-b02ccc8dc888) will pass as soon as the rust toolchain version is updated. I've ran them in the rust-vmm container with the latest rust stable (1.52.1).

The components are still under `crates`. Can we move them under root (suggested here: https://github.com/rust-vmm/vm-virtio/pull/58#issuecomment-854303881), or under `src` (suggested here: https://github.com/rust-vmm/vm-virtio/pull/58#issuecomment-854424428)?

coverage_config_x86_64.json
```diff
@@ -1,5 +1,5 @@
 {
-  "coverage_score": 87.7,
+  "coverage_score": 90,
   "exclude_path": "",
   "crate_features": "backend-stdio"
```
andreeaflorescu, line 4: Shouldn't this be `virtio-blk/backend-stdio`?

crates/virtio-queue/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+
+[dependencies]
+vm-memory = ">=0.4.0"
+vmm-sys-util = ">=0.4.0"
+log = ">=0.4.6"
+
+[dev-dependencies]
+criterion = "0.3.0"
+vm-memory = { version = ">=0.4.0", features = ["backend-mmap", "backend-atomic"] }
+virtio-queue = { path = ".", features = ["test-utils"] }
```
andreeaflorescu, line 23: Hmmm, this shouldn't be needed. Why does the crate depend on itself?

krady21: I was using it to always have the test-utils feature enabled on dev builds, but I think it is not necessary. I'll delete it.

coverage_config_x86_64.json
krady21: Yes. I'll change it immediately

krady21:
> The components are still under `crates`. Can we move them under root (suggested here: [#58 (comment)](https://github.com/rust-vmm/vm-virtio/pull/58#issuecomment-854303881)), or under `src` (suggested here: [#58 (comment)](https://github.com/rust-vmm/vm-virtio/pull/58#issuecomment-854424428))?

After thinking more about this, I am a bit reluctant to switch over to `src`, as that would lead to nested src paths (`src/virtio-queue/src`), which I've never been a huge fan of. I know a couple of other rust projects that use workspaces and follow a similar naming convention to what I used. ( See [ripgrep](https://github.com/BurntSushi/ripgrep) or [rust-analyzer](https://github.com/rust-analyzer/rust-analyzer) for example). Moving the crates under the root directory would not be a perfect solution either, as new crates could probably make it overcrowded. All in all, If I have to ultimately choose between one of the two approaches you've mentioned, I would probably go with what @jiangliu suggested, but I would still think that `crates` is the better choice.

jiangliu:
> > The components are still under `crates`. Can we move them under root (suggested here: [#58 (comment)](https://github.com/rust-vmm/vm-virtio/pull/58#issuecomment-854303881)), or under `src` (suggested here: [#58 (comment)](https://github.com/rust-vmm/vm-virtio/pull/58#issuecomment-854424428))?
> 
> After thinking more about this, I am a bit reluctant to switch over to `src`, as that would lead to nested src paths (`src/virtio-queue/src`), which I've never been a huge fan of. I know a couple of other rust projects that use workspaces and follow a similar naming convention to what I used ( see [ripgrep](https://github.com/BurntSushi/ripgrep) or [rust-analyzer](https://github.com/rust-analyzer/rust-analyzer) for example). Moving the crates under the root directory would not be a perfect solution either, as new crates could probably make it overcrowded. All in all, If I have to ultimately choose between one of the two approaches you've mentioned, I would probably go with what @jiangliu suggested, but I would still think that `crates` is the better choice in the long run.

It's only a suggestions, I'm OK with /crates too.

crates/virtio-queue/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+readme = "README.md"
+license = "Apache-2.0 OR MIT"
+edition = "2018"
+
+[features]
+test-utils = []
+
+[dependencies]
+vm-memory = ">=0.4.0"
+vmm-sys-util = ">=0.4.0"
```
lauralt, line 17: nit: Let's use the same `vmm-sys-util = ">=0.8.0"` as in the other crates.

lauralt (commented):
Looks good!

src/lib.rs
```diff
@@ -1,22 +0,0 @@
-// Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
-//
-// Portions Copyright 2017 The Chromium OS Authors. All rights reserved.
-// Use of this source code is governed by a BSD-style license that can be
-// found in the LICENSE-BSD-3-Clause file.
-//
-// SPDX-License-Identifier: Apache-2.0 AND BSD-3-Clause
-#![deny(missing_docs)]
```
krady21, line 8: Should this lint be added to all the new crates?

andreeaflorescu: Yes it should. Good catch!

crates/virtio-queue/Cargo.toml
krady21: Done!

src/lib.rs
krady21: Done

crates/devices/virtio-blk/src/lib.rs
```diff
@@ -1,11 +1,15 @@
-// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
-//
-// SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
+//! Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//!
+//! SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
```
lauralt, line 3: We shouldn't mark as documentation the copyright license. Instead, we should add a summary of the crate, e.g.

```rust
//! Provides abstractions for the virtio block device.
```
An example [here](https://github.com/rust-vmm/vm-virtio/blob/master/src/lib.rs) or [here](https://github.com/rust-vmm/vm-memory/blob/master/src/lib.rs).

crates/virtio-queue/benches/queue/mock.rs
```diff
@@ -156,10 +156,12 @@ impl<'a, M: GuestMemory> DescriptorTable<'a, M> {
         assert_eq!(indices.len(), len as usize);
 
         for (pos, index_value) in indices.iter().copied().enumerate() {
-            let mut desc = Descriptor::default();
             // Addresses and lens constant for now.
-            desc.addr = 0x1000;
-            desc.len = 0x1000;
+            let mut desc = Descriptor {
```
lauralt, line 160: nit: I would move the Rust update changes to a separate commit.

crates/devices/virtio-blk/src/lib.rs
krady21: Oops, thats silly from me. Thanks for finding it!

crates/virtio-queue/src/lib.rs
```diff
@@ -20,9 +24,14 @@ use vm_memory::{
     Address, ByteValued, Bytes, GuestAddress, GuestAddressSpace, GuestMemory, GuestMemoryError,
 };
 
-pub(super) const VIRTQ_DESC_F_NEXT: u16 = 0x1;
-pub(super) const VIRTQ_DESC_F_WRITE: u16 = 0x2;
-pub(super) const VIRTQ_DESC_F_INDIRECT: u16 = 0x4;
+use log::error;
+
+/// Marks a buffer as continuing via the next field
```
lauralt, line 29: We prefer to end the comments with full stop.

lauralt (approved):

crates/virtio-queue/src/lib.rs
krady21: Done

krady21:
@andreeaflorescu, @lauralt, @jiangliu, can you take one more look over this to see if we can get it merged?

andreeaflorescu (commented):
Can you add more details to the commit message that converts this to a workspace? One thing that seems important to mention and document is the use of the `test-utils` feature. We should write about why we decided to add this feature and what's the connection between it and the workspace.

andreeaflorescu:
LGTM.

.buildkite/custom-pipeline.yml
```diff
@@ -1,52 +1,52 @@
...
+      - cargo build --release --workspace --features virtio-blk/backend-stdio
     retry:
       automatic: false
     agents:
       platform: x86_64.metal
       os: linux
     plugins:
       - docker#v3.0.1:
-          image: "rustvmm/dev:v9"
+          image: "rustvmm/dev:v12"
```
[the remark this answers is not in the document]
krady21: Moved it as suggested.

lauralt (approved):

andreeaflorescu (approved):

slp (approved):

jiangliu (approved):

PR #65, opened 2021-06-16
Add some missing trait implementations.

Gaelan:
This PR:

* Derives `Debug` for a few types that were missing them. Also adds a manual implementation for `DescriptorChainRwIter`, because Rust doesn't seem to be able to generate the `M::T: Debug` constraint
* Adds `#![warn(missing_debug_implementations)]`, so we don't regress with missing Debug implementations in the future
* Derive Clone for DescriptorChainRwIter. This is useful in vhost-user-scsi: the virtio-scsi response includes some metadata at the beginning, so it's convenient to write the beginning of the response last. I accomplish this by cloning the DescriptorChainRwIter (or, actually, a higher-level wrapper), skipping forward past the header, writing the body; then using the original DescriptorChainRwIter to write the header.

### DIFF
--- a/src/block/stdio_executor.rs
+++ b/src/block/stdio_executor.rs
@@ -103,6 +103,7 @@ pub type Result<T> = result::Result<T, Error>;
 /// let file = File::create("foo.txt").unwrap();
 /// let request_exec = StdIoBackend::new(file, 1 << VIRTIO_BLK_F_FLUSH).unwrap();
 /// ```
+#[derive(Debug)]
 pub struct StdIoBackend<B: Backend> {
     /// The block device backing file.
     inner: B,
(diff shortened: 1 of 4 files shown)


slp:
Thanks for the PR. CI is complaining like this:

```
error: type does not implement `std::fmt::Debug`; consider adding `#[derive(Debug)]` or a manual implementation
   --> src/block/stdio_executor.rs:106:1
    |
106 | / pub struct StdIoBackend<B: Backend> {
107 | |     /// The block device backing file.
108 | |     inner: B,
109 | |     /// The number of sectors of `inner`.
...   |
112 | |     features: u64,
113 | | }
    | |_^
    |
    = note: `-D missing-debug-implementations` implied by `-D warnings`
```

This error can also be seen by running `cargo build --all-features`

Gaelan:
Whoops, thanks. Should be fixed now.

slp (approved):
LGTM, thanks!

lauralt (approved):

andreeaflorescu (approved):

PR #69, opened 2021-07-27
Move virtio-queue constant definitions to a dedicated module

krady21:
This is similar to virtio-blk, which already has a dedicated constant definitions module.
My only question is whether to make some of the constants `pub(crate)`, more specifically the ones that define sizes (VIRTQ_AVAIL_ELEMENT_SIZE, etc), as they are only used in virtio-queue crate at the moment.

### DIFF
--- a/crates/virtio-queue/src/defs.rs
+++ b/crates/virtio-queue/src/defs.rs
@@ -1,0 +1,43 @@
+// Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
+
+//! Virtio queue related constant definitions
+
+/// Marks a buffer as continuing via the next field.
+pub const VIRTQ_DESC_F_NEXT: u16 = 0x1;
+
+/// Marks a buffer as device write-only.
+pub const VIRTQ_DESC_F_WRITE: u16 = 0x2;
+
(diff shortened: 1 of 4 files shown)


lauralt (dismissed):

crates/virtio-queue/src/defs.rs
```diff
@@ -0,0 +1,39 @@
+//! Virtio queue related constant definitions
```
lauralt, line 1: Please add the copyright.

lauralt (commented):
> My only question is whether to make some of the constants `pub(crate)`, more specifically the ones that define sizes (VIRTQ_AVAIL_ELEMENT_SIZE, etc), as they are only used in virtio-queue crate at the moment.

Sounds good to me to have them as pub(crate).

crates/virtio-queue/src/defs.rs
```diff
@@ -0,0 +1,43 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
andreeaflorescu, line 1: 2021?

andreeaflorescu:
LGTM.

crates/virtio-queue/src/defs.rs
krady21: Done!

lauralt (approved):

andreeaflorescu (approved):

PR #71, opened 2021-08-03
Unify virtio-queue mocking code

### LINKED ISSUE #72: Unify mock logic
Currently, there are two slightly different mocking interfaces used for testing inside the virtio-queue crate. One used by the benchmark code inside `benches` and one inside the `virtio-queue/src/lib.rs`. The goal is to merge the two interfaces and decide which components should be kept, added or deleted.

krady21:
Closes #72.
This PR tries to merge the two interfaces that are used for mocking the driver side of virtio queue operations: the one used for unit tests and the one used for benchmarks.

### DIFF
--- a/crates/devices/virtio-blk/src/request.rs
+++ b/crates/devices/virtio-blk/src/request.rs
@@ -249,7 +249,7 @@ mod tests {
     use vm_memory::{Address, GuestMemoryMmap};
 
     use virtio_queue::defs::{VIRTQ_DESC_F_NEXT, VIRTQ_DESC_F_WRITE};
-    use virtio_queue::test_utils::VirtQueue;
+    use virtio_queue::mock::MockSplitQueue;
 
     impl PartialEq for Error {
         fn eq(&self, other: &Self) -> bool {
@@ -297,33 +297,31 @@ mod tests {
         descs: &[Descriptor],
     ) -> DescriptorChain<&'a GuestMemoryMmap> {
         // Support a max of 16 descriptors for now.
-        let vq = VirtQueue::new(GuestAddress(0), mem, 16);
+        let vq = MockSplitQueue::new(mem, 16);
         for (idx, desc) in descs.iter().enumerate() {
             let i = idx as u16;
-            vq.dtable(i).addr().store(desc.addr().0);
-            vq.dtable(i).len().store(desc.len());
-
-            if idx == descs.len() - 1 {
+            let addr = desc.addr().0;
+            let len = desc.len();
+            let (flags, next) = if idx == descs.len() - 1 {
                 // Clear the NEXT flag if it was set. The value of the next field of the
                 // Descriptor doesn't matter at this point.
-                vq.dtable(i)
-                    .flags()
-                    .store(desc.flags() & !VIRTQ_DESC_F_NEXT);
+                (desc.flags() & !VIRTQ_DESC_F_NEXT, 0)
             } else {
-                // Ensure the next flag is set.
-                vq.dtable(i).flags().store(desc.flags() | VIRTQ_DESC_F_NEXT);
-                // Ensure we are referring the following descriptor. This ignores
-                // any value is actually present in `desc.next`.
-                vq.dtable(i).next().store(i + 1);
-            }
+                // Ensure that the next flag is set and that we are referring the following
+                // descriptor. This ignores any value is actually present in `desc.next`.
+                (desc.flags() | VIRTQ_DESC_F_NEXT, i + 1)
+            };
+
+            let desc = Descriptor::new(addr, len, flags, next);
+            vq.desc_table().store(i, desc);
         }
 
         // Put the descriptor index 0 in the first available ring position.
-        mem.write_obj(0u16, vq.avail_start().unchecked_add(4))
+        mem.write_obj(0u16, vq.avail_addr().unchecked_add(4))
             .unwrap();
 
         // Set `avail_idx` to 1.
-        mem.write_obj(1u16, vq.avail_start().unchecked_add(2))
+        mem.write_obj(1u16, vq.avail_addr().unchecked_add(2))
             .unwrap();
 
         vq.create_queue(mem)
--- a/crates/virtio-queue/benches/queue/mod.rs
+++ b/crates/virtio-queue/benches/queue/mod.rs
@@ -1,13 +1,11 @@
 // Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
 
-mod mock;
-
 use criterion::{black_box, BatchSize, Criterion};
 use virtio_queue::Queue;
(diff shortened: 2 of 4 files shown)


jiangliu:
Some failures are caused the new vm-memory v0.6 release,  the PR #73 fixes them.

crates/virtio-queue/src/lib.rs
```diff
@@ -12,9 +12,10 @@
 
 //! A crate that exposes the virtio queue API.
 
-#![deny(missing_docs)]
+// #![deny(missing_docs)]
 
 pub mod defs;
+pub mod mock;
```
jiangliu, line 18: This sub module should be conditionally enabled for test only.

krady21: I wanted to add just a few more changes to `MockSplitQueue`. After that I'll squash some of the commits, add some documentation and I'll leave the module behind a dev-dependency enabled feature, as suggested.

crates/virtio-queue/src/mock.rs
```diff
@@ -0,0 +1,327 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
+
+//! Mocking
```
alexandruag, line 4: Would be nice to add a better description here.

crates/virtio-queue/src/mock.rs
```diff
@@ -0,0 +1,327 @@
...
+        let addr = self
+            .addr
+            .checked_add((index * size_of::<T>()) as u64)
+            .unwrap();
+
+        Ref::new(self.mem, addr)
+    }
+}
+
+// impl<'a, M: GuestMemory, T: ByteValued> Index<usize> for ArrayRef<'a, M, T> {
```
alexandruag: We can delete the commented stuff if we don't end up using it later. In particular, the index operator seems like it could have been nice here as a more concise way of calling `ref_at`. Is the commented implementation not working?

crates/virtio-queue/src/lib.rs
```diff
@@ -12,9 +12,10 @@
 
 //! A crate that exposes the virtio queue API.
 
-#![deny(missing_docs)]
+// #![deny(missing_docs)]
```
alexandruag, line 15: Is this commented out because the new module has public objects, methods, etc. which don't have doc comments?

crates/virtio-queue/src/lib.rs
alexandruag: Yeah it looks like some of the commits can (and should :D) be squashed. Also, I think we don't need to leave the module as a dev-dependency enabled feature really; we can have a feature that enables the module for external consumers, and then conditionally compile it with something `like #[cfg(any(test, feature="feature_name"))]` (haven't actually validated that's the correct syntax). That way, it's enabled for external consumers when the feature is selected, and when tests run within the local crate.

crates/devices/virtio-blk/src/request.rs
```diff
@@ -297,34 +297,26 @@ mod tests {
...
-                vq.dtable(i).flags().store(desc.flags() | VIRTQ_DESC_F_NEXT);
-                // Ensure we are referring the following descriptor. This ignores
-                // any value is actually present in `desc.next`.
-                vq.dtable(i).next().store(i + 1);
-            }
+                (desc.flags() | VIRTQ_DESC_F_NEXT, i + 1)
+            };
+
+            let desc = Descriptor::new(addr, len, flags, next);
+            vq.desc_table.store(i, desc);
```
alexandruag, line 312: Hmm, wondering whether we should use accessors (i.e.` vq.desc_table()`) instead of pub members, and also whether things are more straight forward if we add more helpers (i.e. something like the indexing operator for `DescriptorTable` for easier access to individual descriptors). Let's have a quick chat about this.

crates/virtio-queue/src/mock.rs
```diff
@@ -66,29 +67,24 @@ impl<'a, M: GuestMemory, T: ByteValued> ArrayRef<'a, M, T> {
...
-//     type Output = Ref<'a, M, T>;
-
-//     fn index(&self, index: usize) -> &Ref<'a, M, T> {
-//         let addr = self
-//             .addr
-//             .checked_add((index * size_of::<T>()) as u64)
-//             .unwrap();
+/// Extracts the displacement of a field in a struct
+#[macro_export]
+macro_rules! offset_of {
```
alexandruag, line 72: Is this macro still used anywhere?

crates/virtio-queue/benches/queue/mod.rs
```diff
@@ -40,7 +38,7 @@ pub fn benchmark_queue(c: &mut Criterion) {
     let mem = GuestMemoryMmap::<()>::from_ranges(&[(GuestAddress(0x0), 0x1_0000_0000)]).unwrap();
 
     let queue_with_chains = |num_chains, len, indirect| {
-        let mut mq = MockSplitQueue::new(&mem, 256);
+        let mut mq = MockSplitQueue::new(GuestAddress(0), &mem, 256);
```
alexandruag, line 41: Hmm, maybe it would be nicer and more convenient to have `new` always assume `GuestAddress(0)` as the start address, and have a separate constructor that allows us to specify a custom value.

alexandruag (commented):
Hi, apologies for the delay :( Left some comments and looking forward to the version with squashed commits that was mentioned in the conversation so far.

crates/virtio-queue/src/lib.rs
```diff
@@ -85,6 +88,16 @@ pub struct Descriptor {
 
 #[allow(clippy::len_without_is_empty)]
 impl Descriptor {
+    /// Creates a new descriptor
+    pub fn new(addr: u64, len: u32, flags: u16, next: u16) -> Self {
```
andreeaflorescu, line 92: This should only be enabled for tests and when the `test-utils` feature is used.

crates/virtio-queue/src/lib.rs
```diff
@@ -718,254 +731,20 @@ impl<M: GuestAddressSpace> Queue<M> {
     }
 }
 
-#[allow(missing_docs)]
-#[cfg(feature = "test-utils")]
-pub mod test_utils {
+#[cfg(test)]
+mod tests {
     use super::*;
+    use mock::*;
```
andreeaflorescu, line 737: It is a best practice to only import what you're using from the module. Right now you might be using everything from it, but as we add functionality to `mock` that will not be the case anymore. This is why it's better to be explicit in your imports.

andreeaflorescu (commented):
LGTM besides the 2 comments. I'll approve once those are fixed.

crates/virtio-queue/src/lib.rs
krady21: Done. Thank you for the tip.

crates/virtio-queue/benches/queue/mod.rs
krady21: Done.

andreeaflorescu (approved):

jiangliu (approved):

PR #74, opened 2021-08-18
virtio_queue: Add MSI vector field to the Queue

sboeuf:
In order to help with MSI/MSI-X support, it is convenient to store the
value of the interrupt vector inside the Queue directly.

### DIFF
--- a/crates/virtio-queue/src/defs.rs
+++ b/crates/virtio-queue/src/defs.rs
@@ -41,3 +41,6 @@ pub(crate) const VIRTQ_AVAIL_RING_META_SIZE: u64 = VIRTQ_AVAIL_RING_HEADER_SIZE 
 /// The Virtio Spec 1.0 defines the alignment of VirtIO descriptor is 16 bytes,
 /// which fulfills the explicit constraint of GuestMemory::read_obj().
 pub(crate) const VIRTQ_DESCRIPTOR_SIZE: usize = 16;
+
+/// Vector value used to disable MSI for a queue.
+pub const VIRTQ_MSI_NO_VECTOR: u16 = 0xffff;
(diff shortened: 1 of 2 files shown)


sboeuf:
PTAL @jiangliu @slp @andreeaflorescu

slp (approved):
LGTM, thanks!

jiangliu:
I'm still not sure whether it's a good idea to embedded virtio transport layer information into the virtio-queue. Interrupt related information should be assoicated with the device instead of with a queue. It would be better to associate an interrupt notifier with a queue instead of raw interrupt configuration information.

sboeuf:
@jiangliu just let me know when the crate is published to `crates.io` (meaning we reached some sort of stability on the API), I'll revisit the patches we need to make it work with Cloud Hypervisor.

sboeuf:
> I'm still not sure whether it's a good idea to embedded virtio transport layer information into the virtio-queue. Interrupt related information should be assoicated with the device instead of with a queue. It would be better to associate an interrupt notifier with a queue instead of raw interrupt configuration information.

Hi, I have some cycles to tackle this. So are we talking about having a trait `VirtioInterrupt` that would be passed as an `Option<>` to the `QueueState` structure? And this trait would declare something like `fn vector() -> Option<u16>` in order for the caller to retrieve the information about the vector in case it was setup?

Or are you thinking about passing something like `InterruptSourceGroup`. The problem with this approach is that `vm-virtio` would have a dependency on something like `vm-device` where the trait `InterruptSourceGroup` would be defined.

sboeuf:
This can be closed as I implemented it directly on Cloud Hypervisor, without the need for modifying the `QueueState` structure. (for reference: https://github.com/cloud-hypervisor/cloud-hypervisor/pull/3616)

PR #103, opened 2021-10-18
Fixes/various constant tweaks

stsquad:
As they were mentioned today I thought I'd have a quick run at these and check my rust-isms are correct.

### DIFF
--- a/crates/virtio-queue/src/defs.rs
+++ b/crates/virtio-queue/src/defs.rs
@@ -22,7 +22,7 @@ pub(crate) const VIRTQ_USED_ELEMENT_SIZE: u64 = 8;
 /// Used ring header: flags (u16) + idx (u16)
 pub(crate) const VIRTQ_USED_RING_HEADER_SIZE: u64 = 4;
 
-/// This is the size of the used ring metadata: header + used_event (u16).
+/// This is the size of the used ring metadata: header + avail_event (le16).
 /// The total size of the used ring is:
 /// VIRTQ_USED_RING_HMETA_SIZE + VIRTQ_USED_ELEMENT_SIZE * queue_size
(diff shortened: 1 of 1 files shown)


crates/virtio-queue/src/lib.rs
```diff
@@ -574,7 +575,8 @@ impl<M: GuestAddressSpace> QueueState<M> {
     // Helper method that writes `val` to the `avail_event` field of the used ring, using
     // the provided ordering.
     fn set_avail_event(&self, mem: &M::T, val: u16, order: Ordering) -> Result<(), Error> {
-        let offset = (4 + self.actual_size() * 8) as u64;
+        let elem_sz = VIRTQ_USED_ELEMENT_SIZE * u64::from(self.actual_size());
```
lauralt, line 579: Can you help with fixing these ones as well: https://github.com/rust-vmm/vm-virtio/blob/main/crates/virtio-queue/src/lib.rs#L626, https://github.com/rust-vmm/vm-virtio/blob/main/crates/virtio-queue/src/lib.rs#L790 so we can close the issue?
I think I would also define constants for [the alignment](https://github.com/rust-vmm/vm-virtio/blob/main/crates/virtio-queue/src/lib.rs#L696-L702) of the queue parts, e.g. `pub(crate) const DESC_TABLE_ALIGNMENT: u64 = 0Xf;`, but we can leave it for later as well.
Thanks!

stsquad: Sure. I did a cursory search for offset so missed those.

lauralt (approved):
Looks good, the style test is still failing, can you please run a `cargo fmt`?
The commit format test is still complaining about the length of this title: crates/virtio-queue: assert table_len fits in u16.

lauralt (approved):

Issue #94, opened 2021-10-12
move checks from Queue::is_valid to setters


lauralt:
There are a couple of things we are validating in [Queue::is_valid()](https://github.com/rust-vmm/vm-virtio/blob/main/crates/virtio-queue/src/lib.rs#L469) that are requirements according to the virtio spec. Calling `is_valid()` is optional, but we should always try to validate the MUST things from the spec, and not let the user be able to configure something that is not conforming to the spec.
We can move the mandatory checks from `is_valid` to the corresponding Queue setters, for example the following requirement [The driver MUST ensure that the physical address of the first byte of each virtqueue part is a multiple of the specified alignment value in the above table.](https://docs.oasis-open.org/virtio/virtio/v1.1/csprd01/virtio-v1.1-csprd01.html#x1-250001) can be validated when setting the address of these parts. At this point these validations are done [as part of `is_valid`](https://github.com/rust-vmm/vm-virtio/blob/main/crates/virtio-queue/src/lib.rs#L515-L524). There are also requirements for the queue size, e.g. `Queue Size value is always a power of 2.` which can be done in the `set_size()` method that is added in [this PR](https://github.com/rust-vmm/vm-virtio/pull/76).

PR #112, opened 2021-11-01
Moves specification-required checks to setters

germag:
Moves specification-required checks to setters

Moves the mandatory checks, according to the virtio specification,
from Queue::is_valid() to the corresponding Queue setters.
closes: #94

Signed-off-by: German Maglione <gmaglione@redhat.com>

### DIFF
--- a/crates/virtio-device/src/mmio.rs
+++ b/crates/virtio-device/src/mmio.rs
@@ -272,22 +272,22 @@ mod tests {
         assert_eq!(d.last_queue_notify, 2);
 
         assert_eq!(d.cfg.queues[0].state.desc_table.0, 0);
-        d.write(0x80, &1u32.to_le_bytes());
-        assert_eq!(d.cfg.queues[0].state.desc_table.0, 1);
+        d.write(0x80, &16u32.to_le_bytes());
+        assert_eq!(d.cfg.queues[0].state.desc_table.0, 16);
         d.write(0x84, &2u32.to_le_bytes());
-        assert_eq!(d.cfg.queues[0].state.desc_table.0, (2 << 32) + 1);
+        assert_eq!(d.cfg.queues[0].state.desc_table.0, (2 << 32) + 16);
 
         assert_eq!(d.cfg.queues[0].state.avail_ring.0, 0);
-        d.write(0x90, &1u32.to_le_bytes());
-        assert_eq!(d.cfg.queues[0].state.avail_ring.0, 1);
+        d.write(0x90, &2u32.to_le_bytes());
+        assert_eq!(d.cfg.queues[0].state.avail_ring.0, 2);
         d.write(0x94, &2u32.to_le_bytes());
-        assert_eq!(d.cfg.queues[0].state.avail_ring.0, (2 << 32) + 1);
+        assert_eq!(d.cfg.queues[0].state.avail_ring.0, (2 << 32) + 2);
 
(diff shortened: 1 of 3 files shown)


crates/virtio-device/src/mmio.rs
```diff
@@ -35,11 +35,14 @@ fn update_queue_field<M, D, F>(device: &mut D, f: F)
...
     M: GuestAddressSpace,
     D: WithDriverSelect<M> + ?Sized,
-    F: FnOnce(&mut Queue<M>),
+    F: FnOnce(&mut Queue<M>) -> Result<(), Error>,
 {
     if device.check_device_status(status::FEATURES_OK, status::DRIVER_OK | status::FAILED) {
         if let Some(queue) = device.selected_queue_mut() {
-            f(queue);
+            if let Err(e) = f(queue) {
+                device.set_device_status(status::FAILED);
```
lauralt, line 43: From the experiments that I did last week, setting the device status to FAILED doesn't seem to actually make a difference. Sorry for not coming with a response on the issue, but I had a chat only today with @andreeaflorescu and @alexandruag to decide on the path forward here, and we were thinking to keep the setter proposal, but in case of a misconfiguration to set the value of that field to a default. This seems to be how qemu handles this issue as well. So for the queue size, we would keep the max size if the value is invalid, and for the descriptor table address and the others, we would use address 0. What do you think?
I think with this approach, we could get rid of the `actual_size()` method as well, but I'll double-check.

crates/virtio-queue/src/lib.rs
```diff
@@ -735,12 +734,14 @@ impl<M: GuestAddressSpace> QueueStateT<M> for QueueState<M> {
         self.max_size
     }
 
-    fn actual_size(&self) -> u16 {
-        min(self.size, self.max_size)
-    }
+    fn set_size(&mut self, size: u16) -> Result<(), Error> {
```
lauralt, line 737: Sorry for not being explicit enough in my previous comment :(. The way we were thinking to address the issue is to not return an error anymore in case of a misconfiguration, and just ignore the value and log an error, something like:
```rust
    fn set_size(&mut self, size: u16) {
        if size > self.max_size() || size == 0 || (size & (size - 1)) != 0 {
            error!("virtio queue with invalid size: {}", size);
        }
    }
```
It's not necessary to explicitly set the value to the default one, because the `Queue` constructor is already initializing the size and addresses to the default ones. I would though define constants for the default addresses like you already did [here](https://github.com/rust-vmm/vm-virtio/pull/112/files#diff-32f28d703ff40356a906276eb71cdad39cc03aeacaf631c1afc157fc2cd7d641R34), but one for each queue part:
const DEFAULT_DESC_TABLE_ADDR: u64 = 0x0;
const DEFAULT_AVAIL_RING_ADDR: u64 = 0x0;
...

With this approach, we won't need to define other errors, and the logic [here](https://github.com/rust-vmm/vm-virtio/pull/112/files#diff-32f28d703ff40356a906276eb71cdad39cc03aeacaf631c1afc157fc2cd7d641R43) won't be needed anymore.

We should also update the negative tests for these setters, i.e. try to set an invalid value, and check that the value of size/addr remains the default one.

lauralt (approved):
Looks good! Thanks and sorry for the back and forth.

andreeaflorescu (approved):

germag:
> Looks good! Thanks and sorry for the back and forth.

No problem, I'm still learning, so thanks for your patience :)

Issue #95, opened 2021-10-12
validate requirements for indirect descriptors


lauralt:
This [TODO](https://github.com/rust-vmm/vm-virtio/blob/main/crates/virtio-queue/src/lib.rs#L125) should be addressed before publishing. We could do these checks in the [Iterator impl](https://github.com/rust-vmm/vm-virtio/blob/main/crates/virtio-queue/src/lib.rs#L243) right before [processing the indirect table](https://github.com/rust-vmm/vm-virtio/blob/main/crates/virtio-queue/src/lib.rs#L258) (or maybe find a better place?). Relevant spec section here: https://docs.oasis-open.org/virtio/virtio/v1.1/csprd01/virtio-v1.1-csprd01.html#x1-320005 - 2.6.5.3 Indirect Descriptors. [The tests](https://github.com/rust-vmm/vm-virtio/blob/main/crates/virtio-queue/src/lib.rs#L826) should also be updated so they follow the spec, i.e. VIRTQ_DESC_F_INDIRECT and VIRTQ_DESC_F_NEXT must not be both set in flags.

PR #115, opened 2021-11-08
virtio-queue: improve testing

lauralt:
This PR adds or improves tests for validating the notification suppression mechanism, the reset, indirect descriptor requirements and processing of them.

### DIFF
--- a/crates/virtio-queue/src/lib.rs
+++ b/crates/virtio-queue/src/lib.rs
@@ -125,8 +125,6 @@ impl Descriptor {
 
     /// Check whether this descriptor refers to a buffer containing an indirect descriptor table.
     pub fn refers_to_indirect_table(&self) -> bool {
-        // TODO: The are a couple of restrictions in terms of which flags combinations are
-        // actually valid for indirect descriptors. Implement those checks as well somewhere.
         self.flags() & VIRTQ_DESC_F_INDIRECT != 0
     }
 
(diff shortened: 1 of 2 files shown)


jiangliu (dismissed):

crates/virtio-queue/src/lib.rs
```diff
@@ -1182,12 +1182,38 @@ mod tests {
...
+            assert_eq!(c.ttl, 0);
         }
     }
 
+    #[test]
+    fn test_ttl_wrap_around() {
+        let m = &GuestMemoryMmap::<()>::from_ranges(&[(GuestAddress(0), 0x100000)]).unwrap();
+        let vq = MockSplitQueue::new(m, 16);
+
+        for j in 0..15 {
```
andreeaflorescu, line 1200: nit: Why use j?

crates/virtio-queue/src/lib.rs
```diff
@@ -1182,12 +1182,38 @@ mod tests {
...
             assert!(c.next().is_some());
+            // The descriptor above was the last from the chain, so `ttl` should be 0 now.
+            assert_eq!(c.ttl, 0);
             assert!(c.next().is_none());
+            assert_eq!(c.ttl, 0);
         }
     }
 
+    #[test]
+    fn test_ttl_wrap_around() {
```
andreeaflorescu, line 1196: I am not sure why is this called `wrap_around`. Can you add some more details about the test case that you're adding here?

crates/virtio-queue/src/lib.rs
```diff
@@ -1493,11 +1567,35 @@ mod tests {
...
+        q.state.signalled_used = Some(Wrapping(15));
+        assert_eq!(q.state.size, 8);
+        // `create_queue` also marks the queue as ready.
+        assert_eq!(q.state.ready, true);
+        assert_ne!(q.state.desc_table, GuestAddress(DEFAULT_DESC_TABLE_ADDR));
+        assert_ne!(q.state.avail_ring, GuestAddress(DEFAULT_AVAIL_RING_ADDR));
+        assert_ne!(q.state.used_ring, GuestAddress(DEFAULT_USED_RING_ADDR));
+        assert_ne!(q.state.next_avail, Wrapping(0));
+        assert_ne!(q.state.next_used, Wrapping(0));
+        assert_ne!(q.state.signalled_used, None);
```
andreeaflorescu, line 1587: Why do we check these before and after the reset? They seem to have the same values.

crates/virtio-queue/src/lib.rs
```diff
@@ -1544,6 +1642,11 @@ mod tests {
         q.state.next_used = Wrapping(0);
         assert_eq!(q.needs_notification().unwrap(), true);
         assert_eq!(q.needs_notification().unwrap(), false);
+
+        m.write_obj::<u16>(u16::MAX - 3, avail_addr.unchecked_add(4 + qsize as u64 * 2))
```
andreeaflorescu, line 1646: Can you add a comment here to explain why you're calling write again & needs notification is true?

crates/virtio-queue/src/lib.rs
```diff
@@ -1582,4 +1685,154 @@ mod tests {
...
+
+        let mut i = 0;
+
+        loop {
+            i += 1;
+            q.disable_notification().unwrap();
+
+            while let Some(_chain) = q.iter().unwrap().next() {
+                // Do something with the chain. Here the device would consume entries from the
+                // available ring, add an entry in the used ring and optionally notify the driver.
```
andreeaflorescu, line 1728: I would also add a comment that for the purpose of this test what happens with the chain is not relevant, so we just ignore it as we just want to consume the entries in the queue.

andreeaflorescu (commented):
Can you add some more comments in the tests? For some of them it's not obvious why we're adding them, so maybe we can give a short description at the beginning (i.e. this is a negative test, its purpose it to check that in case of  misconfiguration (name the misconfiguration) the device does not cause a panic).

Can you also update the commit message? I think you were saying that some of the tests that we already have were not following the specification. We should name them in the commit message to have a history. We should also say in the commit message what kind of tests we're adding and why. What is particularly interesting about the tests is that we are adding negative tests checking what happens when the driver is not following the spec (i.e. the things with MUST).

Just saying in the commit message that we're "improving testing" is an understatement ;)

crates/virtio-queue/src/lib.rs
lauralt: Added some comments in the test.

crates/virtio-queue/src/lib.rs
lauralt: Just some copy paste from another test, I changed it with `i`.

crates/virtio-queue/src/lib.rs
lauralt: I am using `assert_ne!` before the reset.

crates/virtio-queue/src/lib.rs
lauralt: done

crates/virtio-queue/src/lib.rs
lauralt: done

lauralt:
> Can you add some more comments in the tests? For some of them it's not obvious why we're adding them, so maybe we can give a short description at the beginning (i.e. this is a negative test, its purpose it to check that in case of misconfiguration (name the misconfiguration) the device does not cause a panic).

I added more comments.

> 
> Can you also update the commit message? I think you were saying that some of the tests that we already have were not following the specification. We should name them in the commit message to have a history. We should also say in the commit message what kind of tests we're adding and why. What is particularly interesting about the tests is that we are adding negative tests checking what happens when the driver is not following the spec (i.e. the things with MUST).

I updated the commit message. I'm not fixing yet the tests that were not following the spec. This was part of [this issue](https://github.com/rust-vmm/vm-virtio/issues/95) and I'll handle it in a separate PR.

> 
> Just saying in the commit message that we're "improving testing" is an understatement ;)

lauralt:
> > Can you add some more comments in the tests? For some of them it's not obvious why we're adding them, so maybe we can give a short description at the beginning (i.e. this is a negative test, its purpose it to check that in case of misconfiguration (name the misconfiguration) the device does not cause a panic).
> 
> I added more comments.
> 
> > Can you also update the commit message? I think you were saying that some of the tests that we already have were not following the specification. We should name them in the commit message to have a history. We should also say in the commit message what kind of tests we're adding and why. What is particularly interesting about the tests is that we are adding negative tests checking what happens when the driver is not following the spec (i.e. the things with MUST).
> 
> I updated the commit message. I'm not fixing yet the tests that were not following the spec. This was part of [this issue](https://github.com/rust-vmm/vm-virtio/issues/95) and I'll handle it in a separate PR.
> 
> > Just saying in the commit message that we're "improving testing" is an understatement ;)

I added a separate commit for https://github.com/rust-vmm/vm-virtio/issues/95 and also did some fixes to the already existing test for indirect descriptors. More details in the commit message/comments.

slp (approved):
LGTM, thanks!

jiangliu (approved):

PR #118, opened 2021-11-17
Fix support for big-endian architectures

slp:
This is an attempt at fixing support for big-endian architectures by:

- Stating that `Descriptor` and `VirtqUsedElem` fields are stored with little-endian ordering.
- Transforming from to little-endian in their respective `new` methods, and from little-endian on their accessors.
- Ensuring we're always using the accessors to read the fields.
- Doing endianess conversions when reading/writing individual values from/to memory.

The `::from_le` and `::to_le` methods are no-ops on little-endian machines, so their introduction shouldn't have any impact on those architectures.

I'm marking this PR as a draft, as I don't think it makes sense to merge it until we've reached a final decision about the API change in https://github.com/rust-vmm/vm-virtio/pull/111

### DIFF
--- a/crates/virtio-queue/src/iterator.rs
+++ b/crates/virtio-queue/src/iterator.rs
@@ -131,10 +131,10 @@ mod tests {
             vq.desc_table().store(j, desc);
         }
 
-        vq.avail().ring().ref_at(0).store(0);
-        vq.avail().ring().ref_at(1).store(2);
-        vq.avail().ring().ref_at(2).store(5);
-        vq.avail().idx().store(3);
+        vq.avail().ring().ref_at(0).store(u16::to_le(0));
+        vq.avail().ring().ref_at(1).store(u16::to_le(2));
+        vq.avail().ring().ref_at(2).store(u16::to_le(5));
+        vq.avail().idx().store(u16::to_le(3));
 
         let mut i = q.iter().unwrap();
 
@@ -206,9 +206,9 @@ mod tests {
                 vq.desc_table().store(j, desc);
             }
 
-            vq.avail().ring().ref_at(0).store(0);
-            vq.avail().ring().ref_at(1).store(2);
-            vq.avail().idx().store(2);
+            vq.avail().ring().ref_at(0).store(u16::to_le(0));
+            vq.avail().ring().ref_at(1).store(u16::to_le(2));
+            vq.avail().idx().store(u16::to_le(2));
 
             let mut i = q.iter().unwrap();
 
--- a/crates/virtio-queue/src/iterator.rs
+++ b/crates/virtio-queue/src/iterator.rs
@@ -86,6 +86,7 @@ where
         let head_index: u16 = self
             .mem
             .load(addr, Ordering::Acquire)
+            .map(u16::from_le)
             .map_err(|_| error!("Failed to read from memory {:x}", addr.raw_value()))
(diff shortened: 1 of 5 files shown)


slp:
Since there's no agreement about #111, I'm marking this one as ready to review to move things forward.

jiangliu (approved):

slp:
Turning this PR back into a draft. I want to make use of vm_memory:{Le16, Le32, Le64}, and make sure that all tests that do assertions of `Descriptor` fields apply the proper endianess conversions.

slp:
Rebased and tests working in big-endian, it's now ready to be reviewed.

jiangliu:
Has this patch been tested on any big-endian platform?

slp:
> Has this patch been tested on any big-endian platform?

Yes, on s390x.

jiangliu:
> > Has this patch been tested on any big-endian platform?
> 
> Yes, on s390x.

great, feel much more confident:)

slp:
@jiangliu @alexandruag How would you feel about merging this one? We really need it for virtiofsd.

jiangliu (approved):

andreeaflorescu:
@slp I will take a look at this one today/tomorrow.

crates/virtio-queue/src/iterator.rs
```diff
@@ -131,10 +131,10 @@ mod tests {
             vq.desc_table().store(j, desc);
         }
 
-        vq.avail().ring().ref_at(0).store(0);
-        vq.avail().ring().ref_at(1).store(2);
-        vq.avail().ring().ref_at(2).store(5);
-        vq.avail().idx().store(3);
+        vq.avail().ring().ref_at(0).store(u16::to_le(0));
```
andreeaflorescu, line 134: Should we instead update the `store` and `load` from the mock interface to write and read little endian values so we don't need to use the conversion in all tests?

andreeaflorescu (commented):
Nice written commit messages :+1: 

I have a slight concern that we might miss some of these checks in the tests going forward. It would be nice to setup a big endian platform for testing in the future.

crates/virtio-queue/src/iterator.rs
slp: I thought about that, the problem is that we don't always want to do the endianess translation. For instance, we don't want to do it for `Descriptor`, as it's already in the right endianess. Adding a new dedicated method doesn't solve the problem, as the dev would still need if the conversion needs to be done. I think the only "right" solution would be revamping the mock objects to avoid naked loads/stores from/to memory, having dedicated methods for each field with the right types, but that seems like too huge of a change to me.

Being pragmatic, as this only affects tests and it's probably going to be very hard to integrate a big-endian machine into the CI, I think it's fine if they get broken from time to time. I'll give them a try on each new virtiofsd-rs release and fix them if needed.

andreeaflorescu: Sounds good. We want to do some more improvements to the mock interface, do you mind opening an issue with this suggestion that you just mentioned?

andreeaflorescu (approved):

crates/virtio-queue/src/iterator.rs
slp: Sure, I've created #123 to track this. Thanks!

PR #119, opened 2021-11-18
virtio-queue: Publicize DescriptorChain::new()

vireshk:
Make DescriptorChain::new() public, so it can be used by other crates
(specially for testing).

### DIFF
--- a/crates/virtio-queue/src/lib.rs
+++ b/crates/virtio-queue/src/lib.rs
@@ -179,7 +179,7 @@ impl<M: GuestAddressSpace> DescriptorChain<M> {
     }
 
     /// Create a new `DescriptorChain` instance.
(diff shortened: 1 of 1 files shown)


vireshk:
Required for https://github.com/rust-vmm/vhost-device/pull/63.

lauralt:
I didn't do an in-depth review of https://github.com/rust-vmm/vhost-device/pull/63, just had a quick look, but isn't it possible to create the descriptor chain similar to how that is done in [build_desc_chain](https://github.com/krady21/vm-virtio/blob/resolver2/crates/devices/virtio-blk/src/request.rs#L295)?

vireshk:
Thanks, I was able to make it work that way.

PR #120, opened 2021-11-21
Split virtio-queue into multiple submodules for maintenance

### LINKED ISSUE #87: [virtio-queue] Split the code in multiple modules
Right now everything is in 1 file, and it's pretty clutter.

jiangliu:
Fix #87 by split the lib.rs of virtio-queue into multiple submodules:
1) split it into several submodules for maintenance
2) use setter/getter to avoid direct access to data structure fields
3) refine unit test cases to improve code coverage, now it reaches 90% code coverage.


bonzini:
The QueueStateSync API is absolutely not ready for being published.

jiangliu:
> The QueueStateSync API is absolutely not ready for being published.

Well formed API designs are always welcomed:)
But we hope we could publish the virtio-queue crate sooner than later, because several other projects depend on it. The situation has last more than one year. We have tried several times during the past year to migrate to the upstream vm-virtio implementation but failed. Then the cloud hypervisor, virtiofsd and our projects turned to maintain their own private versions, it's really a burden for us:(

jiangliu:
> The QueueStateSync API is absolutely not ready for being published.

BTW, I have one more patch to convert QueueStateGuard from an enum into a trait, just as one of your previous patch did. It does help to make the API clear and extensible:)

bonzini:
The problem is that Queue is not solving the problem in a way that works for the vhost-backend crate for example.

I think we don't even need Queue or QueueSync. What we need is a good guard object so that one doesn't have to use the flexible API with GuestMemory arguments, when the easier one can be used.

In addition, signal_used_queue must definitely be a trait and the guard should know about it. I'll try to come up with something later this week.

jiangliu:
> The problem is that Queue is not solving the problem in a way that works for the vhost-backend crate for example.
> 
Actually we need to support several use cases:
1) support normal virtio devices, such as virtio-blk/virtio-net
2) support vhost daemons with vhost-user-backend
3) support rust async io to improve performance

> I think we don't even need Queue or QueueSync. What we need is a good guard object so that one doesn't have to use the flexible API with GuestMemory arguments, when the easier one can be used.

Is this something you are thinking of? https://github.com/rust-vmm/vm-virtio/pull/121 It introduces two changes:
- convert enum QueueStateGuard to trait QueueStateGuard for extensibility and zero-cost abstraction.
- introduce "struct QueueGuard` to encapsulate a QueueStateGuard object and associated guest memory object.

> 
> In addition, signal_used_queue must definitely be a trait and the guard should know about it. I'll try to come up with something later this week.

crates/virtio-queue/src/state_sync.rs
```diff
@@ -78,26 +85,207 @@ impl QueueStateT for QueueStateSync {
...
     }
 
     fn set_next_avail(&mut self, next_avail: u16) {
-        self.state.lock().unwrap().set_next_avail(next_avail);
+        self.lock_state().set_next_avail(next_avail);
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
alexandruag, line 113: Hmm, looks like most of the tests here would apply to other `QueueStateT` implementations just as well. It would actually be great to have a battery of tests for the `QueueStateT` interface that can then be reused to test all implementations (so far there's `QueueState` and `QueueStateSync`), and then separate tests for stuff that's specific to `QueueStateSync` in particular (altho there's not that much truly specific stuff IMO; even the test where multiple threads are spawned can be seen as generic due to `lock` being part of the high level interface).

Not sure exactly just how much effort that would require, so maybe doing it as part of this PR is not the best approach, but then I think we should at least have an issue and tackle it b4 publishing.

alexandruag (commented):
This def looks good as breaking up the code into separate modules is something that's been long overdue. Thanks Gerry! Left just one comment regarding some of the tests.

crates/virtio-queue/src/state_sync.rs
jiangliu: Will open one issue for it once the PR get merged:)

alexandruag (approved):
Cool, I think the refactoring here is def helpful while we work to wrap up the other ongoing discussions.

slp (approved):
LGTM. Thanks @jiangliu

PR #121, opened 2021-11-22
Refine QueueStateGuard and introduce QueueGuard

jiangliu:
The last two patches tries to:
1) convert `enum QueueStateGuard` to `trait QueueStateGuard` for extensibility and zero-cost abstraction.
2) introduce "struct QueueGuard` to encapsulate a QueueStateGuard object and associated guest memory object.

### DIFF
--- a/crates/virtio-queue/src/queue.rs
+++ b/crates/virtio-queue/src/queue.rs
@@ -50,7 +50,7 @@ impl<M: GuestAddressSpace, S: QueueStateT> Queue<M, S> {
     ///
     /// Logically this method will acquire the underlying lock protecting the `QueueState` Object.
     /// The lock will be released when the returned object gets dropped.
-    pub fn lock(&mut self) -> QueueStateGuard {
+    pub fn lock(&mut self) -> <S as QueueStateGuard>::G {
         self.state.lock()
     }
 
--- a/crates/virtio-queue/src/queue.rs
+++ b/crates/virtio-queue/src/queue.rs
@@ -16,7 +16,7 @@ use std::sync::atomic::Ordering;
 
 use vm_memory::GuestAddressSpace;
 
-use crate::{AvailIter, Error, QueueState, QueueStateGuard, QueueStateT};
+use crate::{AvailIter, Error, QueueGuard, QueueState, QueueStateGuard, QueueStateT};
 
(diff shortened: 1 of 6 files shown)


crates/virtio-queue/src/queue.rs
```diff
@@ -555,4 +700,70 @@ mod tests {
             }
         }
     }
+
+    #[test]
+    fn test_queeu_guard_object() {
```
slp, line 705: Typo here (queeu).

slp (changes requested):
Thanks @jiangliu. The trait does indeed look nicer than the enum (perhaps we should do the same in in vhost-user-backend's vring). Could you please rebase this on top of #120 ?

jiangliu:
> Thanks @jiangliu. The trait does indeed look nicer than the enum (perhaps we should do the same in in vhost-user-backend's vring). Could you please rebase this on top of #120 ?

yes, will handle vhost-user-backend once the virtio-queue settle down:)

crates/virtio-queue/src/queue.rs
jiangliu: Fixed

slp (dismissed):
LGTM and doesn't break the current API. Nice one @jiangliu

crates/virtio-queue/src/queue.rs
```diff
@@ -11,13 +11,147 @@
...
+use vm_memory::{GuestAddressSpace, GuestMemory};
 
 use crate::{AvailIter, Error, QueueState, QueueStateGuard, QueueStateT};
 
+/// A guard object to exclusively access an `Queue` object.
+///
+/// The guard object holds an exclusive lock to the underlying `QueueState` object, with an
+/// associated guest memory object. It helps to guarantee that the whole session is served
+/// with the same guest memory object.
+pub struct QueueGuard<M, S> {
```
alexandruag, line 26: Hmm, could we move `QueueGuard` to its own module as well?

alexandruag (commented):
Just had a few small comments, but otherwise looks good to merge this and then I can rebase some potential changes on top, to have a discussion about how to proceed before publishing 0.1.0 based on pending proposals, including Paolo's PR.

slp (approved):

alexandruag (approved):

Issue #99, opened 2021-10-12
[virtio-queue] document `unsafe impl` safety


lauralt:
We have two `unsafe impl`s: https://github.com/rust-vmm/vm-virtio/blob/main/crates/virtio-queue/src/lib.rs#L144, https://github.com/rust-vmm/vm-virtio/blob/main/crates/virtio-queue/src/lib.rs#L385 that should be documented on why they are safe.

bonzini:
The description for ByteValued says

> A type `T` is `ByteValued` if and only if it can be initialized by reading its contents from a byte array.  This is generally true for all plain-old-data structs.  It is notably not true for any type that includes a reference.
>
> Implementing this trait guarantees that it is safe to instantiate the struct with random data.

So the description would be that 1) `Descriptor` or `VirtqElem` structs *are* plain-old-data 2) all accesses through safe vm-memory API will validate any garbage that could be included in there. Is that correct or am I missing something?

PR #124, opened 2021-12-08
document/fix unchecked operations

### LINKED ISSUE #97: [virtio-queue] document unchecked_* operations safety
We have a couple of `unchecked_*` operations that would benefit from a comment on why they are safe to use, e.g. https://github.com/rust-vmm/vm-virtio/blob/main/crates/virtio-queue/src/lib.rs#L581.

lauralt:
Fixes #97 #99

### DIFF
--- a/crates/virtio-queue/src/state.rs
+++ b/crates/virtio-queue/src/state.rs
@@ -74,9 +74,14 @@ impl QueueState {
         val: u16,
         order: Ordering,
     ) -> Result<(), Error> {
-        let elem_sz = VIRTQ_USED_ELEMENT_SIZE * u64::from(self.size);
-        let offset = VIRTQ_USED_RING_HEADER_SIZE + elem_sz;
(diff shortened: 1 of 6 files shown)


crates/virtio-queue/src/state.rs
```diff
@@ -74,9 +74,16 @@ impl QueueState {
         val: u16,
         order: Ordering,
     ) -> Result<(), Error> {
+        // The multiplication can not overflow an u64 since we are multiplying an u16 with a
```
andreeaflorescu, line 77: Can we combine these two comments into one saying that the following operations cannot overflow because they're working with small numbers? I find the comments to be too much and getting in the way of the readability. Same for the other ones where there are 2 subsequent operations.

lauralt: Done, I also combined some of the operations.

andreeaflorescu (approved):

jiangliu (approved):

Issue #128, opened 2022-01-10
add abstraction for the vsock packet


lauralt:
We were thinking to start upstreaming the virtio vsock device, starting with the vsock packet. We already have some abstractions for the block device request, and we can go the same way for the vsock packet, i.e. have a new crate in vm-virtio: `virtio-vsock`.

For a bit of context, the virtio vsock packet is defined in the standard as having a header of type `virtio_vsock_hdr` and an optional `data` array of bytes. There are multiple operations that can be requested within a packet, e.g. VIRTIO_VSOCK_OP_RST for resetting the connection,  VIRTIO_VSOCK_OP_RW for sending payload. Most operations are of the VIRTIO_VSOCK_OP_RW type, which means for data transfer, and the other ones are used for connection and buffer space management. data is non-empty only for the VIRTIO_VSOCK_OP_RW operations.

```c
struct virtio_vsock_packet {
    struct virtio_vsock_hdr hdr;
    u8 data[];
};
```
Our proposal is to use vm-memory’s `VolatileSlice`, and define the` VsockPacket` structure as follows:

```rust
pub struct VsockPacket<'a, B: BitmapSlice> {
    header_slice: VolatileSlice<'a, B>,
    header: PacketHeader,
    data_slice: Option<VolatileSlice<'a, B>>,
}
```

The advantage with this approach is that we don’t tie the packet implementation to the `GuestMemory` interface (which would happen if we would use `GuestAddress`es instead of `VolatileSlice`s). A packet can be configured from simply pointers to a slice, not necessarily from a `GuestMemory` object (e.g. the vm-virtio’s `DescriptorChain`), so with this approach we would support this use case too. A limitation of this approach is that it doesn’t cover the scenario where the header or data buffer doesn't fit in a single `VolatileSlice` because the guest memory regions of the buffer are contiguous in the guest physical address space, but not in the host virtual one as well. This is not currently happening for the known customers (Firecracker, Cloud Hypervisor); the vsock packet corresponds to only one region of the host virtual space. For the multiple `VolatileSlice`s use case, we can later extend this solution to use an array of `VolatileSlice`s for the header and data.

**Design**

[...]

Issue #114, opened 2021-11-04
[virtio-queue/device] limit the maximum number of queues


lauralt:
We should consider limiting the maximum number of queues. Right now the MMIO transport implementation allows [up to u16::MAX](https://github.com/rust-vmm/vm-virtio/blob/main/crates/virtio-device/src/mmio.rs#L151) queues which is a bit too much. We could instead define a maximum number and when the driver wants to select the queue index, validate this index against the max one. qemu seems to be using [1024](https://github.com/qemu/qemu/blob/master/include/hw/virtio/virtio.h#L66), which seems reasonable.
This is not a blocker for publishing the virtio-queue as it is more related to the `virtio-device` scope.

RaduNiculae:
The current implementation of the MMIO transport provides a way for the device to limit the number of queues the driver can access. If the driver tries to select a queue that has not been initialized by the device it will receive [the value 0](https://github.com/rust-vmm/vm-virtio/blob/main/crates/virtio-device/src/mmio.rs#L85) in the [QueueNumMax](https://docs.oasis-open.org/virtio/virtio/v1.1/csprd01/virtio-v1.1-csprd01.html#x1-1460002) register and should consider the queue to be invalid. This way the driver is limited by the number of queues created by the device. A limitation could be added to the number of queues that can be created, but that should be device specific since adding a maximum of 1024 in the virtio-device would not allow for devices such as the network device to create up to 0x8000 queues as it is defined in the [specification](https://docs.oasis-open.org/virtio/virtio/v1.1/csprd01/virtio-v1.1-csprd01.html#x1-2010001). The issue could be better handled in each device implementation instead of the generic virtio-device. Synced with @lauralt as well, and decided to close the issue.

PR #131, opened 2022-01-20
add VsockPacket implementation

lauralt:
This PR is adding an implementation of the vsock packet, based on the design proposal from https://github.com/rust-vmm/vm-virtio/issues/128. The documentation is in a decent shape already, but I will open a separate PR with some improvements to it and also code examples.
Open questions:
1. Do we want to have an upper limit for the data array size? I didn't add one yet, cloud-hypervisor and firecracker are using [the one defined in the linux kernel](https://elixir.bootlin.com/linux/v4.9.12/source/include/linux/virtio_vsock.h#L14). Do we assume this one to be a good enough limit for a vsock driver?
2. On the TX path, should we go with the `len` from the header for the data array size or do we keep the one from the data descriptor? I'll think a bit more about this.

### DIFF
--- a/crates/devices/virtio-vsock/Cargo.toml
+++ b/crates/devices/virtio-vsock/Cargo.toml
@@ -1,0 +1,18 @@
+[package]
+name = "virtio-vsock"
+version = "0.1.0"
+authors = ["rust-vmm community", "rust-vmm AWS maintainers <rust-vmm-maintainers@amazon.com>"]
+description = "virtio vsock device implementation"
+repository = "https://github.com/rust-vmm/vm-virtio"
+keywords = ["virtio", "vsock"]
+readme = "README.md"
+license = "Apache-2.0 OR BSD-3-Clause"
+edition = "2018"
+
+[dependencies]
+virtio-queue = "0.2.0"
+vm-memory = "0.7.0"
+
+[dev-dependencies]
+virtio-queue = { version = "0.2.0", features = ["test-utils"] }
+vm-memory = { version = "0.7.0", features = ["backend-mmap", "backend-atomic"] }
--- a/crates/devices/virtio-vsock/src/packet.rs
+++ b/crates/devices/virtio-vsock/src/packet.rs
@@ -138,6 +138,7 @@ pub type Result<T> = std::result::Result<T, Error>;
 /// The vsock packet, implemented as a wrapper over a virtio descriptor chain:
 /// - the chain head, holding the packet header;
 /// - an optional data/buffer descriptor, only present for data packets (for VSOCK_OP_RW requests).
+#[derive(Debug)]
 pub struct VsockPacket<'a, B: BitmapSlice> {
     // When writing to the header slice, we are using the `write` method of `VolatileSlice`s Bytes
     // implementation. Because that can only return an error if we pass an invalid offset, we can
@@ -189,6 +190,7 @@ impl<'a, B: BitmapSlice> VsockPacket<'a, B> {
         self.header = header;
         Ok(())
     }
+
     /// Return the `src_cid` of the header.
     pub fn src_cid(&self) -> u64 {
         self.header.src_cid.into()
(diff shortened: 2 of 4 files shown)


lauralt:
@jiangliu @sboeuf Can you also have a look at this PR? Thank you!

andreeaflorescu:
> 1. Do we want to have an upper limit for the data array size? I didn't add one yet, cloud-hypervisor and firecracker are using [the one defined in the linux kernel](https://elixir.bootlin.com/linux/v4.9.12/source/include/linux/virtio_vsock.h#L14). Do we assume this one to be a good enough limit for a vsock driver?

Who is filling the data array? It is a good practice to set a size, using the same defined as in the Linux Kernel sounds like a good plan.

> 2. On the TX path, should we go with the `len` from the header for the data array size or do we keep the one from the data descriptor? I'll think a bit more about this.

I'll take a look at the code as well and come back to this.

lauralt:
> > 1. Do we want to have an upper limit for the data array size? I didn't add one yet, cloud-hypervisor and firecracker are using [the one defined in the linux kernel](https://elixir.bootlin.com/linux/v4.9.12/source/include/linux/virtio_vsock.h#L14). Do we assume this one to be a good enough limit for a vsock driver?
> 
> Who is filling the data array? It is a good practice to set a size, using the same defined as in the Linux Kernel sounds like a good plan.

It is filled by the driver on the path where it would make sense to check for this limit. @harshanavkis @stsquad Is this limit ok for you too? I've noticed you don't check for one in the vhost-user-vsock packet implementation.

stsquad:
> > > 1. Do we want to have an upper limit for the data array size? I didn't add one yet, cloud-hypervisor and firecracker are using [the one defined in the linux kernel](https://elixir.bootlin.com/linux/v4.9.12/source/include/linux/virtio_vsock.h#L14). Do we assume this one to be a good enough limit for a vsock driver?
> > Who is filling the data array? It is a good practice to set a size, using the same defined as in the Linux Kernel sounds like a good plan.
> It is filled by the driver on the path where it would make sense to check for this limit. @harshanavkis @stsquad Is this limit ok for you too? I've noticed you don't check for one in the vhost-user-vsock packet implementation.

AFAICT it's a fairly arbitrary limit from the Linux kernel which isn't part of the spec and my general feeling is not to embed specific implementation limitations into the back-end. However given there aren't any alternate vsock front-ends it's not the end of the world if it has a benefit to the code - does having a limit make things easier for the rust compiler?

andreeaflorescu (commented):
Part 1

harshanavkis:
> > > > 1. Do we want to have an upper limit for the data array size? I didn't add one yet, cloud-hypervisor and firecracker are using [the one defined in the linux kernel](https://elixir.bootlin.com/linux/v4.9.12/source/include/linux/virtio_vsock.h#L14). Do we assume this one to be a good enough limit for a vsock driver?
> > > >    Who is filling the data array? It is a good practice to set a size, using the same defined as in the Linux Kernel sounds like a good plan.
> > > >    It is filled by the driver on the path where it would make sense to check for this limit. @harshanavkis @stsquad Is this limit ok for you too? I've noticed you don't check for one in the vhost-user-vsock packet implementation.
> 
> AFAICT it's a fairly arbitrary limit from the Linux kernel which isn't part of the spec and my general feeling is not to embed specific implementation limitations into the back-end. However given there aren't any alternate vsock front-ends it's not the end of the world if it has a benefit to the code - does having a limit make things easier for the rust compiler?

Yes I didn't specify a data size in the packet implementation for the device because I think 64KiB is the maximum transfer size possible and the packet at runtime may be of a smaller data size because of flow control for SOCK_STREAM type vsock sockets. As far as I know [SEQ_PACKETS](https://lwn.net/Articles/856826/) doesn't enforce a max size of the data packet. And we will need to add it to the device in the future as well.

crates/devices/virtio-vsock/src/packet.rs
```diff
@@ -0,0 +1,386 @@
...
+    /// Read only descriptor that protocol says to write to.
+    UnexpectedReadOnlyDescriptor,
+    /// Write only descriptor that protocol says to read from.
+    UnexpectedWriteOnlyDescriptor,
+}
+
+// The vsock packet header is defined by the following structure:
+//
+// ```
+// struct virtio_vsock_hdr {
```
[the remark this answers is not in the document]
lauralt: I'll give it a shot.

crates/devices/virtio-vsock/src/packet.rs
```diff
@@ -0,0 +1,934 @@
...
+            .map_err(Error::InvalidMemoryAccess)?;
+
+        let mut pkt = Self {
+            header,
+            header_inner,
+            data: None,
+        };
+
+        // If the `len` field of the header is zero, then the packet doesn't have a `data` element.
+        if pkt.len() == 0 {
```
gsserge, line 338: I don't have a very strong opinion about this, but maybe it's worth it to implement `.is_empty()` and get rid of `#[allow(clippy::len_without_is_empty)]` if we are doing the empty check here.

crates/devices/virtio-vsock/src/packet.rs
```diff
@@ -0,0 +1,934 @@
...
+            return Err(Error::UnexpectedReadOnlyDescriptor);
+        }
+
+        let data = mem
+            .get_slice(data_desc.addr(), data_desc.len() as usize)
+            .map_err(Error::InvalidMemoryAccess)?;
+
+        Ok(Self {
+            header,
+            header_inner: Default::default(),
```
gsserge, line 408: Why are we using the default values for `header_inner` instead of initializing it based on `header` similar to `from_tx_virtq_chain `?

lauralt: On the RX path there is no data in the packet's header and data buffers. The driver is providing empty buffers to be later filled by the device.

crates/devices/virtio-vsock/src/packet.rs
lauralt: Good idea, I'll add it.

lauralt:
> > > > 1. Do we want to have an upper limit for the data array size? I didn't add one yet, cloud-hypervisor and firecracker are using [the one defined in the linux kernel](https://elixir.bootlin.com/linux/v4.9.12/source/include/linux/virtio_vsock.h#L14). Do we assume this one to be a good enough limit for a vsock driver?
> > > >    Who is filling the data array? It is a good practice to set a size, using the same defined as in the Linux Kernel sounds like a good plan.
> > > >    It is filled by the driver on the path where it would make sense to check for this limit. @harshanavkis @stsquad Is this limit ok for you too? I've noticed you don't check for one in the vhost-user-vsock packet implementation.
> 
> AFAICT it's a fairly arbitrary limit from the Linux kernel which isn't part of the spec and my general feeling is not to embed specific implementation limitations into the back-end. However given there aren't any alternate vsock front-ends it's not the end of the world if it has a benefit to the code - does having a limit make things easier for the rust compiler?

The benefit wouldn't be to the code, but from what I know it's a good practice to have some sort of limit for arrays, such that the device implementation doesn't allow a driver to use super huge chunks of memory, and potentially fill up the entire memory. The same rationale was behind [this issue](https://github.com/rust-vmm/vm-virtio/issues/114) as well.

lauralt:
@andreeaflorescu @gsserge I addressed your comments.

Regarding the max vsock packet data size discussion, I opened [this thread](https://lists.oasis-open.org/archives/virtio-comment/202202/msg00053.html) in virtio-spec; the proposal from there would be to have a new field in the vsock configuration space for that. Since this will involve changes that are going to take some time, I propose we merge this and open separate PRs for supporting a configurable max data size, as already discussed in private with @stsquad.

CC @jiangliu @sboeuf

crates/devices/virtio-vsock/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+license = "Apache-2.0 OR BSD-3-Clause"
+edition = "2018"
+
+[dependencies]
+vm-memory = "0.7.0"
+virtio-queue = { path = "../../virtio-queue" }
+
+[dev-dependencies]
+vm-memory = { version = "0.7.0", features = ["backend-mmap", "backend-atomic"] }
+virtio-queue = { path = "../../virtio-queue", features = ["test-utils"] }
```
sboeuf, line 18: Ditto.

sboeuf (commented):
Looks good, but could you please provide more elaborate commit messages. I know you described the intent of this patchset through the PR description, but unfortunately it won't be part of the git history :)

lauralt:
> Looks good, but could you please provide more elaborate commit messages. I know you described the intent of this patchset through the PR description, but unfortunately it won't be part of the git history :)

Sure

crates/devices/virtio-vsock/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+description = "virtio vsock device implementation"
+repository = "https://github.com/rust-vmm/vm-virtio"
+keywords = ["virtio", "vsock"]
+readme = "README.md"
+license = "Apache-2.0 OR BSD-3-Clause"
+edition = "2018"
+
+[dependencies]
+vm-memory = "0.7.0"
+virtio-queue = { path = "../../virtio-queue" }
```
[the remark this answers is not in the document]
lauralt: Makes sense. I updated it.

crates/devices/virtio-vsock/Cargo.toml
lauralt: done

lauralt:
@sboeuf I added more details to the commit messages. Please have another look.

sboeuf (approved):

andreeaflorescu (approved):

crates/devices/virtio-vsock/src/packet.rs
```diff
@@ -0,0 +1,386 @@
...
+}
+
+#[allow(clippy::len_without_is_empty)]
+impl<'a, B: BitmapSlice> VsockPacket<'a, B> {
+    /// Return a reference to the `header` of the packet.
+    pub fn header(&self) -> &VolatileSlice<'a, B> {
+        &self.header
+    }
+
+    /// Return a reference to the `data` of the packet.
```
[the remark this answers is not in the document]
lauralt: I'll address it in the next PR. I have to do some improvements to docs anyway. Thanks.

PR #132, opened 2022-01-24
virtio-queue: Add helpers for accessing queue information

sboeuf:
These helpers are meant to help crate's consumers getting and setting
information about the queue.

### DIFF
--- a/crates/virtio-queue/src/lib.rs
+++ b/crates/virtio-queue/src/lib.rs
@@ -154,6 +154,9 @@ pub trait QueueStateT: for<'a> QueueStateGuard<'a> {
     /// Read the `idx` field from the available ring.
     fn avail_idx<M: GuestMemory>(&self, mem: &M, order: Ordering) -> Result<Wrapping<u16>, Error>;
 
+    /// Read the `idx` field from the used ring.
+    fn used_idx<M: GuestMemory>(&self, mem: &M, order: Ordering) -> Result<Wrapping<u16>, Error>;
+
     /// Put a used descriptor head into the used ring.
     fn add_used<M: GuestMemory>(&mut self, mem: &M, head_index: u16, len: u32)
         -> Result<(), Error>;
@@ -181,4 +184,10 @@ pub trait QueueStateT: for<'a> QueueStateGuard<'a> {
 
     /// Set the index of the next entry in the available ring.
     fn set_next_avail(&mut self, next_avail: u16);
+
+    /// Return the index for the next descriptor in the used ring.
+    fn next_used(&self) -> u16;
+
+    /// Set the index for the next descriptor in the used ring.
+    fn set_next_used(&mut self, next_used: u16);
 }
--- a/crates/virtio-queue/src/queue.rs
+++ b/crates/virtio-queue/src/queue.rs
@@ -197,6 +197,14 @@ impl<M: GuestAddressSpace, S: QueueStateT> Queue<M, S> {
         self.state.avail_idx(self.mem.memory().deref(), order)
     }
 
+    /// Reads the `idx` field from the used ring.
+    ///
+    /// # Arguments
+    /// * `order` - the memory ordering used to access the `idx` field from memory.
+    pub fn used_idx(&self, order: Ordering) -> Result<Wrapping<u16>, Error> {
+        self.state.used_idx(self.mem.memory().deref(), order)
+    }
+
     /// Put a used descriptor head into the used ring.
     ///
     /// # Arguments
@@ -236,12 +244,25 @@ impl<M: GuestAddressSpace, S: QueueStateT> Queue<M, S> {
         self.state.next_avail()
     }
 
+    /// Returns the index for the next descriptor in the used ring.
+    pub fn next_used(&self) -> u16 {
+        self.state.next_used()
+    }
+
     /// Set the index of the next entry in the available ring.
     ///
     /// # Arguments
     /// * `next_avail` - the index of the next available ring entry.
     pub fn set_next_avail(&mut self, next_avail: u16) {
         self.state.set_next_avail(next_avail);
+    }
+
+    /// Sets the index for the next descriptor in the used ring.
+    ///
+    /// # Arguments
+    /// * `next_used` - the index of the next used ring entry.
+    pub fn set_next_used(&mut self, next_used: u16) {
+        self.state.set_next_used(next_used);
     }
 }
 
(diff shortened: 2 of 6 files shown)


crates/virtio-queue/src/lib.rs
```diff
@@ -154,6 +154,9 @@ pub trait QueueStateT: for<'a> QueueStateGuard<'a> {
     /// Read the `idx` field from the available ring.
     fn avail_idx<M: GuestMemory>(&self, mem: &M, order: Ordering) -> Result<Wrapping<u16>, Error>;
 
+    /// Read the `idx` field from the used ring.
+    fn used_idx<M: GuestMemory>(&self, mem: &M, order: Ordering) -> Result<Wrapping<u16>, Error>;
```
lauralt, line 158: Can you also add tests for the new setters/getters?

crates/virtio-queue/src/lib.rs
```diff
@@ -179,6 +182,12 @@ pub trait QueueStateT: for<'a> QueueStateGuard<'a> {
...
     fn next_avail(&self) -> u16;
 
+    /// Return the index for the next descriptor in the used ring.
+    fn next_used(&self) -> u16;
+
     /// Set the index of the next entry in the available ring.
     fn set_next_avail(&mut self, next_avail: u16);
+
+    /// Set the index for the next descriptor in the used ring.
+    fn set_next_used(&mut self, next_used: u16);
```
lauralt, line 192: Just curious, I had a look at how you are using these in CH, and I was wondering if this one for example would still be needed if you'd switch to the upstream `QueueState`. I'm not very familiar with PCI.

lauralt: We should also add support for the `Versionize` queue state in the near future, based on the agreed proposal from [here](https://github.com/rust-vmm/community/pull/118).

crates/virtio-queue/src/state.rs
```diff
@@ -311,6 +311,14 @@ impl QueueStateT for QueueState {
             .map_err(Error::GuestMemory)
     }
 
+    fn used_idx<M: GuestMemory>(&self, mem: &M, order: Ordering) -> Result<Wrapping<u16>, Error> {
+        let addr = self.used_ring.unchecked_add(2);
```
lauralt, line 315: I think I would use `checked_add` here for now because the queue's and the state's fields are public, so the user is allowed to configure a queue with invalid values.
I plan to update a bit the save/restore support for the queue, and we can reconsider the `(un)checked_*` story then.

crates/virtio-queue/src/lib.rs
sboeuf: I will.

crates/virtio-queue/src/state.rs
sboeuf: Cool

crates/virtio-queue/src/lib.rs
sboeuf: Yes I think once QueueState will be derived with Versionize, Cloud Hypervisor will be able to rely fully on it rather than redefining locally. But for now there's no other way.

crates/virtio-queue/src/lib.rs
sboeuf: Done!

sboeuf:
PR's ready for review.

crates/virtio-queue/src/lib.rs
```diff
@@ -179,6 +182,12 @@ pub trait QueueStateT: for<'a> QueueStateGuard<'a> {
     /// Return the index of the next entry in the available ring.
     fn next_avail(&self) -> u16;
 
+    /// Return the index for the next descriptor in the used ring.
```
jiangliu, line 185: Could we group getter and setter together by moving line 185-187 to line 190?

jiangliu:
Thanks for improving the unit test logic:)

crates/virtio-queue/src/lib.rs
sboeuf: Sure!

sboeuf: Done.

crates/virtio-queue/src/queue.rs
```diff
@@ -533,10 +556,14 @@ mod tests {
...
 
-            while let Some(_chain) = q.iter().unwrap().next() {
+            while let Some(chain) = q.iter().unwrap().next() {
                 // Here the device would consume entries from the available ring, add an entry in
                 // the used ring and optionally notify the driver. For the purpose of this test, we
                 // don't need to do anything with the chain, only consume it.
+                let head_index = chain.head_index();
+                let mut desc_len = 0;
+                chain.for_each(|d| desc_len += d.len());
+                q.add_used(head_index, desc_len).unwrap();
```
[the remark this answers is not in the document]
lauralt: nit: It would also be very nice to keep some descriptors as device-readable, and for those to not update desc_len.

crates/virtio-queue/src/queue.rs
```diff
@@ -533,10 +556,14 @@ mod tests {
             i += 1;
             q.disable_notification().unwrap();
 
-            while let Some(_chain) = q.iter().unwrap().next() {
+            while let Some(chain) = q.iter().unwrap().next() {
                 // Here the device would consume entries from the available ring, add an entry in
```
lauralt, line 560: Can you also update the comments in this test as they are not accurate anymore? Thanks!

jiangliu (dismissed):

sboeuf:
@lauralt PR updated!

crates/virtio-queue/src/queue.rs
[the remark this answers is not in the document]
sboeuf: I think I addressed your comments, let me know if I missed something.

crates/virtio-queue/src/queue.rs
```diff
@@ -555,8 +582,19 @@ mod tests {
...
+            while let Some(chain) = q.iter().unwrap().next() {
+                // Process the descriptor chain, and then add entries to the
+                // used ring.
+                let head_index = chain.head_index();
+                let mut desc_len = 0;
+                chain.for_each(|d| {
+                    if d.flags() & VIRTQ_DESC_F_WRITE == VIRTQ_DESC_F_WRITE {
+                        desc_len += d.len();
+                    }
+                });
```
lauralt, line 594: We can probably update every for_each with the way it is here so we don't have to care if the descriptors' flags have a particular value.

lauralt (approved):
Thanks for the improvements!

crates/virtio-queue/src/queue.rs
sboeuf: If I do so, I'll have to set in each test a few descriptors with the flag `VIRTQ_DESC_F_WRITE` otherwise the `add_used()` won't be called, and this will change the expected used index. WDYT?

lauralt: But if you parse a chain and there are no writable descriptors, I think you still have to add an used entry, even if with len = 0.

sboeuf: Actually I think you're correct, let me update the tests :)

sboeuf: Done!

lauralt (approved):

rbradford (approved):

Issue #122, opened 2021-11-26
Unify unit test cases of QueueState and QueueStateSync


jiangliu:
One review comments for #120 
Hmm, looks like most of the tests here would apply to other QueueStateT implementations just as well. It would actually be great to have a battery of tests for the QueueStateT interface that can then be reused to test all implementations (so far there's QueueState and QueueStateSync), and then separate tests for stuff that's specific to QueueStateSync in particular (altho there's not that much truly specific stuff IMO; even the test where multiple threads are spawned can be seen as generic due to lock being part of the high level interface).

Not sure exactly just how much effort that would require, so maybe doing it as part of this PR is not the best approach, but then I think we should at least have an issue and tackle it b4 publishing.

andreeaflorescu:
For now we should start with de-duplicated the following tests that are exactly the same (except for the initialization):
- https://github.com/rust-vmm/vm-virtio/blob/15effcedd42453eabc4be7c645d05e3a1fc0303e/crates/virtio-queue/src/queue_guard.rs#L211
- https://github.com/rust-vmm/vm-virtio/blob/15effcedd42453eabc4be7c645d05e3a1fc0303e/crates/virtio-queue/src/queue.rs#L643

We can work next on having all the tests ported to the queue guard as well. This should be though a separate issue.

PR #140, opened 2022-02-28
Remove code duplication in tests

DeliaPavel:
Create function test_state in state.rs that holds the duplicated code
from function test_invalid_avail_idx from queue.rs and function
test_queue_guard_object from queue_guard.rs

Fixes https://github.com/rust-vmm/vm-virtio/issues/122

Signed-off-by: Delia-Maria PAVEL (101431) <delia_maria.pavel@stud.acs.upb.ro>

### Summary of the PR

### Requirements

requirements:

  description line.
  test.


crates/virtio-queue/Cargo.toml
```diff
@@ -13,7 +13,7 @@ edition = "2018"
 test-utils = []
 
 [dependencies]
-vm-memory = "0.7.0"
+vm-memory = { version = "0.7.0", features = ["backend-mmap"] }
```
andreeaflorescu, line 16: Why is this dependency needed? We should not depend on this feature in the implementation, we're just using it in the tests so that custom VMMs can have their on vm-memory implementation

crates/virtio-queue/src/descriptor.rs
```diff
@@ -113,7 +113,6 @@ impl Descriptor {
     }
 }
 
-#[cfg(any(test, feature = "test-utils"))]
```
andreeaflorescu, line 116: This should stay as before, and instead mark the test where it is needed with the `test-utilities` feature.

```rust
#[test]
#[cfg(feature = "test-utils")]
fn test_invalid_avail_idx() {
     ....
}
```

crates/virtio-queue/src/state.rs
```diff
@@ -149,6 +150,77 @@ impl QueueState {
             .map(Wrapping)
             .map_err(Error::GuestMemory)
     }
+
+    /// Used in the test_invalid_avail_idx from queue.rs and test_queue_guard_object from
+    /// queue_guard.rs in order to de-duplicate the tests
+    pub fn test_state<M: GuestMemory>(&mut self, mem: &M, vq: &MockSplitQueue<GuestMemoryMmap>) {
```
andreeaflorescu, line 156: Since this is not a test, but a test helper, it should be named starting with `check_` or something similar.

We should name this more appropriate to what it's doing because in the future we will add more common checks here that can be re-used in queue and queue guard. A good name would be `check_invalid_avail_idx`, and for the comment we can write that this is a test case that checks that an invalid `avail_idx` is not triggering a panic in the queue implementation.

crates/virtio-queue/Cargo.toml
DeliaPavel: I added this dependency because without it, when building with cargo build --release, I received an error:
error[E0432]: unresolved import `vm_memory::GuestMemoryMmap` from crates/virtio-queue/src/state.rs:14:60.

crates/virtio-queue/src/descriptor.rs
DeliaPavel: I tried this, but I receive this error: 
error[E0599]: no function or associated item named `new` found for struct `Descriptor` in the current scope
   --> crates/virtio-queue/src/state.rs:171:36. I added #[cfg(feature = "test-utils")] for both test_invalid_avail_idx from queue.rs and test_queue_guard_object from queue_guard.rs

crates/virtio-queue/src/state.rs
DeliaPavel: Done

crates/virtio-queue/src/state.rs
```diff
@@ -149,6 +150,87 @@ impl QueueState {
             .map(Wrapping)
             .map_err(Error::GuestMemory)
     }
+
+    /// This is a test case that checks that an invalid avail_idx is not triggering a panic in
+    /// the queue implementation.
+    pub fn check_invalid_avail_idx<M: GuestMemory>(
```
andreeaflorescu, line 156: This function does not belong here and that's reason for the other problems you're observing. You should revert all changes that add `GuestMemoryMmap` and the lines that were deleting the `#[cfg(any(test, feature = "test-utils"))]`. For now, let's just move this to `mod tests` in `queue.rs`. To be able to re-use, you'll need to make `mod tests` public, and then import the function in queue_quard.rs similarly to how you're doing it now.

DeliaPavel: Done

andreeaflorescu (commented):
LGTM, can you also fix the git conflict?

crates/virtio-queue/src/queue_guard.rs
```diff
@@ -200,86 +200,21 @@ where
...
-    use crate::defs::{VIRTQ_DESC_F_NEXT, VIRTQ_DESC_F_WRITE};
     use crate::mock::MockSplitQueue;
-    use crate::Descriptor;
 
+    use crate::queue::tests::check_invalid_avail_idx;
     use vm_memory::{GuestAddress, GuestMemoryMmap};
 
     #[test]
     fn test_queue_guard_object() {
         let m = &GuestMemoryMmap::<()>::from_ranges(&[(GuestAddress(0), 0x10000)]).unwrap();
```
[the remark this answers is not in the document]
DeliaPavel: Done

crates/virtio-queue/src/queue.rs
```diff
@@ -668,14 +665,15 @@ mod tests {
         vq.avail().ring().ref_at(2).store(u16::to_le(5));
         // Let the device know it can consume chains with the index < 2.
         vq.avail().idx().store(u16::to_le(3));
+
         // No descriptor chains are consumed at this point.
-        assert_eq!(q.next_avail(), 0);
-        assert_eq!(q.next_used(), 0);
+        assert_eq!(state.next_avail(), 0);
```
lauralt, line 670: The problem with this refactoring is that you are no longer calling these methods from the `Queue`/`QueueGuard` implementation, but always from the `QueueState` one. Because of this, having two separate tests doesn't actually make much sense anymore. I'll think about what we can do regarding this.

andreeaflorescu: I didn't find any solution to this without some serious refactoring. One way of achieving this is to implement the same trait on both QueueGuard and Queue, and the test case receive as a parameter a parametrized Q where Q implements the common trait. This is extending far beyond the scope of the opened issue. The code in Queue and QueueGuard is duplicated as well, but I think that needs to be handled as a separate issue

lauralt: > I didn't find any solution to this without some serious refactoring. One way of achieving this is to implement the same trait on both QueueGuard and Queue, and the test case receive as a parameter a parametrized Q where Q implements the common trait. This is extending far beyond the scope of the opened issue. The code in Queue and QueueGuard is duplicated as well, but I think that needs to be handled as a separate issue

I agree, so should we drop this patch for now?

Issue #144, opened 2022-03-03
Improve the usability of queue interfaces


alexandruag:
After adding the abstractions that facilitate working with queues in multi-threaded scenarios, looks like the only way to actually consume descriptor chains right now is via the `iter` method called on `QueueState` after locking, which returns an iterator. Moreover the iterator is now the only one with the `go_to_previous_position` method, because it’s guaranteed that only one consumer can have an iterator at any given time, so there’s no race condition. That generates some limitations:

- Holding the interator (so we can also call `go_to_previous_position` when necessary) automatically implies having a very large critical section, because no one else gets to pull descriptor chains out of the queue while an iterator is held.
- Multithreading-related concerns aside, there are a number of use cases where even with a simple `QueueState`, an iterator is unwieldy because it keeps `&mut self` borrowed while it lives, which leads to all sorts of problems with the borrow checker.

In terms of alleviating the descriptor chain consumption side of things, one simple way of addressing both the above concerns is introducing some sort of `pop_descriptor_chain method`. It generates a comparatively much smaller critical section (so multiple parallel consumers can make progress simultaneously), and also produces a descriptor chain without tying it to a` &mut self` borrow for the entire lifetime. Looks like adding this method is totally fine, as there are no potential race conditions. 

The trickier part is how to handle the equivalent of `go_to_previous_position`. We have seen at least several use cases where we first need to pop a descriptor chain from the queue, and then while processing it we realize there’s a blocker somewhere (for example, the backend cannot currently execute an operation for some reason), so we cannot mark it as used. This means we either need to return it to the queue, or store it somewhere. The former is complicated by race conditions in multithreaded scenarios, while the latter generated unwanted extra complexity. One straightforward way to address this is to add a companion trait to `QueueStateT` (named along the lines of `QueueStateOwnedT: QueueStateT` or something like that) which stands for an object that's owned by a single thread and thus can safely and directly implement operations such as `go_to_previous_position` and `iter`. Will post an example patch later, and looking forward to any other thoughts and perspectives here.

aghecenco:
#148

PR #148, opened 2022-03-04
Improve the usability of queue interfaces

alexandruag:
This RFC suggests a solution addressing the shortcomings mentioned in #144 (there are some very small orthogonal changes that got through as well). Not sure if the choice of naming for the new trait is the best, but the approach is to use that to explicitly gate operations that are only safe/sane when we know for sure accesses are not performed in a multi-threaded context.

### DIFF
--- a/crates/virtio-queue/src/iterator.rs
+++ b/crates/virtio-queue/src/iterator.rs
@@ -102,7 +102,7 @@ pub struct AvailIter<'b, M> {
 impl<'b, M> AvailIter<'b, M>
 where
     M: Deref,
-    M::Target: GuestMemory + Sized,
+    M::Target: GuestMemory,
 {
     /// Create a new instance of `AvailInter`.
     ///
--- a/crates/virtio-queue/src/lib.rs
+++ b/crates/virtio-queue/src/lib.rs
@@ -206,3 +206,20 @@ pub trait QueueStateT: for<'a> QueueStateGuard<'a> {
         M: Clone + Deref,
         M::Target: GuestMemory;
 }
+
+/// Trait to access and manipulate a Virtio queue that's known to be exclusively accessed
+/// by a single execution thread.
+pub trait QueueStateOwnedT: QueueStateT {
+    /// Get a consuming iterator over all available descriptor chain heads offered by the driver.
+    ///
+    /// # Arguments
+    /// * `mem` - the `GuestMemory` object that can be used to access the queue buffers.
+    fn iter<M>(&mut self, mem: M) -> Result<AvailIter<'_, M>, Error>
+    where
+        M: Deref,
+        M::Target: GuestMemory;
+
+    /// Undo the last advancement of the next available index field by decrementing its
+    /// value by one.
+    fn go_to_previous_position(&mut self);
+}
--- a/crates/virtio-queue/src/lib.rs
+++ b/crates/virtio-queue/src/lib.rs
@@ -155,7 +155,9 @@ pub trait QueueStateT: for<'a> QueueStateGuard<'a> {
     fn set_event_idx(&mut self, enabled: bool);
 
     /// Read the `idx` field from the available ring.
-    fn avail_idx<M: GuestMemory>(&self, mem: &M, order: Ordering) -> Result<Wrapping<u16>, Error>;
+    fn avail_idx<M>(&self, mem: &M, order: Ordering) -> Result<Wrapping<u16>, Error>
+    where
+        M: GuestMemory + ?Sized;
 
     /// Read the `idx` field from the used ring.
     fn used_idx<M: GuestMemory>(&self, mem: &M, order: Ordering) -> Result<Wrapping<u16>, Error>;
@@ -193,4 +195,14 @@ pub trait QueueStateT: for<'a> QueueStateGuard<'a> {
 
     /// Set the index for the next descriptor in the used ring.
     fn set_next_used(&mut self, next_used: u16);
+
+    /// Pop and return the next available descriptor chain, or `None` when there are no more
+    /// descriptor chains available.
+    ///
+    /// This enables the consumption of available descriptor chains in a "one at a time"
+    /// manner, without having to hold a borrow after the method returns.
+    fn pop_descriptor_chain<M>(&mut self, mem: M) -> Option<DescriptorChain<M>>
+    where
+        M: Clone + Deref,
+        M::Target: GuestMemory;
 }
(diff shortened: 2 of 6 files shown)


crates/virtio-queue/src/iterator.rs
```diff
@@ -102,7 +102,7 @@ pub struct AvailIter<'b, M> {
 impl<'b, M> AvailIter<'b, M>
 where
     M: Deref,
-    M::Target: GuestMemory + Sized,
+    M::Target: GuestMemory,
```
andreeaflorescu, line 105: Can you add a description about this in the commit message as well? It will help with writing the changelog.

crates/virtio-queue/src/lib.rs
```diff
@@ -190,4 +192,11 @@ pub trait QueueStateT: for<'a> QueueStateGuard<'a> {
 
     /// Set the index for the next descriptor in the used ring.
     fn set_next_used(&mut self, next_used: u16);
+
+    /// Pop and return the next available descriptor chain, or `None` when there are no more
```
andreeaflorescu, line 196: Can you also add some description about how this is suppose to be used (i.e. in respect to the already existing iterator; i think you mention this is an alternative way of consuming the iterator)?

andreeaflorescu (commented):
LGTM with a few tiny comments.

crates/virtio-queue/src/state.rs
```diff
@@ -426,15 +419,39 @@ impl QueueStateT for QueueState {
...
         self.next_used = Wrapping(next_used);
     }
+
+    fn pop_descriptor_chain<M>(&mut self, mem: M) -> Option<DescriptorChain<M>>
+    where
+        M: Clone + Deref,
+        M::Target: GuestMemory,
+    {
+        // Default, iter-based impl. Can most likely be improved.
+        self.iter(mem).ok()?.next()
```
lauralt, line 440: Should we for now log an error here instead of completely disregarding the error?

crates/virtio-queue/src/lib.rs
alexandruag: Added an extra phrase to the comment. Didn't mention the iterator, as it doesn't even exist strictly in the context of the `QueueStateT` trait.

crates/virtio-queue/src/state.rs
alexandruag: sure

crates/virtio-queue/src/iterator.rs
alexandruag: Hmm, I think it's better to relax the trait bounds for the other methods as well first via a separate PR, and add that to the changelog. Created #152 to track this.

lauralt:
Can you also add some tests? After that, I think it should be ready for merging.

crates/virtio-queue/src/iterator.rs
lauralt: Is this change required in this commit? If not, I would postpone it when fixing #152, if yes, I would still leave some sort of explanation about it in the commit message or at least give a pointer to #152.

crates/virtio-queue/src/lib.rs
```diff
@@ -190,4 +192,31 @@ pub trait QueueStateT: for<'a> QueueStateGuard<'a> {
...
+    /// # Arguments
+    /// * `mem` - the `GuestMemory` object that can be used to access the queue buffers.
+    fn iter<M>(&mut self, mem: M) -> Result<AvailIter<'_, M>, Error>
+    where
+        M: Deref,
+        M::Target: GuestMemory;
+
+    /// Undo the last advancement of the next available index field by decrementing its
+    /// value by one.
+    fn go_to_previous_position(&mut self);
```
jiangliu, line 221: This is still design flaw about `go_to_previous_position()`. There is no clear definition
whether caller could call `go_to_previous_position()` multiple times and what's the expected behavior.
All that we have is `all callers will only call it once`.

jiangliu:
This proposal looks good to me, except more work is needed about `go_to_previous_position()`.

crates/virtio-queue/src/lib.rs
alexandruag: This is an interesting and tricky to improve situation. Add the method here does not enable users of the crate to do anything that wasn't already possible before with `iter().go_to_previous_position()`. In terms of multiple consumers concurrently calling the method and thus ending in an undefined situation, this should not happen because the trait stands for queues that are handled by a single entity. You are right that even a single consumer can theoretically call this a sufficiently large number of times and break things, but it's not super clear right now how that can be avoided without making things difficult to use (also, `add_used` is in a similar situation, as it can be called any number of times with the same or invalid indices).

One approach can be to require the consumption of a `DescriptorChain` (that cannot be cloned, or manually instantiated, etc.) for these operations (but even that wouldn't be absolutely foolproof I think, because we can get a chain from another queue for example). We need to think a bit more about what a better solution can be here, but meanwhile I think the existing methods are serviceable and not that prone to misuse.

crates/virtio-queue/src/queue.rs
```diff
@@ -266,13 +267,24 @@ impl<M: GuestAddressSpace, S: QueueStateT> Queue<M, S> {
     pub fn set_next_used(&mut self, next_used: u16) {
         self.state.set_next_used(next_used);
     }
+
+    /// Pop and return the next available descriptor chain, or `None` when there are no more
+    /// descriptor chains available.
+    pub fn pop_descriptor_chain(&mut self) -> Option<DescriptorChain<M::T>> {
```
lauralt, line 273: Hmm the commit message says there are no methods added to `Queue`, but looks like there are some. Am I missing something?

alexandruag: You are definitely right, some methods from a previous iteration actually slipped through, I'll remove when with the next push, thanks for catching this!

alexandruag (commented):
@jiangliu I was wondering if the more verbose comment above regarding `go_to_previous_position` makes sense from your perspective. Basically I think there's a two-pronged answer here: functionality equivalent to `go_to_previous_position` on the state is already callable right now (via the iterator) so we're not weakening the current posture to any extent, and I think we need to take any bigger changes meant to add extra safety to chain handling as a separate topic for a subsequent iteration as they will likely have more profound implication on how the abstractions are used.

jiangliu (dismissed):

crates/virtio-queue/src/lib.rs
jiangliu: The flaw is not introduced by this PR, actually it already exists for a long time. So let's handle it separately.

alexandruag:
Rebased and pushed a small fix to address the last outstanding comment. Thank you everyone for the reviews!

lauralt (approved):

jiangliu (approved):

Issue #143, opened 2022-03-02
Provide a single Virtio Queue implementation


andreeaflorescu:
Currently we have 2 objects in vm-virtio that are providing mostly the same implementation: Queue and QueueGuard. The later was added in the context of supporting multi threaded queues, but its implementation is fairly similar to the one of the Queue.

This issue tracks having a single implementation that can be used by all customers. Currently Alibaba uses QueueGuard, while Cloud Hypervisor uses Queue.

More details TBD.

alexandruag:
For a bit more context, `Queue` and `QueueState` are in principle wrappers around some `QueueState` information and a handle to a memory object to perform the relevant accesses, with some differences: `Queue` owns the inner state object (i.e. it contains an `S: QueueStateT`) and the `M` in queue is an `M: GuestAddressSpace`  (which means that whenever an operation in `Queue` needs an `M: GuestMemory` it will go through the indirection of calling the `GuestAddressSpace::memory` to get an eventually-ephemeral handle). On the other hand, `QueueGuard` right now behaves as a sort of both wrapper and reference; the `M` within is just something that derefs  to a target type that implements `GuestMemory`, and the `S` within is just something that derefs to a `QueueState` ( in particular, it cannot own a state object unless it’s wrapped in something like an `Arc`, `Rc`, etc.).

Looks like what we want to do a bunch of stuff here. First see, out of all these abstractions, which ones do we want/need to keep. For example, there were discussions about getting rid of the current `Queue`, and turning what is now `QueueState` into `Queue`. We should also try to improve naming as part of this discussion. Last but not least, we need to improve the documentation to provide a clearer picture of the design and how abstractions are supposed to be used and interact with each other.

andreeaflorescu:
FIxed by #150

PR #149, opened 2022-03-04
add QueueStateSer

lauralt:
### Summary of the PR

Based on the design agreed [here](https://github.com/rust-vmm/community/pull/118) (and similar to what is already done in vm-superio), created a new crate, virtio-queue-ser, and added QueueStateSer. Once we have a fix for https://github.com/rust-vmm/vm-virtio/issues/143, this PR will probably need some updates, but high level it should remain the same. Marking this as draft until that.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/crates/virtio-queue-ser/src/state.rs
+++ b/crates/virtio-queue-ser/src/state.rs
@@ -1,0 +1,126 @@
+// Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
+
+use std::num::Wrapping;
+
+use serde::{Deserialize, Serialize};
+use versionize::{VersionMap, Versionize, VersionizeResult};
+use versionize_derive::Versionize;
+use virtio_queue::QueueState;
+use vm_memory::GuestAddress;
+
+/// Wrapper over a `QueueState` that has serialization capabilities.
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize, Versionize)]
+pub struct QueueStateSer {
+    /// The maximum size in elements offered by the device.
+    pub max_size: u16,
+    /// Tail position of the available ring.
+    pub next_avail: u16,
+    /// Head position of the used ring.
+    pub next_used: u16,
+    /// VIRTIO_F_RING_EVENT_IDX negotiated.
+    pub event_idx_enabled: bool,
+    /// The last used value when using VIRTIO_F_EVENT_IDX.
+    pub signalled_used: Option<u16>,
+    /// The queue size in elements the driver selected.
+    pub size: u16,
+    /// Indicates if the queue finished with the configuration.
+    pub ready: bool,
+    /// Guest physical address of the descriptor table.
+    pub desc_table: u64,
+    /// Guest physical address of the available ring.
+    pub avail_ring: u64,
+    /// Guest physical address of the used ring.
+    pub used_ring: u64,
+}
+
+// The following `From` implementations can be used to convert from a `QueueStateSer` to the
+// `QueueState` from the base crate and vice versa.
+// WARNING: They don't check for any invalid data, that would otherwise be checked when initializing
+// a queue object from scratch (for example setting a queue size greater than its max possible
+// size). The problem with the current `QueueState` implementation is that it can be used as the
+// queue object itself. `QueueState`'s fields are public, which allows the user to set up and use an
+// invalid queue. Once we fix https://github.com/rust-vmm/vm-virtio/issues/143, `QueueState` will be
+// renamed to `Queue`, its fields will no longer be public, and `QueueState` will consist of the
+// actual state. This way we can also check against any invalid data when trying to get a `Queue`
+// from the state object.
+// Nevertheless, we don't make any assumptions in the virtio-queue code about the queue's state that
+// would otherwise result in a panic, when initialized with random data, so from this point of view
+// these conversions are safe to use.
+impl From<&QueueStateSer> for QueueState {
+    fn from(state: &QueueStateSer) -> Self {
+        QueueState {
+            max_size: state.max_size,
+            next_avail: Wrapping(state.next_avail),
+            next_used: Wrapping(state.next_used),
+            event_idx_enabled: state.event_idx_enabled,
+            signalled_used: state.signalled_used.map(Wrapping),
+            size: state.size,
+            ready: state.ready,
+            desc_table: GuestAddress(state.desc_table),
+            avail_ring: GuestAddress(state.avail_ring),
+            used_ring: GuestAddress(state.used_ring),
+        }
+    }
+}
+
+impl From<&QueueState> for QueueStateSer {
+    fn from(state: &QueueState) -> Self {
+        QueueStateSer {
+            max_size: state.max_size,
+            next_avail: state.next_avail.0,
+            next_used: state.next_used.0,
+            event_idx_enabled: state.event_idx_enabled,
+            signalled_used: state.signalled_used.map(|value| value.0),
+            size: state.size,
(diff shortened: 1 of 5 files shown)


lauralt:
Since it is actually going to take some time to agree on the path forward with https://github.com/rust-vmm/vm-virtio/pull/150, I propose we review/merge this one before, and I can take care afterwards of any other changes to the state (if there will be any) that can be needed after the interface changes. Marking this as ready for review.

crates/virtio-queue-ser/src/state.rs
```diff
@@ -0,0 +1,115 @@
...
+use virtio_queue::QueueState;
+use vm_memory::GuestAddress;
+
+/// /// Wrapper over a `QueueState` that has serialization capabilities.
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize, Versionize)]
+pub struct QueueStateSer {
+    /// The maximum size in elements offered by the device.
+    pub max_size: u16,
+    /// Tail position of the available ring.
+    pub next_avail: Wrapping<u16>,
```
andreeaflorescu, line 19: I think we should completely separate state from functionality, as we already stated in the generic save/restore documentation. The fact that this is wrapping is a a matter of function, not state. In theory we should be able to save this directly as an u16.

crates/virtio-queue-ser/src/state.rs
```diff
@@ -0,0 +1,115 @@
...
+            ready: state.ready,
+            desc_table: GuestAddress(state.desc_table),
+            avail_ring: GuestAddress(state.avail_ring),
+            used_ring: GuestAddress(state.used_ring),
+        }
+    }
+}
+
+impl From<&QueueState> for QueueStateSer {
+    fn from(state: &QueueState) -> Self {
```
andreeaflorescu, line 58: I wonder if there is anything that we need to check here. Is there any invalid state that would otherwise not be able to get into when just initializing a Queue from scratch?

andreeaflorescu, line 58: I wonder if there is anything that we need to check here. Is there any invalid state that would otherwise not be able to get into when just initializing a Queue from scratch?

lauralt: When initializing a Queue from scratch, there are some checks that are done on the queue size, on the addresses of the queue components (in their corresponding setters). So yes, the state here can be invalid if the user is initializing it with random values.
This is actually a problem at the `QueueState` level and related to all that confusion around the queue interfaces. The `QueueState`s fields are public, which is okay for a state object, but since the `QueueState` is actually the `Queue`, the fact that its fields are public is no longer good.
I guess for now I can bring the logic from those setters here as well, but that won't fix the entire problem, and is not a clean fix either. I think the clean fix would be to have the `QueueState` -> `Queue` interface update (and not having the queue fields public), and when trying to get a `Queue` from a `QueueState`, do all of the checks from those setters. That way, the code in this file wouldn't require any checks.

lauralt: > Why just wonder and not reason about it?
> 
> For example, my thinking is even if _some_ of the fields could be sanity checked (`size <= max_size` for example), I suggest deferring validations to when the actual `Queue` is built, since without `GuestMemory` most of the state doesn't make any sense.

That is sort of my suggestion as well. The thing here is the way the queue interfaces are right now, you can more easily use `QueueState` as the Queue object. The queue object doesn't need to have a memory handle in its composition. [This](https://github.com/rust-vmm/vm-virtio/pull/150#issuecomment-1064706223) is basically one of the things we want to update, and after that the `where should I check stuff` discussion would be much easier to have. 

> Then you can rightly call its [is_valid()](https://github.com/rust-vmm/vm-virtio/blob/2f01935f959485d7521d0bb57b79da644d2053ea/crates/virtio-queue/src/state.rs#L174) method to make sure all is kosher.
> 
> I too am lazy to point out exactly what extra checks `is_valid()` would need when also considering deserialization, but once you get there you can look at [Firecracker impl](https://github.com/firecracker-microvm/firecracker/blob/f2743bfd2e9f9740dd08e55cdb83ba7b94aabb69/src/devices/src/virtio/queue.rs#L224) to get an idea.

In virtio-queue, `is_valid` is no longer doing all those checks that are done in Firecracker. The checks that are part of the spec were moved to the corresponding setters of those fields, because as opposed to `is_valid`, which is optional for the user to call, the setters are not (when starting from scratch). But yes, this is the way to do the validations after deserialization.

acatangiu: Lgtm! But as a heads-up, you might have an easier time if you created a `QueueBuilder` object where you'd have `fn with_x(self, x) -> Self` instead of setters you mention above.
The advantage is that ultimately you have `fn build(self) -> Result<Queue, Error<Self>>` that can do more complex validations that cannot be done in simple setters.

Validations depending on `X` **and** `Y` cannot be done in `set_x` or `set_y` (for example [this check](https://github.com/firecracker-microvm/firecracker/blob/f2743bfd2e9f9740dd08e55cdb83ba7b94aabb69/src/devices/src/virtio/queue.rs#L278) can't be done in a setter, but can be done in `qbuilder.build()` or `queue.is_valid()`).

lauralt: > Lgtm! But as a heads-up, you might have an easier time if you created a `QueueBuilder` object where you'd have `fn with_x(self, x) -> Self` instead of setters you mention above. The advantage is that ultimately you have `fn build(self) -> Result<Queue, Error<Self>>` that can do more complex validations that cannot be done in simple setters.
> 
We thought about using the builder pattern before 0.1.0, but it's a bit incompatible with how the queue is constructed when using for example MMIO writes (and there won't be any `build()` in this case; calling `build()` will still be optional).

> Validations depending on `X` **and** `Y` cannot be done in `set_x` or `set_y` (for example [this check](https://github.com/firecracker-microvm/firecracker/blob/f2743bfd2e9f9740dd08e55cdb83ba7b94aabb69/src/devices/src/virtio/queue.rs#L278) can't be done in a setter, but can be done in `qbuilder.build()` or `queue.is_valid()`).

Not all de validations are supposed to happen in the setters, but what we are aiming with this is not having the MUSTs from the spec as optional.

acatangiu: > Not all de validations are supposed to happen in the setters, but what we are aiming with this is not having the MUSTs from the spec as optional.

Sure! Validating in the setters against the spec is great!
I'm just saying as a heads-up that, on the slightly orthogonal subject of deserialization, you will need _some extra_ validations that you can't do in setters.

andreeaflorescu: I personally dislike the pattern with `is_valid`. The callers of the interface should not know that they **have** to call `is_valid`. But this should be part of a bigger refactoring effort.

If you're using the current interface correctly (i.e. you don't set arbitrary values to the queue fields, but just use the setters), then the Queue that you are getting is a valid one. From my point of view, this is something that we need to check when initializing from a saved state as well. We can split the checks in 2 categories:
- what can malicious actors do
- what can happen because of programming errors

Right now with the current interface is really hard to guard against programming errors because the interface is too open (i.e. with the public fields). Nevertheless, we should have checks for things that can be initialized by malicious actors. As per the threat model the deserialized state is not trusted because we cannot know where it originates from. If we cannot address this in this PR it's fine, but we should know about this shortcoming and at least document it.

crates/virtio-queue-ser/src/state.rs
lauralt: done

crates/virtio-queue-ser/src/state.rs
lauralt: Added some more clarifications in code comments.

andreeaflorescu (approved):

PR #150, opened 2022-03-07
Export a single Queue implementation

andreeaflorescu:
### Summary of the PR

The purpose of this PR is to reduce the code duplication between the tests and code in `QueueState`, `Queue` and `QueueGuard`, and export a single queue implementation. The `Queue` interface right now it too open for change, which means that we can end up with the queue in invalid states. Creating an invalid queue yields no error and during the operation of the queue we can hit runtime errors like overflows, division by 0, or simply undefined behavior. To account for these, the queue inner state can only be changed through setters that are checking the validity of the input.

The design choices and reasons behind them are documented in the respective commits. As this is a pretty big PR, the best way to review it is commit by commit which should give enough context about the change.

### Requirements

requirements:

  description line.
  test.


andreeaflorescu:
Once we agree with the approach, I'll also go ahead and update the tests that are currently in `queue.rs` to make them depend on `QueueStateT` instead. This will help with testing both `Queue` and `QueueGuard` with the same test cases.

jiangliu:
> Once we agree with the approach, I'll also go ahead and update the tests that are currently in `queue.rs` to make them depend on `QueueStateT` instead. This will help with testing both `Queue` and `QueueGuard` with the same test cases.

It would be great to remove `Queue` and rename `QueueState` to `Queue`, but that may be a break change to Cloud Hypervisor. @sboeuf

sboeuf:
> > Once we agree with the approach, I'll also go ahead and update the tests that are currently in `queue.rs` to make them depend on `QueueStateT` instead. This will help with testing both `Queue` and `QueueGuard` with the same test cases.
> 
> It would be great to remove `Queue` and rename `QueueState` to `Queue`, but that may be a break change to Cloud Hypervisor. @sboeuf

It will certainly be a breaking change, but we recently moved Cloud Hypervisor to `virtio-queue` v0.2.0, meaning virtio-queue can introduce breaking changes, Cloud Hypervisor will fix what needs to be fixed once it moves to the version 0.3.0 (which will hopefully contain the breaking changes).

jiangliu:
There are two sets of interfaces:
- Queue/QueueGuard
- QueueState/QueueStateGuard

The only difference between these two sets of interface is that `Queue/QueueGuard` has an embedded `GuestAddressSpace` object, so the caller does not need to passing `mem: &GuestMemory` explicitly.

So if the plan is to remove `Queue`, we should remove all of `Queue`, `QueueGuard` and `QueueStateT::lock_with_memory()`.

sboeuf:
Hi @andreeaflorescu, I've just had a look at the latest version you submitted. This introduces important API breaking changes.
Changing a name is one thing, but you moved away from the design where the Queue had a hold onto the guest memory. And because of this, you had to update a bunch of methods so they now expect the caller to provide this handle to the guest memory.

IIRC having the Queue holding a handle to the guest memory was an improvement that we introduced at some point (wasn't there at the beginning) and it caused a breaking API change at the time. So now it feels we have to revert all of what we did to make this work.

I just want to make sure moving away from that design is something that we all (/cc @jiangliu @slp @bonzini @rbradford) think will help in a way, because I don't want to spend some time adapting Cloud Hypervisor code (again) if there's no benefit.

andreeaflorescu:
@sboeuf @jiangliu @alexandruag @slp @bonzini @rbradford any updates on this?

I have a few follow-up PRs after this one as well that are helping with improving the security of this crate. We are now running fuzzing for this crate (we will also upstream the fuzzer in the near future), and some of the problems we're discovering are related to the design of this crate.

Using the Virtio Queue interface in a secure manner requires more work from the VMM that it would be necessary. When we implemented this we did not take an approach of secure by default, and the interface can be misused because of the fields of the Virtio Queue being public. While all these findings are low risk (because the guests can only shoot themselves in the foot), we would like to improve the design and offer an interface that does not allow misuse.

Specifically, these are the changes that I still have in my pipeline:
- All fields of the `Queue` are private and can only be set by using setters which are checking the validity of the values. For example, this finding #173 would not be possible if we would only set the queue size using the [set_size](https://github.com/rust-vmm/vm-virtio/blob/b47a799d3b317034c7e95efa98737792d891f90d/crates/virtio-queue/src/state.rs#L248) function.
- To be able to initialize the `Queue` from a previously saved state, we're going to define a `QueueState` where all the fields are public. This will represent a pure state of the queue that holds no information about the implementation details.

Is there anything that we can do to minimize the changes required in Cloud Hypervisor, or anything else we can help with so that this can move forward?

rbradford:
>     * To be able to initialize the `Queue` from a previously saved state, we're going to define a `QueueState` where all the fields are public. This will represent a pure state of the queue that holds no information about the implementation details.
> 
> 
> Is there anything that we can do to minimize the changes required in Cloud Hypervisor, or anything else we can help with so that this can move forward?

After the removal of the `Clone` implementation we now have to do this:

https://github.com/cloud-hypervisor/cloud-hypervisor/blob/7db741092091ce76466fe75fee02113be9e2b079/vm-virtio/src/lib.rs#L128

I think a `new_from_queue()` would be a nicer way to do this without having to make the internals public.

andreeaflorescu:
> I think a `new_from_queue()` would be a nicer way to do this without having to make the internals public.

Yap, I have something similar in mind. We can have the `QueueState` with all fields public, and then have a `try_from` function to convert to a `Queue`. The `try_from` can returns errors though because we are considering the queue state to be untrusted as it can be deserialized from untrusted sources.

rbradford:
> > I think a `new_from_queue()` would be a nicer way to do this without having to make the internals public.
> 
> Yap, I have something similar in mind. We can have the `QueueState` with all fields public, and then have a `try_from` function to convert to a `Queue`. The `try_from` can returns errors though because we are considering the queue state to be untrusted as it can be deserialized from untrusted sources.

Okay. Sounds reasonable. I would propose you include that in this PR and then we can do a test run of porting CH to use it and that would tell us if its sufficient. (As well as the ability to derive a new `Queue` from an existing `Queue` we will need to do the same from a serialisable state for snapshot-restore/migration.)

andreeaflorescu:
Did another rebase, I am not expecting big changes in the near future that need to go before this PR, so I'll reserve time to get this closer to what my vision was regarding the updated interface. This is going to be quite a big change though, which is why I was a bit reluctant to do it in one go.

andreeaflorescu:
Ping @rbradford @jiangliu @sboeuf can you take another look? Now the PR is ready for review, the logic is all in there, I just need to do some non-functional updates. The rebases for this PRs are becoming quite hard to manage, and I would really appreciate some feedback.

sboeuf:
I can have a look next week :)

andreeaflorescu:
As this is quite a large PR I recommend going commit by commit. I tried to add as many information as I could in the commit message to explain the motivation behind the changes.

sboeuf:
@andreeaflorescu from a quick first check of the changes you've introduced, the only thing missing would be a helper method `state()` for the `Queue` structure that would return a `QueueState`. This would be convenient and would give us the opposite of what `try_from()` provides to turn a `QueueState` into a `Queue`.

andreeaflorescu:
> @andreeaflorescu from a quick first check of the changes you've introduced, the only thing missing would be a helper method `state()` for the `Queue` structure that would return a `QueueState`. This would be convenient and would give us the opposite of what `try_from()` provides to turn a `QueueState` into a `Queue`.

Good point, I think I've missed that. We were favoring the use of `try_from` or `from`, but I can also add a `state` function.

With this PR the tests are probably going to fail because I didn't got a chance to do a rebase with the bug fix from #177. I will reserve some time to do that today.

andreeaflorescu:
@sboeuf I pushed another revision with adds the `QueueState::from(Queue)`. If you prefer to have a `state` function in `Queue`, please let me know and I'll update it. I also did the rebase so now the tests should be passing.

Let me know if there's anything else that you'd need. Btw, I am not sure if you're aware, but we also have a crate that provides version aware serialization for the queue in `virtio-queue-ser` in case you would prefer to consume that directly instead of having an implementation yourself.

andreeaflorescu:
The problem with the `From` approach is that it's consuming the `Queue`. In case we want this to not be consume we'll need to implement a `state` function on the Queue instead.

sboeuf:
> The problem with the `From` approach is that it's consuming the `Queue`. In case we want this to not be consume we'll need to implement a `state` function on the Queue instead or implement From for `&Queue`.

Consuming is not what I had in mind. Having a `state()` method would basically generate the `QueueState`, representing a snapshot of the Queue at this time.

andreeaflorescu:
@sboeuf I pushed the latest changes and added a `state` function, I will try to do a PR in vhost-user-backend as well so we can test this end to end in CH as well.

sboeuf:
> @sboeuf I pushed the latest changes and added a `state` function, I will try to do a PR in vhost-user-backend as well so we can test this end to end in CH as well.

Thanks that's awesome! I'll be out next week, but I can update the port/POC the week after, based on both your `virtio-queue` and `vhost-user-backend` PRs :)

stsquad (approved):

coverage_config_x86_64.json
```diff
@@ -1,5 +1,5 @@
 {
-  "coverage_score": 88.6,
+  "coverage_score": 87.2,
```
stsquad, line 2: The coverage is dropping here quite a bit. Do we know what's getting missed in the new paths?

stsquad (dismissed):

coverage_config_x86_64.json
andreeaflorescu: Good catch! I ran the coverage test locally and looked through the report. There are 4 functions in QueueSync added that are not covered by unit tests:
- desc_table
- avail_ring
- used_ring
- event_idx_enabled
These are getters that are calling into the `Queue` interface which is already covered by unit tests. For redundancy, I'll add some tests for these ones as well.

Then, there are also some new error types defined in this PR for which we don't have unit tests, so that's also contributing to the drop.

Besides these functions, I also deleted a lot of code that was previously covered by unit tests, so presumably that's also part of the reason why we're seeing this drop.

stsquad: Otherwise it looks pretty good to me. I'm looking at the vhost-user-backend work in progress now. The remaining build errors seems to be all about satisfying trait bounds so I'll need to get my "thinking like the compiler" hat on ;-)

andreeaflorescu: I think the problem is how I added the `mem` field in fact. I might be able to think about it a bit more next week if nobody else has time to pick it up.

andreeaflorescu: @stsquad I added unit tests for the new getters and setters and now the coverage is up to 88.5.

sboeuf: @andreeaflorescu @stsquad please let me know when you'll have the `vhost-user-backend` PR ready so that I can fully port Cloud Hypervisor to this new version of `virtio-queue` and run the proper testing on it.

crates/virtio-queue-ser/src/state.rs
```diff
@@ -46,78 +42,79 @@ pub struct QueueStateSer {
 // Nevertheless, we don't make any assumptions in the virtio-queue code about the queue's state that
 // would otherwise result in a panic, when initialized with random data, so from this point of view
 // these conversions are safe to use.
-impl From<&QueueStateSer> for Queue {
```
lauralt: The comments above should be updated.

coverage_config_x86_64.json
andreeaflorescu: @sboeuf I managed to integrate the changes in vhost-user-backend: https://github.com/andreeaflorescu/vhost-user-backend/tree/update_virtio_queue. I've also opened a PR: https://github.com/rust-vmm/vhost-user-backend/pull/78

sboeuf: Cool! I'll have a try then!

coverage_config_x86_64.json
```diff
@@ -1,5 +1,5 @@
 {
-  "coverage_score": 88.6,
```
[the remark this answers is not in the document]
andreeaflorescu: Which commit are you referring to? In the last commit where I am updating the coverage, the coverage update comes from writing the tests. Is there an update in another commit? I can't seem to find it now.

crates/virtio-queue-ser/src/state.rs
```diff
@@ -143,9 +138,8 @@ mod tests {
             used_ring: 276,
         };
 
-        let mut queue = Queue::from(&queue_ser);
-        let desc_chain = vec![Descriptor::new(0x0, 0x100, 0, 0)];
-        vq.build_desc_chain(&desc_chain).unwrap();
-        assert!(queue.pop_descriptor_chain(m).is_none());
+        let queue_state = QueueState::from(&queue_ser);
+        let err = Queue::try_from(queue_state).unwrap_err();
+        assert_eq!(err.to_string(), Error::InvalidSize.to_string());
```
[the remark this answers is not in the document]
andreeaflorescu: This is only possible after the last commit that adds the `PartialEq` implementation for the queue error. I'll have the change part of that commit.

coverage_config_x86_64.json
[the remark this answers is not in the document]
andreeaflorescu: Thanks, done.

lauralt (approved):

slp (approved):
LGTM. Thanks @andreeaflorescu !

sboeuf (approved):
Yep it looks good to me as well. I don't want to hold this PR for a longer time.

andreeaflorescu:
Thanks everyone for the reviews!

sboeuf:
@andreeaflorescu are you gonna cut a release?

andreeaflorescu:
> @andreeaflorescu are you gonna cut a release?

Yes, will be working on that first thing on Monday.

PR #151, opened 2022-03-08
queue/mock: add method for adding a desc chain to the queue

lauralt:
### Summary of the PR

Added a new method to the mock queue that can be used to add a descriptor chain at any offset in the table. This way multiple chains can be added to the queue. This wasn't possible before with `build_desc_chain` which can add a chain only at the beginning of the descriptor table.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/crates/virtio-queue/src/mock.rs
+++ b/crates/virtio-queue/src/mock.rs
@@ -11,7 +11,10 @@ use vm_memory::{
     Address, ByteValued, Bytes, GuestAddress, GuestAddressSpace, GuestMemory, GuestUsize,
 };
 
-use crate::defs::{VIRTQ_DESC_F_INDIRECT, VIRTQ_DESC_F_NEXT};
+use crate::defs::{
+    VIRTQ_AVAIL_ELEMENT_SIZE, VIRTQ_AVAIL_RING_HEADER_SIZE, VIRTQ_DESC_F_INDIRECT,
+    VIRTQ_DESC_F_NEXT,
+};
 use crate::{Descriptor, DescriptorChain, Queue, QueueState, VirtqUsedElem};
 
 /// Wrapper struct used for accessing a particular address of a GuestMemory area.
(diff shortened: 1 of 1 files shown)


crates/virtio-queue/src/mock.rs
```diff
@@ -368,17 +371,29 @@ impl<'a, M: GuestMemory> MockSplitQueue<'a, M> {
         q
     }
 
-    /// Writes a descriptor chain to the memory object of the queue and returns the associated
-    /// `DescriptorChain` object.
-    // `descs` represents a slice of `Descriptor` objects which are used to populate the chain.
+    /// Writes a single descriptor chain to the memory object of the queue, at the beginning of the
+    /// descriptor table, and returns the associated `DescriptorChain` object.
+    pub fn build_desc_chain(&self, descs: &[Descriptor]) -> DescriptorChain<&M> {
```
andreeaflorescu, line 376: Why do we still need this function? It is not clear to me how you have to use build_desc_chain and add_desc_chain. From the name it sounds like you can use it as:
- first create a descriptor chain using `build_desc_chain`
- then you can add more descriptors using `add_desc_chain`
Is this assumption correct?

lauralt: Not necessarily, the `build` one is also returning that chain by iterating the queue, the second one can be used to simply populate the queue with a new chain. I kept the old one because it helps in tests to also have that chain returned. The naming is pretty bad, but since I didn't necessarily have a much better idea of a name for this one, I chose to at least keep it the same as before so that there are no other name changes involved in the PR.

lauralt: You can use `add_desc_chain` without any `build_desc_chain` call. Basically what `build_desc_chain` does differently is that is also starts to iterate over the queue.

andreeaflorescu: I find this to be a bit strange, but let's just merge it an fix it when we're extending the mocking framework.

andreeaflorescu (approved):

PR #153, opened 2022-03-13
Minor refinement to QueueState and QueueStateT

jiangliu:
### Summary of the PR

1) Remove `#[derive(Clone)]` for `QueueState`.
2) Add method `size()` to `QueueStateT`.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/crates/virtio-queue/CHANGELOG.md
+++ b/crates/virtio-queue/CHANGELOG.md
@@ -1,3 +1,11 @@
+# Unreleased
+
+## Added
+- QueueStateT::size(): [#153](https://github.com/rust-vmm/vm-virtio/pull/153)
+
+## Removed
+- #[derive(Clone)] for QueueState: [#153](https://github.com/rust-vmm/vm-virtio/pull/153)
+
 # v0.2.0
 
 ## Added
--- a/crates/virtio-queue/src/state.rs
+++ b/crates/virtio-queue/src/state.rs
@@ -22,7 +22,7 @@ use crate::defs::{
 use crate::{error, AvailIter, Descriptor, Error, QueueStateGuard, QueueStateT, VirtqUsedElem};
 
 /// Struct to maintain information and manipulate state of a virtio queue.
-#[derive(Clone, Debug)]
+#[derive(Debug)]
 pub struct QueueState {
     /// The maximum size in elements offered by the device.
     pub max_size: u16,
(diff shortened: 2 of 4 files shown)


crates/virtio-queue/src/state.rs
```diff
@@ -242,6 +242,10 @@ impl QueueStateT for QueueState {
         self.max_size
     }
 
+    fn actual_size(&self) -> u16 {
+        std::cmp::min(self.max_size, self.size)
```
lauralt, line 246: Based on some recent changes, the `size` of the queue can not be greater than `max_size` anymore, it is documented [here](https://github.com/rust-vmm/vm-virtio/blob/main/crates/virtio-queue/src/queue.rs#L127). [Here](https://github.com/rust-vmm/vm-virtio/pull/112/commits/3c84b4cca454de1b3ee2d77bfe02cdd60ad6c644) is the commit where the `actual_size` was removed. If we want to have a method that returns the size, can we rename it to `size()` and return `self.size`?

jiangliu: Sure, will rename it to `size()`.

crates/virtio-queue/src/state.rs
```diff
@@ -242,6 +242,10 @@ impl QueueStateT for QueueState {
         self.max_size
     }
 
+    fn size(&self) -> u16 {
```
lauralt, line 245: Just wondering, the reason for not adding this method to `Queue` and `QueueGuard` too is the upcoming change from https://github.com/rust-vmm/vm-virtio/pull/150?

crates/virtio-queue/CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+# Unreleased
+
+## Added
+- QueueStateT::size(): [#153](https://github.com/rust-vmm/vm-virtio/pull/153)
+
+## Removed
+- #[derive(Clone)] for QueueState: [#153](https://github.com/rust-vmm/vm-virtio/pull/153)
```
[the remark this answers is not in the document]
jiangliu: Do we need to enforce 80-chars-per-line rule for markdown files?

crates/virtio-queue/src/state.rs
jiangliu: yes, to avoid confliction.

crates/virtio-queue/CHANGELOG.md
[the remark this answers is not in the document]
lauralt: This is a good practice we've been adhering to, but it's not a hard requirement I would say.

andreeaflorescu (approved):

PR #159, opened 2022-04-08
virtio-queue: build: Relax vm-memory dependency

rbradford:
This allows the use of newer versions of vm-memory and matches with
other crates (e.g. vhost, linux-loader, etc) who also have such a
relaxed vm-memory dependency.

Signed-off-by: Rob Bradford <robert.bradford@intel.com>


crates/virtio-queue/Cargo.toml
```diff
@@ -13,13 +13,13 @@ edition = "2018"
 test-utils = []
 
 [dependencies]
-vm-memory = "0.7.0"
+vm-memory = ">=0.7.0"
```
lauralt, line 16: Should we also update to >=0.8.0 if it's the newest release at this point?

lauralt (dismissed):

crates/virtio-queue/Cargo.toml
rbradford: Great question. I intentionally didn't do that since that version is not strictly required for the functionality of this crate to work.

rbradford:
I had to update the PR since there were other uses of `vm-memory` in this repository. But I think you have a fundamental problem with the `virtio-vsock` crate as that has a strict dependency on `virtio-queue` 0.2.0 which will be pulled from crates.io with the fixed `vm-memory` version.

lauralt:
> I had to update the PR since there were other uses of `vm-memory` in this repository. But I think you have a fundamental problem with the `virtio-vsock` crate as that has a strict dependency on `virtio-queue` 0.2.0 which will be pulled from crates.io with the fixed `vm-memory` version.

Thanks for the heads up, we will probably have a release of virtio-queue sooner than one for the vsock, and we can update then the virtio-queue dependency with 0.3.0.

rbradford:
> > I had to update the PR since there were other uses of `vm-memory` in this repository. But I think you have a fundamental problem with the `virtio-vsock` crate as that has a strict dependency on `virtio-queue` 0.2.0 which will be pulled from crates.io with the fixed `vm-memory` version.
> 
> Thanks for the heads up, we will probably have a release of virtio-queue sooner than one for the vsock, and we can update then the virtio-queue dependency with 0.3.0.

The CI is failing because of this very problem because something is trying to link to two the two versions. I think this probably needs manual intervention of making a new release of the virtio-queue crate with the relaxed dependencies.

rbradford:
> > > I had to update the PR since there were other uses of `vm-memory` in this repository. But I think you have a fundamental problem with the `virtio-vsock` crate as that has a strict dependency on `virtio-queue` 0.2.0 which will be pulled from crates.io with the fixed `vm-memory` version.
> > 
> > 
> > Thanks for the heads up, we will probably have a release of virtio-queue sooner than one for the vsock, and we can update then the virtio-queue dependency with 0.3.0.
> 
> The CI is failing because of this very problem because something is trying to link to two the two versions. I think this probably needs manual intervention of making a new release of the virtio-queue crate with the relaxed dependencies.

One solution would be to make the `virtio-vsock` crate use the `virtio-queue` crate in the repo instead. This is how the `virtio-blk` crate does it.

lauralt:
> > > > I had to update the PR since there were other uses of `vm-memory` in this repository. But I think you have a fundamental problem with the `virtio-vsock` crate as that has a strict dependency on `virtio-queue` 0.2.0 which will be pulled from crates.io with the fixed `vm-memory` version.
> > > 
> > > 
> > > Thanks for the heads up, we will probably have a release of virtio-queue sooner than one for the vsock, and we can update then the virtio-queue dependency with 0.3.0.
> > 
> > 
> > The CI is failing because of this very problem because something is trying to link to two the two versions. I think this probably needs manual intervention of making a new release of the virtio-queue crate with the relaxed dependencies.
> 
> One solution would be to make the `virtio-vsock` crate use the `virtio-queue` crate in the repo instead. This is how the `virtio-blk` crate does it.

Yes, we can also use `path` together with `version`. But it's a bit strange, the crates should be as independent as possible, i.e. virtio-vsock build with virtio-queue 0.2.0, and the others with the current HEAD.

rbradford:
> One solution would be to make the `virtio-vsock` crate use the `virtio-queue` crate in the repo instead. This is how the `virtio-blk` crate does it.

My PR now has a patch that does this. After releasing the `virtio-queue` crate with the relaxed version you can go back to using crates.io

lauralt:
> > One solution would be to make the `virtio-vsock` crate use the `virtio-queue` crate in the repo instead. This is how the `virtio-blk` crate does it.
> 
> My PR now has a patch that does this. After releasing the `virtio-queue` crate with the relaxed version you can go back to using crates.io

In this case, can you also update to >=0.8.0? 0.8.0 is fixing a vm-memory bug that can affect the vsock crate, and it would be better to enforce 0.8.0 here.

lauralt (approved):

crates/devices/virtio-vsock/Cargo.toml
```diff
@@ -10,9 +10,9 @@ license = "Apache-2.0 OR BSD-3-Clause"
 edition = "2018"
 
 [dependencies]
-virtio-queue = "0.2.0"
-vm-memory = "0.7.0"
+virtio-queue = { path = "../../virtio-queue" }
```
jiangliu, line 13: Should this be
```
virtio-queue = { version = "0.3", path = "../../virtio-queue" }
```

lauralt: It should, but we won't be allowed to publish it anyway, so I was thinking to update it at that point.

rbradford: @jiangliu I just make it consistent with the other crates.

sboeuf (approved):

PR #162, opened 2022-04-26
Add changes to mock.rs

DeliaPavel:
Add error handling
Add function called build_chain_of_descriptors
that receives a vector of Descriptor structures
and creates a chain of them.

Signed-off-by: Delia-Maria Pavel <delia_maria.pavel@stud.acs.upb.ro>

### Summary of the PR

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/crates/virtio-queue/src/mock.rs
+++ b/crates/virtio-queue/src/mock.rs
@@ -16,6 +16,35 @@ use crate::defs::{
     VIRTQ_DESC_F_NEXT,
 };
 use crate::{Descriptor, DescriptorChain, Queue, QueueState, VirtqUsedElem};
+use std::fmt::{self, Debug, Display};
+
+/// Mock related errors.
+#[derive(Debug)]
+pub enum MockError {
+    /// Invalid Ref index
+    InvalidIndex,
+    /// Invalid next avail
+    InvalidNextAvail,
+}
+
+impl Display for MockError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        use self::MockError::*;
+
+        match self {
+            InvalidIndex => write!(
+                f,
+                "invalid index for pointing to an address in a region when defining a Ref object"
+            ),
+            InvalidNextAvail => write!(
+                f,
+                "invalid next available descriptor chain head in the queue"
+            ),
+        }
+    }
+}
+
+impl std::error::Error for MockError {}
 
 /// Wrapper struct used for accessing a particular address of a GuestMemory area.
 pub struct Ref<'a, M, T> {
@@ -64,16 +93,17 @@ impl<'a, M: GuestMemory, T: ByteValued> ArrayRef<'a, M, T> {
 
     /// Return a `Ref` object pointing to an address defined by a particular
     /// index offset in the region.
-    pub fn ref_at(&self, index: usize) -> Ref<'a, M, T> {
-        // TODO: add better error handling to the mock logic.
-        assert!(index < self.len);
+    pub fn ref_at(&self, index: usize) -> Result<Ref<'a, M, T>, MockError> {
+        if index >= self.len {
+            return Err(MockError::InvalidIndex);
+        }
 
         let addr = self
             .addr
             .checked_add((index * size_of::<T>()) as u64)
             .unwrap();
 
-        Ref::new(self.mem, addr)
+        Ok(Ref::new(self.mem, addr))
     }
 }
 
@@ -162,13 +192,17 @@ impl<'a, M: GuestMemory> DescriptorTable<'a, M> {
     }
 
     /// Read one descriptor from the specified index.
-    pub fn load(&self, index: u16) -> Descriptor {
-        self.table.ref_at(index as usize).load()
+    pub fn load(&self, index: u16) -> Result<Descriptor, MockError> {
+        self.table
+            .ref_at(index as usize)
+            .map(|load_ref| load_ref.load())
     }
 
     /// Write one descriptor at the specified index.
-    pub fn store(&self, index: u16, value: Descriptor) {
-        self.table.ref_at(index as usize).store(value)
+    pub fn store(&self, index: u16, value: Descriptor) -> Result<(), MockError> {
+        self.table
+            .ref_at(index as usize)
+            .map(|store_ref| store_ref.store(value))
     }
 
     /// Return the total size of the DescriptorTable in bytes.
@@ -177,7 +211,7 @@ impl<'a, M: GuestMemory> DescriptorTable<'a, M> {
     }
 
     /// Create a chain of descriptors.
-    pub fn build_chain(&mut self, len: u16) -> u16 {
+    pub fn build_chain(&mut self, len: u16) -> Result<u16, MockError> {
         let indices = self
             .free_descriptors
             .iter()
@@ -199,10 +233,10 @@ impl<'a, M: GuestMemory> DescriptorTable<'a, M> {
             } else {
                 desc.set_flags(0);
             }
-            self.store(index_value, desc);
+            self.store(index_value, desc)?;
         }
 
-        indices[0]
+        Ok(indices[0])
     }
 }
 
@@ -308,13 +342,14 @@ impl<'a, M: GuestMemory> MockSplitQueue<'a, M> {
         self.used_addr
     }
 
-    fn update_avail_idx(&mut self, value: u16) {
+    fn update_avail_idx(&mut self, value: u16) -> Result<(), MockError> {
         let avail_idx = self.avail.idx.load();
-        self.avail.ring.ref_at(avail_idx as usize).store(value);
+        self.avail.ring.ref_at(avail_idx as usize)?.store(value);
         self.avail.idx.store(avail_idx.wrapping_add(1));
+        Ok(())
     }
 
-    fn alloc_indirect_chain(&mut self, len: u16) -> GuestAddress {
+    fn alloc_indirect_chain(&mut self, len: u16) -> Result<GuestAddress, MockError> {
         // To simplify things for now, we round up the table len as a multiple of 16. When this is
         // no longer the case, we should make sure the starting address of the descriptor table
         // we're  creating below is properly aligned.
@@ -326,36 +361,37 @@ impl<'a, M: GuestMemory> MockSplitQueue<'a, M> {
         };
 
         let mut table = DescriptorTable::new(self.mem, self.indirect_addr, table_len);
-        let head_decriptor_index = table.build_chain(len);
+        let head_decriptor_index = table.build_chain(len)?;
         // When building indirect descriptor tables, the descriptor at index 0 is supposed to be
         // first in the resulting chain. Just making sure our logic actually makes that happen.
         assert_eq!(head_decriptor_index, 0);
 
         let table_addr = self.indirect_addr;
         self.indirect_addr = self.indirect_addr.checked_add(table.total_size()).unwrap();
-        table_addr
+        Ok(table_addr)
     }
 
     /// Add a descriptor chain to the table.
-    pub fn add_chain(&mut self, len: u16) {
-        let head_idx = self.desc_table.build_chain(len);
-        self.update_avail_idx(head_idx);
+    pub fn add_chain(&mut self, len: u16) -> Result<(), MockError> {
+        self.desc_table
+            .build_chain(len)
+            .and_then(|head_idx| self.update_avail_idx(head_idx))
     }
 
     /// Add an indirect descriptor chain to the table.
-    pub fn add_indirect_chain(&mut self, len: u16) {
-        let head_idx = self.desc_table.build_chain(1);
+    pub fn add_indirect_chain(&mut self, len: u16) -> Result<(), MockError> {
+        let head_idx = self.desc_table.build_chain(1)?;
 
         // We just allocate the indirect table and forget about it for now.
-        let indirect_addr = self.alloc_indirect_chain(len);
+        let indirect_addr = self.alloc_indirect_chain(len)?;
 
-        let mut desc = self.desc_table.load(head_idx);
+        let mut desc = self.desc_table.load(head_idx)?;
         desc.set_flags(VIRTQ_DESC_F_INDIRECT);
         desc.set_addr(indirect_addr.raw_value());
         desc.set_len(u32::from(len) * size_of::<Descriptor>() as u32);
 
-        self.desc_table.store(head_idx, desc);
-        self.update_avail_idx(head_idx);
+        self.desc_table.store(head_idx, desc)?;
+        self.update_avail_idx(head_idx)
     }
 
     /// Creates a new `Queue`, using the underlying memory regions represented
@@ -373,14 +409,13 @@ impl<'a, M: GuestMemory> MockSplitQueue<'a, M> {
 
     /// Writes a single descriptor chain to the memory object of the queue, at the beginning of the
     /// descriptor table, and returns the associated `DescriptorChain` object.
-    pub fn build_desc_chain(&self, descs: &[Descriptor]) -> DescriptorChain<&M> {
-        self.add_desc_chain(descs, 0);
-
+    pub fn build_desc_chain(&self, descs: &[Descriptor]) -> Result<DescriptorChain<&M>, MockError> {
+        self.add_desc_chain(descs, 0)?;
         self.create_queue(self.mem)
             .iter()
             .unwrap()
             .next()
-            .expect("failed to build desc chain")
+            .ok_or(MockError::InvalidNextAvail)
     }
 
     /// Adds a descriptor chain to the memory object of the queue.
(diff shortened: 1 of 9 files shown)


crates/virtio-queue/src/mock.rs
```diff
@@ -371,16 +397,60 @@ impl<'a, M: GuestMemory> MockSplitQueue<'a, M> {
         q
     }
 
-    /// Writes a single descriptor chain to the memory object of the queue, at the beginning of the
+    /// Writes descriptors to the memory object of the queue, at the beginning of the
     /// descriptor table, and returns the associated `DescriptorChain` object.
-    pub fn build_desc_chain(&self, descs: &[Descriptor]) -> DescriptorChain<&M> {
-        self.add_desc_chain(descs, 0);
+    pub fn build_chain_of_descriptors(
```
andreeaflorescu, line 402: This function looks like a generalization of `build_desc_chain`. We should think about a way to just keep one of them.

crates/virtio-queue/src/mock.rs
```diff
@@ -371,16 +397,60 @@ impl<'a, M: GuestMemory> MockSplitQueue<'a, M> {
...
+    }
+
+    /// Writes a single descriptor chain to the memory object of the queue, at the beginning of the
+    /// descriptor table, and returns the associated `DescriptorChain` object.
+    pub fn build_desc_chain(&self, descs: &[Descriptor]) -> Result<DescriptorChain<&M>, Error> {
+        if self.add_desc_chain(descs, 0).is_ok() {
+            Ok(self
+                .create_queue(self.mem)
+                .iter()
+                .unwrap()
```
andreeaflorescu, line 448: This function now returns a result, but we're still using `unwrap` and `expect`. Is there any reason for that?

I would also recommend trying to rewrite this using Rust combinators because it makes the code easier to read. Here are some examples:
- https://doc.rust-lang.org/rust-by-example/error/option_unwrap/question_mark.html
- https://doc.rust-lang.org/rust-by-example/error/option_unwrap/map.html
- https://doc.rust-lang.org/rust-by-example/error/option_unwrap/and_then.html

crates/virtio-queue/src/mock.rs
```diff
@@ -407,7 +477,9 @@ impl<'a, M: GuestMemory> MockSplitQueue<'a, M> {
             };
 
             let desc = Descriptor::new(addr, len, flags, next);
-            self.desc_table().store(i, desc);
+            if let Err(e) = self.desc_table().store(i, desc) {
```
andreeaflorescu, line 480: This can be written as: ```self.desc_table().store(i, desc)?;```. Same for all the other similar code patterns in this PR.

andreeaflorescu (changes requested):
We should separate this in 2 commits:
- one for the error handling; we should also explain in the commit message why this is needed (i.e. otherwise during fuzzing we get panics that are not related to the code)
- one for adding the new function, and also explain why this is needed

crates/virtio-queue/src/mock.rs
DeliaPavel: I believe the unwrap should exist because the`iter` function returns a `Result<AvailIter<'_, M::T>, Error>` and the `next` function should be called on an `AvailIter` object. Also, the next function returns `Option<Self::Item>`, where `Item` is an alias for `DescriptorChain<M>`, which the function returns. Maybe this could be replaced with something else, but I don't know with what exactly.

crates/virtio-queue/src/mock.rs
DeliaPavel: Done. I only kept build_desc_chain and added a bool parameter that, when true, only one descriptor chain is built and, when false, multiple descriptor chains are built.

crates/virtio-queue/src/mock.rs
DeliaPavel: Done

crates/virtio-queue/src/mock.rs
```diff
@@ -371,43 +380,64 @@ impl<'a, M: GuestMemory> MockSplitQueue<'a, M> {
...
     }
 
-    /// Writes a single descriptor chain to the memory object of the queue, at the beginning of the
+    /// Writes one or more descriptor chains to the memory object of the queue, depending on the
+    /// value of the bool variable one_chain, at the beginning of the
     /// descriptor table, and returns the associated `DescriptorChain` object.
-    pub fn build_desc_chain(&self, descs: &[Descriptor]) -> DescriptorChain<&M> {
-        self.add_desc_chain(descs, 0);
-
+    pub fn build_desc_chain(
```
andreeaflorescu, line 386: Having bool as a parameter that changes the behavior of the function is an antipattern. There are some more details here about this: https://softwareengineering.stackexchange.com/questions/147977/is-it-wrong-to-use-a-boolean-parameter-to-determine-behavior.

Instead, we should have 2 functions, one that builds just one descriptor, and one that builds multiple as specified in the input. Now this would mean going back to the initial solution you proposed. The problem there was the code duplication. One way of solving this is to have a private helper function that takes care of the common code.

andreeaflorescu, line 386: Having bool as a parameter that changes the behavior of the function is an antipattern. There are some more details here about this: https://softwareengineering.stackexchange.com/questions/147977/is-it-wrong-to-use-a-boolean-parameter-to-determine-behavior.

Instead, we should have 2 functions, one that builds just one descriptor, and one that builds multiple as specified in the input. Now this would mean going back to the initial solution you proposed. The problem there was the code duplication. One way of solving this is to have a private helper function that takes care of the common code.

crates/virtio-queue/src/mock.rs
```diff
@@ -64,16 +64,18 @@ impl<'a, M: GuestMemory, T: ByteValued> ArrayRef<'a, M, T> {
 
     /// Return a `Ref` object pointing to an address defined by a particular
     /// index offset in the region.
-    pub fn ref_at(&self, index: usize) -> Ref<'a, M, T> {
+    pub fn ref_at(&self, index: usize) -> Result<Ref<'a, M, T>, Error> {
         // TODO: add better error handling to the mock logic.
-        assert!(index < self.len);
+        if index >= self.len {
+            return Err(Error::InvalidDescriptorIndex);
```
lauralt, line 70: We don't know about `Descriptor`s at the `ArrayRef` level, i.e. it can be used for any ByteValued type. Can we rename the error variant to `InvalidIndex` or maybe something more accurate?

crates/virtio-queue/src/mock.rs
```diff
@@ -308,13 +314,15 @@ impl<'a, M: GuestMemory> MockSplitQueue<'a, M> {
         self.used_addr
     }
 
-    fn update_avail_idx(&mut self, value: u16) {
+    fn update_avail_idx(&mut self, value: u16) -> Result<(), Error> {
         let avail_idx = self.avail.idx.load();
-        self.avail.ring.ref_at(avail_idx as usize).store(value);
+        let avail_ref = self.avail.ring.ref_at(avail_idx as usize)?;
+        avail_ref.store(value);
```
lauralt, line 320: nit: you don't need to split this into two, you can also use:
`self.avail.ring.ref_at(avail_idx as usize)?.store(value);`

crates/virtio-queue/src/mock.rs
```diff
@@ -373,14 +382,13 @@ impl<'a, M: GuestMemory> MockSplitQueue<'a, M> {
...
-        self.add_desc_chain(descs, 0);
-
+    pub fn build_desc_chain(&self, descs: &[Descriptor]) -> Result<DescriptorChain<&M>, Error> {
+        self.add_desc_chain(descs, 0)?;
         self.create_queue(self.mem)
             .iter()
             .unwrap()
             .next()
-            .expect("failed to build desc chain")
+            .ok_or(Error::InvalidDescriptorIndex)
```
lauralt, line 396: Does `next` fail only in case of an invalid index? Also I think we are iterating over descriptor chains, so I'm not sure if `Descriptor` is correct there.

lauralt (commented):
First part of the review.

crates/virtio-queue/src/mock.rs
DeliaPavel: I added an InvalidNextAvail Error because I saw that the next function fails because of next_avail

crates/virtio-queue/src/mock.rs
DeliaPavel: Done

crates/virtio-queue/src/mock.rs
```diff
@@ -390,30 +393,37 @@ impl<'a, M: GuestMemory> MockSplitQueue<'a, M> {
...
+    /// Adds descriptor chains to the memory object of the queue.
+    // `descs` represents a slice of `Descriptor` objects which are used to populate the chains, and
+    // `offset` is the index in the descriptor table where the chains should be added.
+    // The descriptor chain related information is written in memory starting with address 0.
+    // The `addr` fields of the input descriptors should start at a sufficiently
+    // greater location (i.e. 1MiB, or `0x10_0000`).
+    pub fn add_desc_chain(&self, descs: &[Descriptor], offset: u16) -> Result<(), Error> {
+        for (idx, desc) in descs.iter().enumerate() {
+            let i = idx as u16 + offset;
+            let desc = Descriptor::new(desc.addr().0, desc.len(), desc.flags(), desc.next());
```
lauralt, line 426: I don't think we need to create again the descriptor if it has the same fields as the initial one.

crates/virtio-queue/src/mock.rs
```diff
@@ -390,30 +393,37 @@ impl<'a, M: GuestMemory> MockSplitQueue<'a, M> {
...
+    }
 
-            let desc = Descriptor::new(addr, len, flags, next);
+    /// Adds descriptor chains to the memory object of the queue.
+    // `descs` represents a slice of `Descriptor` objects which are used to populate the chains, and
+    // `offset` is the index in the descriptor table where the chains should be added.
+    // The descriptor chain related information is written in memory starting with address 0.
+    // The `addr` fields of the input descriptors should start at a sufficiently
+    // greater location (i.e. 1MiB, or `0x10_0000`).
+    pub fn add_desc_chain(&self, descs: &[Descriptor], offset: u16) -> Result<(), Error> {
```
lauralt, line 423: The logic here is a bit broken. So from what I understand, this can be used to add more descriptor chains, and when the next flag is not set, I assume that will be the head of a new descriptor chain. In that case, we should add every head index into the available ring, not only the first index, and increment the avail idx with the number of new entries from the ring. Also, the function should be renamed to `add_desc_chains`.

crates/virtio-queue/src/mock.rs
```diff
@@ -379,9 +379,12 @@ impl<'a, M: GuestMemory> MockSplitQueue<'a, M> {
         q
     }
 
-    /// Writes a single descriptor chain to the memory object of the queue, at the beginning of the
-    /// descriptor table, and returns the associated `DescriptorChain` object.
-    pub fn build_desc_chain(&self, descs: &[Descriptor]) -> Result<DescriptorChain<&M>, Error> {
+    /// Writes a multiple descriptor chains to the memory object of the queue, at the beginning of
+    /// the descriptor table, and returns the associated `DescriptorChain` object.
```
lauralt, line 383: I think it should be `returns the first DescriptorChain available` or something like that.

crates/virtio-queue/src/lib.rs
```diff
@@ -57,6 +57,10 @@ pub enum Error {
     InvalidChain,
     /// Invalid descriptor index.
     InvalidDescriptorIndex,
+    /// Invalid Ref index
+    InvalidIndex,
```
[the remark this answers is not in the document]
DeliaPavel: Done

crates/virtio-queue/src/mock.rs
DeliaPavel: Yes, you are right, I deleted it

crates/virtio-queue/src/mock.rs
DeliaPavel: Done

crates/virtio-queue/src/mock.rs
DeliaPavel: I changed the logic following your instructions, if I omitted something or if I misunderstood please tell me

crates/virtio-queue/src/lib.rs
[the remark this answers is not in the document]
lauralt: If you chose to go with the separate enum, we don't need to keep that in `lib.rs`. We can have it in the `mock` module and it will be automatically `cfg`ed (so no need for the `[cfg(any(test, feature = "test-utils"))]`s anymore).

crates/virtio-queue/src/lib.rs
```diff
@@ -74,13 +80,25 @@ impl Display for Error {
...
+}
+
+#[cfg(any(test, feature = "test-utils"))]
+impl Display for MockError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        use self::MockError::*;
+
+        match self {
             InvalidIndex => write!(f, "invalid ref index"),
             InvalidNextAvail => write!(f, "invalid next avail"),
```
lauralt, line 94: Can you use clearer error messages here? `ref` is very vague for example.

DeliaPavel: Done.

crates/virtio-queue/src/mock.rs
DeliaPavel: You are right, I modified it.

crates/virtio-queue/src/lib.rs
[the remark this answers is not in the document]
DeliaPavel: Done.

crates/virtio-queue/src/mock.rs
```diff
@@ -379,62 +407,78 @@ impl<'a, M: GuestMemory> MockSplitQueue<'a, M> {
...
+
+        for (idx, desc) in descs.iter().enumerate() {
+            let i = idx as u16 + offset;
+            self.desc_table().store(i, *desc)?;
+
+            if desc.flags() == VIRTQ_DESC_F_NEXT || idx == 0 {
+                // Update the available ring position.
+                self.mem
+                    .write_obj(
+                        offset + new_entries,
```
lauralt, line 466: I think this should be `offset + idx` (or `i`), the entry from the available ring must be the index in the descriptor table of the chain head.

crates/virtio-queue/src/mock.rs
```diff
@@ -379,62 +407,78 @@ impl<'a, M: GuestMemory> MockSplitQueue<'a, M> {
...
-                    VIRTQ_AVAIL_RING_HEADER_SIZE + avail_idx as u64 * VIRTQ_AVAIL_ELEMENT_SIZE,
-                ),
-            )
-            .unwrap();
+
+        for (idx, desc) in descs.iter().enumerate() {
+            let i = idx as u16 + offset;
+            self.desc_table().store(i, *desc)?;
+
+            if desc.flags() == VIRTQ_DESC_F_NEXT || idx == 0 {
```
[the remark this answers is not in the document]
lauralt: No, you have to check that the flag is **not** set, otherwise yes that's how we should do it.

crates/virtio-queue/src/mock.rs
lauralt: Hmm I think this is now `offset + i`, what I meant is offset + idx = i. The explanation is that you need to write in the available ring the index of the descriptor chain head from the descriptor table, and you are adding these descriptors at indexes: offset, offset + 1, offset + 2 ..., where idx = 0, 1, 2 ...

crates/virtio-queue/src/mock.rs
```diff
@@ -407,62 +407,78 @@ impl<'a, M: GuestMemory> MockSplitQueue<'a, M> {
...
-                    VIRTQ_AVAIL_RING_HEADER_SIZE + avail_idx as u64 * VIRTQ_AVAIL_ELEMENT_SIZE,
-                ),
-            )
-            .unwrap();
+
+        for (idx, desc) in descs.iter().enumerate() {
+            let i = idx as u16 + offset;
+            self.desc_table().store(i, *desc)?;
+
+            if idx == 0 || descs[idx - 1].flags() != VIRTQ_DESC_F_NEXT {
```
lauralt, line 462: This is still not correct :(, you should check that VIRTQ_DESC_F_NEXT bit is not set in flags, not that the flags value is different than that. For example, you can have both VIRTQ_DESC_F_NEXT and VIRTQ_DESC_F_WRITE flags set, and with the current code, you consider that descriptor to be followed by a chain head, which is not actually correct.

crates/virtio-queue/src/mock.rs
DeliaPavel: Done.

crates/virtio-queue/src/mock.rs
DeliaPavel: Thank you, you are right, this fixed the test that was failing.

lauralt (dismissed):
Good job! :+1:

andreeaflorescu (dismissed):

andreeaflorescu (approved):

lauralt (approved):

PR #163, opened 2022-05-04
add missing documentation to the vsock packet

lauralt:
### Summary of the PR

Added the documentation that was missing for publishing the vsock packet.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/crates/devices/virtio-vsock/README.md
+++ b/crates/devices/virtio-vsock/README.md
@@ -1,0 +1,63 @@
+# virtio-vsock
+
+The `virtio-vsock` crate provides abstractions for the components of the
+[vsock device](https://docs.oasis-open.org/virtio/virtio/v1.1/csprd01/virtio-v1.1-csprd01.html#x1-39000010).
+For now, it offers only an implementation for the vsock packet.
+The support is provided only for stream sockets.
+
+The vsock device is a socket device that can be used as a communication
+mechanism between the host and the guest. It is implemented using the virtio
+standard. The vsock device has three queues: an RX one, a TX one and an event
+one. In a simplified manner, the communication between the host and the guest
+is realized with the buffers that are exchanged using the device’s queues.
+These buffers are called packets in the vsock device context.
+
+## Vsock Packet
+
+### Design
+
+The virtio vsock packet is defined in the standard as having a header of type
+`virtio_vsock_hdr` and an optional `data` array of bytes. There are multiple
+operations that can be requested within a packet, e.g. `VIRTIO_VSOCK_OP_RST`
+for resetting the connection, `VIRTIO_VSOCK_OP_RW` for sending payload. Most
+operations are of the `VIRTIO_VSOCK_OP_RW` type, which means for data transfer,
+and the other ones are used for connection and buffer space management.
+`data` is non-empty only for the `VIRTIO_VSOCK_OP_RW` operations.
+
+The abstraction used for the packet implementation is the `VsockPacket`.
+It is using
+[`VolatileSlice`](https://github.com/rust-vmm/vm-memory/blob/fc7153a4f63c352d1fa9419c4654a6c9aec408cb/src/volatile_memory.rs#L266)s
+for representing the header and the data. We chose to use the `VolatileSlice`
+because it's a safe wrapper over the unsafe Rust's raw pointers, and it is also
+generic enough to allow creating packets from pointers to slices. Going with a
+`GuestMemory` based approach would not make such configuration possible.
+More details (including design
+limitations) in [the `packet`'s module-level documentation](src/packet.rs).
+
+A `VsockPacket` instance is created by parsing a descriptor chain from either
+the TX or the RX virtqueue. The `VsockPacket` API is also providing methods for
+creating/setting up packets directly from pointers to slices.
+It also offers setters and getters for each `virtio_vsock_hdr` field (e.g.
+*src_cid*, *dst_port*, *op*).
+
+### Usage
+
+The driver queues receive buffers on the RX virtqueue, and outgoing packets on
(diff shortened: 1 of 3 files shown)


crates/devices/virtio-vsock/README.md
```diff
@@ -0,0 +1,120 @@
...
+`virtio_vsock_hdr` and an optional `data` array of bytes. There are multiple
+operations that can be requested within a packet, e.g. `VIRTIO_VSOCK_OP_RST`
+for resetting the connection, `VIRTIO_VSOCK_OP_RW` for sending payload. Most
+operations are of the `VIRTIO_VSOCK_OP_RW` type, which means for data transfer,
+and the other ones are used for connection and buffer space management.
+`data` is non-empty only for the `VIRTIO_VSOCK_OP_RW` operations.
+
+The abstraction used for the packet implementation is the `VsockPacket`.
+It is using
+[`VolatileSlice`](https://github.com/rust-vmm/vm-memory/blob/fc7153a4f63c352d1fa9419c4654a6c9aec408cb/src/volatile_memory.rs#L266)s
```
andreeaflorescu, line 29: Can you also add a comment here about why we choose to go with VolatileSlice?

crates/devices/virtio-vsock/README.md
```diff
@@ -0,0 +1,120 @@
...
+for representing the header and the data.
+More details (including design
+limitations) in [the `packet`'s module-level documentation](src/packet.rs).
+
+A `VsockPacket` instance is created by parsing a descriptor chain from either
+the TX or the RX virtqueue. The `VsockPacket` API is also providing methods for
+creating/setting up packets directly from pointers to slices.
+It also offers setters and getters for each `virtio_vsock_hdr` field (e.g.
```
andreeaflorescu, line 37: Can you also add some details about how the VsockPacket header operations work? Specifically, i think it would be important to note that the vsock packet header is read once to avoid toctou errors. This is a design choice that's important and could be easily missed.

crates/devices/virtio-vsock/README.md
```diff
@@ -0,0 +1,120 @@
...
+On the TX side, the device processes the TX queue using
+`VsockPacket::from_tx_virtq_chain`, packages the read buffers as vsock packets,
+and then sends them to the backend.
+
+### Examples
+
+Below is an example of how the device would consume RX chains and populate the
+buffers with data.
+
+```rust
```
andreeaflorescu, line 55: Since we are not running tests for these examples they have the tendency to rot and easily become outdated as we push changes. Instead, we could create an `example` folder and move it there. And in this section of the readme just point to the example module. Here is an example layout: https://doc.rust-lang.org/cargo/guide/project-layout.html

crates/devices/virtio-vsock/src/packet.rs
```diff
@@ -398,6 +476,78 @@ impl<'a, B: BitmapSlice> VsockPacket<'a, B> {
...
+    /// # const TYPE_STREAM: u16 = 1;
+    /// # const OP_RW: u16 = 5;
+    /// # const FLAGS: u32 = 7;
+    /// # const FLAG: u32 = 8;
+    /// # const BUF_ALLOC: u32 = 256;
+    /// # const FWD_CNT: u32 = 9;
+    ///
+    /// let mem = GuestMemoryMmap::<()>::from_ranges(&[(GuestAddress(0), 0x10000)]).unwrap();
+    /// let mut queue = MockSplitQueue::new(&mem, 16);
+    /// let mut q = queue.create_queue(&mem);
```
andreeaflorescu, line 512: Should we also hide these lines? They are creating the queue using the mock framework. Maybe this can cause some confusion about how it's supposed to be used.

crates/devices/virtio-vsock/README.md
andreeaflorescu: On a second thought, I am not sure we need this example here at all since we already have something similar in the `.rs` files.

crates/devices/virtio-vsock/README.md
lauralt: I added more details about this in the `lib.rs` docs, and didn't want to repeat those.

crates/devices/virtio-vsock/README.md
lauralt: I'm fine with removing it from the README as long as it's not a requirement to have the `Examples` section in this file.

andreeaflorescu: It shouldn't be a requirement to have examples here. I think the guidelines are about having examples in general.

crates/devices/virtio-vsock/README.md
lauralt: Done.

crates/devices/virtio-vsock/README.md
lauralt: I removed the example, and just linked the ones from the packet module.

crates/devices/virtio-vsock/README.md
lauralt: Sorry, the documentation was from `packet.rs`, not `lib.rs`: https://github.com/rust-vmm/vm-virtio/blob/main/crates/devices/virtio-vsock/src/packet.rs#L22. Is there anything else we should add in any of these 2 docs?

crates/devices/virtio-vsock/src/packet.rs
lauralt: I didn't remove them so that it's clear what is `q`, but they are indeed confusing. I changed to using the same approach from the `virtio-queue` examples, and have a separate function for the queue creation. I'm a bit disappointed I forgot about it until now :(.

PR #167, opened 2022-05-17
Add initial virtio console implementation

RaduNiculae:
Handles the queue chain processing for the virtio console device.

Signed-off-by: Niculae Radu <r.niculae99@gmail.com>

### Summary of the PR

This is part of the virtio console implementation.
The purpose of this component is to process the queue chains and
to implement additional features (for example resize). The features
are not yet implemented. 

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/crates/devices/virtio-console/Cargo.toml
+++ b/crates/devices/virtio-console/Cargo.toml
@@ -1,9 +1,10 @@
 [package]
 name = "virtio-console"
 version = "0.1.0"
-authors = ["rust-vmm community", "rust-vmm AWS maintainers <rust-vmm-maintainers@amazon.com>"]
+authors = ["rust-vmm community", "rust-vmm AWS maintainers <rust-vmm-maintainers@amazon.com>",
+    "Niculae Radu-Alexandru <r.niculae99@gmail.com>"]
 description = "virtio console device implementation"
-repository = "https://github.com/RaduNiculae/vm-virtio"
+repository = "https://github.com/rust-vmm/vm-virtio"
 keywords = ["virtio", "console"]
 readme = "README.md"
 license = "Apache-2.0 OR BSD-3-Clause"
@@ -11,5 +12,5 @@ edition = "2021"
 
 
 [dependencies]
-virtio-queue = { path = "../../virtio-queue" }
+virtio-queue = { path = "../../virtio-queue", version = ">=0.3.0" }
 vm-memory = "0.8.0"
--- a/crates/devices/virtio-console/Cargo.toml
+++ b/crates/devices/virtio-console/Cargo.toml
@@ -1,0 +1,15 @@
+[package]
+name = "virtio-console"
+version = "0.1.0"
+authors = ["rust-vmm community", "rust-vmm AWS maintainers <rust-vmm-maintainers@amazon.com>"]
+description = "virtio console device implementation"
+repository = "https://github.com/RaduNiculae/vm-virtio"
+keywords = ["virtio", "console"]
+readme = "README.md"
+license = "Apache-2.0 OR BSD-3-Clause"
+edition = "2021"
+
+
+[dependencies]
+virtio-queue = { path = "../../virtio-queue" }
+vm-memory = "0.8.0"
--- a/crates/devices/virtio-console/src/console.rs
+++ b/crates/devices/virtio-console/src/console.rs
@@ -1,0 +1,105 @@
+use std::cmp;
+use std::io::{Write};
+use std::ops::Deref;
+use std::sync::{Arc, Mutex};
+use virtio_queue::{DescriptorChain};
+use vm_memory::{Bytes, GuestMemory};
+
+/// Console device resize feature.
+pub const VIRTIO_CONSOLE_F_SIZE: u64 = 0;
+
+/// Console device multiport feature.
+pub const VIRTIO_CONSOLE_F_MULTIPORT: u64 = 1;
+
+/// Console device emergency write feature.
+pub const VIRTIO_CONSOLE_F_EMERG_WRITE: u64 = 2;
+
+
+/// Console struct
+pub struct Console {
+    /// Data to send to the driver via receiveq
+    input_buffer: Arc<Mutex<Vec<u8>>>,
+
+    // /// Driver acked features
+    // features: u64,
+}
+
+impl Console {
+    /// Create new console object
+    pub fn new() -> Self{
+        Console {
+            input_buffer: Arc::new(Mutex::new(Vec::new())),
+            // features,
+        }
+    }
+
+    // fn has_feature(&self, feature_pos: u64) -> bool {
+    //     (self.features & (1u64 << feature_pos)) != 0
+    // }
+
+    /// Add data to the input buffer
+    pub fn enqueue_data(&self, data: &mut Vec<u8>) {
+        let mut input_buffer = self.input_buffer.lock().unwrap();
+        input_buffer.append(data);
+    }
+
+    /// Check if input buffer is empty
+    pub fn is_input_buffer_empty(&self) -> bool {
+        let input_buffer = self.input_buffer.lock().unwrap();
+        input_buffer.is_empty()
+    }
+
+    /// Process transmitq chain
+    pub fn process_transmitq_chain<M, T>(&self, desc_chain: &mut DescriptorChain<M>, output: &mut T) -> usize
+        where
+            M: Deref,
+            M::Target: GuestMemory,
+            T: Write,
+    {
+        let mut len = 0;
+        while let Some(desc) = desc_chain.next() {
+            len = len + desc_chain.memory().write_to(
+                desc.addr(),
+                output,
+                desc.len() as usize,
+            ).unwrap();
+            let _ = output.flush();
+        }
+
+        len
+    }
+
+    /// Process receiveq chain
+    pub fn process_receiveq_chain<M>(&self, desc_chain: &mut DescriptorChain<M>) -> usize
+        where
+            M: Deref,
+            M::Target: GuestMemory,
+    {
+        let mut sent_bytes: usize = 0;
+        let mut input_buffer = self.input_buffer.lock().unwrap();
+
+        if input_buffer.is_empty() {
+            return 0;
+        }
+
+        while let Some(desc) = desc_chain.next() {
+            let len_write = cmp::min(desc.len() as usize, input_buffer.len() as usize);
+            let buffer_slice = input_buffer.drain(..len_write).collect::<Vec<u8>>();
+
+            if let Err(err) = desc_chain.memory().write_slice(
+                &buffer_slice[..],
+                desc.addr()
+            ) {
+                println!("Failed to write slice: {:?}", err);
+                break;
+            }
+            sent_bytes += len_write;
+
+            if input_buffer.is_empty() {
+                break;
+            }
+        }
+
+        sent_bytes
+    }
+}
--- a/crates/devices/virtio-console/src/console.rs
+++ b/crates/devices/virtio-console/src/console.rs
@@ -1,98 +1,169 @@
+// Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
 use std::cmp;
-use std::io::{Write};
+use std::fmt::{self, Display};
+use std::io::Write;
 use std::ops::Deref;
 use std::sync::{Arc, Mutex};
-use virtio_queue::{DescriptorChain};
-use vm_memory::{Bytes, GuestMemory};
+use virtio_queue::DescriptorChain;
+use vm_memory::{Bytes, GuestMemory, GuestMemoryError};
 
-/// Console device resize feature.
-pub const VIRTIO_CONSOLE_F_SIZE: u64 = 0;
+/// Virtio console chain processing abstraction.
+///
+/// This module offers an abstraction for working with the descriptor chains of the receive and
+/// transmit queues.
+///  * The transmit queue is used to send data from the driver to the device and it's buffers are
+///    read-only for the device.
+///  * The receive queue is used to send data from the device to the driver and it's buffers are
+///    write-only for the device.
+///
+/// To implement this abstraction the [`Console`](struct.Console.html)  structure is provided which
+/// consists of the `input_buffer` member and exposes several methods.
+/// * The [`input_buffer`](struct.Console.html#field.input_buffer) is a buffer used to store data
+///   before sending it to the driver. The following methods are provided to interact with it:
+///   - [`enqueue_data`](struct.Console.html#method.enqueue_data) which is used to add data to the
+///     end of the buffer.
+///   - [`is_input_buffer_empty`](struct.Console.html#method.is_input_buffer_empty) which is used
+///     assess if there is any data waiting to be sent.
+///
+/// The methods for processing a descriptor chain are the following:
+/// *  [`process_transmitq_chain`](struct.Console.html#method.process_transmitq_chain) which is
+///    used to read the data from a descriptor chain of the transmit queue.
+/// * [`process_receiveq_chain`](struct.Console.html#method.process_receiveq_chain) which is used
+///   to send the data to the driver by writing it into the buffers of a receive queue descriptor
+///   chain.
 
-/// Console device multiport feature.
-pub const VIRTIO_CONSOLE_F_MULTIPORT: u64 = 1;
+/// Console device errors.
+#[derive(Debug)]
+pub enum Error {
+    /// Failed to write slice to guest memory.
+    WriteToGuestFailed(GuestMemoryError),
+}
 
-/// Console device emergency write feature.
-pub const VIRTIO_CONSOLE_F_EMERG_WRITE: u64 = 2;
+impl Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        use self::Error::*;
 
+        match self {
+            WriteToGuestFailed(ref err) => {
+                write!(f, "console failed to write slice to guest memory: {}", err)
+            }
+        }
+    }
+}
 
-/// Console struct
+/// Console struct that implements the abstraction of virtio console descriptor chain handling.
 pub struct Console {
-    /// Data to send to the driver via receiveq
+    /// Buffer that stores data to be sent to the driver.
     input_buffer: Arc<Mutex<Vec<u8>>>,
-
-    // /// Driver acked features
-    // features: u64,
 }
 
 impl Console {
-    /// Create new console object
-    pub fn new() -> Self{
+    /// Create new console object.
+    pub fn new() -> Self {
         Console {
             input_buffer: Arc::new(Mutex::new(Vec::new())),
-            // features,
         }
     }
 
-    // fn has_feature(&self, feature_pos: u64) -> bool {
-    //     (self.features & (1u64 << feature_pos)) != 0
-    // }
-
-    /// Add data to the input buffer
+    /// Adds data in the form of raw bytes to the input buffer.
+    ///
+    /// # Arguments
+    ///
+    /// * `data` - A mutable reference to a `Vec<u8>` containing the raw bytes that will be
+    ///            sent to the driver when the `process_receiveq_chian` method is called.
     pub fn enqueue_data(&self, data: &mut Vec<u8>) {
         let mut input_buffer = self.input_buffer.lock().unwrap();
         input_buffer.append(data);
     }
 
-    /// Check if input buffer is empty
+    /// Checks if input buffer is empty.
     pub fn is_input_buffer_empty(&self) -> bool {
         let input_buffer = self.input_buffer.lock().unwrap();
         input_buffer.is_empty()
     }
 
-    /// Process transmitq chain
-    pub fn process_transmitq_chain<M, T>(&self, desc_chain: &mut DescriptorChain<M>, output: &mut T) -> usize
-        where
-            M: Deref,
-            M::Target: GuestMemory,
-            T: Write,
+    /// Read data from a `desc_chain` of the transmit queue and write it to a specified output.
+    ///
+    /// # Arguments
+    /// * `desc_chain` - A mutable reference to the descriptor chain that should point to the
+    ///                  buffers that the virtio console driver has sent to the device.
+    /// * `output` - An object that implements the `Write` trait and specifies where the data
+    ///              should be written to (usually `stdout`). The `output` object is not included
+    ///              as part of the [`Console`](struct.Console.html) struct and rather passed as an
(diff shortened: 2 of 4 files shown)


crates/devices/virtio-console/src/console.rs
```diff
@@ -0,0 +1,105 @@
...
+    {
+        let mut sent_bytes: usize = 0;
+        let mut input_buffer = self.input_buffer.lock().unwrap();
+
+        if input_buffer.is_empty() {
+            return 0;
+        }
+
+        while let Some(desc) = desc_chain.next() {
+            let len_write = cmp::min(desc.len() as usize, input_buffer.len() as usize);
```
lauralt, line 86: `input_buffer.len()` is already an usize, no need for an explicit conversion.

crates/devices/virtio-console/src/console.rs
```diff
@@ -0,0 +1,105 @@
...
+            let buffer_slice = input_buffer.drain(..len_write).collect::<Vec<u8>>();
+
+            if let Err(err) = desc_chain.memory().write_slice(
+                &buffer_slice[..],
```
lauralt, line 90: I think the [..] can be omitted.

crates/devices/virtio-console/src/console.rs
```diff
@@ -0,0 +1,105 @@
+use std::cmp;
+use std::io::{Write};
+use std::ops::Deref;
+use std::sync::{Arc, Mutex};
+use virtio_queue::{DescriptorChain};
+use vm_memory::{Bytes, GuestMemory};
+
+/// Console device resize feature.
+pub const VIRTIO_CONSOLE_F_SIZE: u64 = 0;
```
lauralt, line 9: Let's define these only when we are also implementing them.

crates/devices/virtio-console/src/console.rs
```diff
@@ -0,0 +1,105 @@
...
+                output,
+                desc.len() as usize,
+            ).unwrap();
+            let _ = output.flush();
+        }
+
+        len
+    }
+
+    /// Process receiveq chain
```
andreeaflorescu, line 72: Can you add a more descriptive text here? For example, you can write about this function processing the receive queue chain passed as a parameter, and what it returns in case the process is successful.

crates/devices/virtio-console/src/console.rs
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    /// Check if input buffer is empty
+    pub fn is_input_buffer_empty(&self) -> bool {
+        let input_buffer = self.input_buffer.lock().unwrap();
+        input_buffer.is_empty()
+    }
+
+    /// Process transmitq chain
+    pub fn process_transmitq_chain<M, T>(&self, desc_chain: &mut DescriptorChain<M>, output: &mut T) -> usize
```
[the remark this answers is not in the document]
RaduNiculae: I thought it would be more flexible to have the `output` sink given as a parameter instead of being part of the `Console`. This way we can change it on the fly, without having to modify or reinitialize the `Console`. I am not sure how useful this is, I could make it part of the `Console` to have it more unified.

crates/devices/virtio-console/src/console.rs
```diff
@@ -0,0 +1,105 @@
...
+        where
+            M: Deref,
+            M::Target: GuestMemory,
+            T: Write,
+    {
+        let mut len = 0;
+        while let Some(desc) = desc_chain.next() {
+            len = len + desc_chain.memory().write_to(
```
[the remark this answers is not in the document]
RaduNiculae: Done. Thank you.

crates/devices/virtio-console/src/console.rs
RaduNiculae: Done.

crates/devices/virtio-console/src/console.rs
RaduNiculae: Done.

crates/devices/virtio-console/src/console.rs
RaduNiculae: Done.

crates/devices/virtio-console/src/console.rs
```diff
@@ -0,0 +1,105 @@
...
+            let buffer_slice = input_buffer.drain(..len_write).collect::<Vec<u8>>();
+
+            if let Err(err) = desc_chain.memory().write_slice(
+                &buffer_slice[..],
+                desc.addr()
+            ) {
+                println!("Failed to write slice: {:?}", err);
+                break;
+            }
+            sent_bytes += len_write;
```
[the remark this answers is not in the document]
RaduNiculae: What do you mean by unchecked? I updated the error handling to return a `Result` now. Could you have a look if it's ok?

crates/devices/virtio-console/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+description = "virtio console device implementation"
+repository = "https://github.com/RaduNiculae/vm-virtio"
+keywords = ["virtio", "console"]
+readme = "README.md"
+license = "Apache-2.0 OR BSD-3-Clause"
+edition = "2021"
+
+
+[dependencies]
+virtio-queue = { path = "../../virtio-queue" }
```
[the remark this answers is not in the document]
RaduNiculae: Done. Thank you.

crates/devices/virtio-console/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "virtio-console"
+version = "0.1.0"
+authors = ["rust-vmm community", "rust-vmm AWS maintainers <rust-vmm-maintainers@amazon.com>"]
```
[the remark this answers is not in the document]
RaduNiculae: Done.

crates/devices/virtio-console/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+description = "virtio console device implementation"
+repository = "https://github.com/RaduNiculae/vm-virtio"
```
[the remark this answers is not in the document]
RaduNiculae: Done.

crates/devices/virtio-console/src/console.rs
```diff
@@ -0,0 +1,105 @@
+use std::cmp;
+use std::io::{Write};
```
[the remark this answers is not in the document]
RaduNiculae: Done.

crates/devices/virtio-console/src/console.rs
```diff
@@ -0,0 +1,105 @@
...
+        while let Some(desc) = desc_chain.next() {
+            let len_write = cmp::min(desc.len() as usize, input_buffer.len() as usize);
+            let buffer_slice = input_buffer.drain(..len_write).collect::<Vec<u8>>();
+
+            if let Err(err) = desc_chain.memory().write_slice(
+                &buffer_slice[..],
+                desc.addr()
+            ) {
+                println!("Failed to write slice: {:?}", err);
+                break;
```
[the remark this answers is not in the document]
RaduNiculae: I updated the method to return a `Result`.

crates/devices/virtio-console/src/console.rs
RaduNiculae: I updated the documentation.

crates/devices/virtio-console/src/console.rs
```diff
@@ -0,0 +1,176 @@
...
+        match self {
+            WriteToGuestFailed(ref err) => {
+                write!(f, "console failed to write slice to guest memory: {}", err)
+            }
+        }
+    }
+}
+
+/// Console struct that implements the abstraction of virtio console descriptor chain handling.
+pub struct Console {
```
andreeaflorescu, line 57: Can you make sure to derive all the useful traits? Examples and explanations are available here: https://rust-lang.github.io/api-guidelines/interoperability.html

crates/devices/virtio-console/src/console.rs
```diff
@@ -0,0 +1,176 @@
...
+    /// Buffer that stores data to be sent to the driver.
+    input_buffer: Arc<Mutex<Vec<u8>>>,
+}
+
+impl Console {
+    /// Create new console object.
+    pub fn new() -> Self {
+        Console {
+            input_buffer: Arc::new(Mutex::new(Vec::new())),
```
andreeaflorescu, line 66: I think input can be forwarded to untrusted data. To make sure we protect against DoS attacks, we should limit the size of the vector. For this to be possible, we should define a capacity as well, and when we enqueue bytes, we should check that there is still space.

What would be interesting to find out is what would be a good capacity. One thing we could do it we can pass the capacity as input to the constructor.

crates/devices/virtio-console/src/console.rs
```diff
@@ -0,0 +1,176 @@
...
+
+    /// Adds data in the form of raw bytes to the input buffer.
+    ///
+    /// # Arguments
+    ///
+    /// * `data` - A mutable reference to a `Vec<u8>` containing the raw bytes that will be
+    ///            sent to the driver when the `process_receiveq_chian` method is called.
+    pub fn enqueue_data(&self, data: &mut Vec<u8>) {
+        let mut input_buffer = self.input_buffer.lock().unwrap();
+        input_buffer.append(data);
```
andreeaflorescu, line 78: We should probably add a check here to not allow unlimited growth of the input buffer.

crates/devices/virtio-console/src/console.rs
```diff
@@ -0,0 +1,176 @@
...
+    ) -> usize
+    where
+        M: Deref,
+        M::Target: GuestMemory,
+        T: Write,
+    {
+        // Iterate through all the buffers in the chain and write the data to output.
+        let mut len = 0;
+        while let Some(desc) = desc_chain.next() {
+            len += desc_chain
```
andreeaflorescu, line 114: The operation here is unchecked in the sense that we do not check whether len + write_len still fits in a `usize`. There are 2 possibilities here:
- either it's impossible for the operation to overflow, in which case we should write a comment about why is that the case, or
- the operation can overflow. In this case we have multiple options as well. We can either use `checked_add` and return an error when the overflow happens (reference for checked operations: https://doc.rust-lang.org/std/primitive.usize.html#method.checked_add), or we can make len a type that wraps around, or we can use [saturating_add](https://doc.rust-lang.org/std/primitive.usize.html#method.saturating_add). I think the best option would be to return an error in case of overflow.

crates/devices/virtio-console/src/console.rs
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    /// Process receiveq chain
+    pub fn process_receiveq_chain<M>(&self, desc_chain: &mut DescriptorChain<M>) -> usize
+        where
+            M: Deref,
+            M::Target: GuestMemory,
+    {
+        let mut sent_bytes: usize = 0;
+        let mut input_buffer = self.input_buffer.lock().unwrap();
```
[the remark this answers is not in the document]
andreeaflorescu: That makes sense. Resolving this comment.

andreeaflorescu (commented):
Can you all tests for the added functionality?

crates/devices/virtio-console/src/console.rs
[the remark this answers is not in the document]
andreeaflorescu: Here is an explanation about why it means that the operation is unchecked: https://github.com/rust-vmm/vm-virtio/pull/167#discussion_r883380113

crates/devices/virtio-console/src/console.rs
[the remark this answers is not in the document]
andreeaflorescu: Why do we need to change the output on the fly?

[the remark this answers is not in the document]
RaduNiculae: For example, I was thinking you could redirect it to a file in case of a large transfer and when that is done you could go back to `stdout`.

crates/devices/virtio-console/src/console.rs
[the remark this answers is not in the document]
RaduNiculae: Thank you. I updated.

crates/devices/virtio-console/src/console.rs
```diff
@@ -0,0 +1,176 @@
...
+///     end of the buffer.
+///   - [`is_input_buffer_empty`](struct.Console.html#method.is_input_buffer_empty) which is used
+///     assess if there is any data waiting to be sent.
+///
+/// The methods for processing a descriptor chain are the following:
+/// *  [`process_transmitq_chain`](struct.Console.html#method.process_transmitq_chain) which is
+///    used to read the data from a descriptor chain of the transmit queue.
+/// * [`process_receiveq_chain`](struct.Console.html#method.process_receiveq_chain) which is used
+///   to send the data to the driver by writing it into the buffers of a receive queue descriptor
+///   chain.
```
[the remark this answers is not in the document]
RaduNiculae: Done. Thank you.

crates/devices/virtio-console/src/console.rs
```diff
@@ -0,0 +1,176 @@
...
+use std::cmp;
+use std::fmt::{self, Display};
+use std::io::Write;
+use std::ops::Deref;
+use std::sync::{Arc, Mutex};
+use virtio_queue::DescriptorChain;
+use vm_memory::{Bytes, GuestMemory, GuestMemoryError};
+
+/// Virtio console chain processing abstraction.
+///
```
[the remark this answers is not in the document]
RaduNiculae: Done.

crates/devices/virtio-console/src/console.rs
RaduNiculae: I added `Default`, `Clone` and `Debug`. Is there a need for anything else?

crates/devices/virtio-console/src/console.rs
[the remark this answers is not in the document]
andreeaflorescu: This can be a future improvement if we end up needing it. For now I think it's better to also take care of the output similarly to the input.

andreeaflorescu (commented):
Just a few more nits, otherwise LGTM.

lauralt (commented):
Looks good!

crates/devices/virtio-console/src/console.rs
```diff
@@ -0,0 +1,492 @@
...
+        assert_eq!(console.output[..INPUT_SIZE], vec![INPUT_VALUE; INPUT_SIZE]);
+        assert_eq!(
+            console.output[INPUT_SIZE..],
+            vec![INPUT_VALUE * 2; INPUT_SIZE]
+        );
+    }
+
+    #[test]
+    fn test_process_receiveq_chain() {
+        let guest_mem = &mut [0u8; INPUT_SIZE];
```
lauralt, line 384: Why do we call this `guest_mem`?

RaduNiculae: I used this buffer to copy slices of `GuestMemory` and compare it with the input data.

crates/devices/virtio-console/src/console.rs
lauralt: I think we can also add another default constructor that uses the default capacity and receives as parameter the output object.

crates/devices/virtio-console/src/console.rs
```diff
@@ -0,0 +1,472 @@
...
+            console
+                .process_transmitq_chain(&mut chain, &mut output)
+                .unwrap(),
+            2 * INPUT_SIZE
+        );
+        assert_eq!(output, vec![INPUT_VALUE; 2 * INPUT_SIZE]);
+    }
+
+    #[test]
+    fn test_process_receiveq_chain() {
```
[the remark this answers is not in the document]
lauralt: Cool, let's track this in a separate issue/PR. Resolving this for now.

crates/devices/virtio-console/src/console.rs
```diff
@@ -0,0 +1,472 @@
...
+                    desc_chain
+                        .memory()
+                        .write_to(desc.addr(), output, desc.len() as usize)
+                        .map_err(Error::WriteToOutputFailed)?,
+                )
+                .ok_or(Error::ChainLengthOverflow)?;
+            let _ = output.flush();
+        }
+
+        Ok(len)
```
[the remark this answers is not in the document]
lauralt: Synced in private, and decided to remove `len`. In case it is useful in the future, we can then update the API.

crates/devices/virtio-console/src/console.rs
```diff
@@ -0,0 +1,516 @@
...
+        );
+
+        assert_eq!(
+            Console::new_with_capacity(MAX_CAPACITY + 1, vec![0; DEFAULT_CAPACITY]).unwrap_err(),
+            Error::InvalidBufferCapacity
+        );
+
+        let console =
+            Console::new_with_capacity(DEFAULT_CAPACITY, vec![0; DEFAULT_CAPACITY]).unwrap();
+        assert!(!console
```
lauralt, line 325: Instead of `assert!(!....is_err())`s you can use `assert!(....is_ok())`.

crates/devices/virtio-console/src/console.rs
```diff
@@ -0,0 +1,516 @@
...
+    WriteToOutputFailed(GuestMemoryError),
+    /// Transmitq descriptor chain length has overflown
+    ChainLengthOverflow,
+    /// Input buffer maximum capacity exceeded
+    BufferCapacityExceeded,
+    /// Read only descriptor that protocol says to write to.
+    UnexpectedReadOnlyDescriptor,
+    /// Write only descriptor that protocol says to read from.
+    UnexpectedWriteOnlyDescriptor,
+    /// The constructor's parameter for maximum capacity is 0
```
lauralt, line 37: I think we are returning this error also when the capacity > max one.

crates/devices/virtio-console/src/console.rs
```diff
@@ -0,0 +1,516 @@
...
+{
+    /// Create new console object with the default `capacity`.
+    ///
+    /// # Arguments
+    /// * `output` - The output sink where all the data received from
+    ///              the transmitq will be written.
+    /// # Returns
+    /// The console object.
+    pub fn new(output: T) -> Self {
+        Self::new_with_capacity(DEFAULT_CAPACITY, output).unwrap()
```
lauralt, line 123: We should explain why it is safe to unwrap here (even when it seems obvious).

crates/devices/virtio-console/src/console.rs
```diff
@@ -0,0 +1,516 @@
...
+    /// Buffer that stores data to be sent to the driver.
+    input_buffer: Arc<Mutex<Vec<u8>>>,
+    /// Capacity of the input buffer.
+    capacity: usize,
+    /// Output sink.
+    output: T,
+}
+
+// Default maximum capacity
+// TODO: Find the optimal CAPACITY and MAX_CAPACITY through testing
```
lauralt, line 91: You can also leave a comment with why you chose 4096 (a link to the linux driver line).

lauralt (approved):
Nice work!

crates/devices/virtio-console/src/console.rs
andreeaflorescu: We should update both the comment and the implementation of display below to correct the error message.

crates/devices/virtio-console/src/console.rs
```diff
@@ -0,0 +1,516 @@
...
+const DEFAULT_CAPACITY: usize = 4096;
+
+impl Default for Console<Stdout> {
+    fn default() -> Self {
+        Self::new_with_capacity(DEFAULT_CAPACITY, stdout()).unwrap()
+    }
+}
+
+/// Used to check if two consoles are of the same size.
+impl<T> PartialEq for Console<T>
```
andreeaflorescu, line 102: We shouldn't implement `PartialEq` and export it in the public interface like this because it doesn't make sense now that we also have the `output` as part of the `Console`. We could have a dummy `PartialEq` implemented for testing purposes, but this should not be part of the public interface. Also, for that one, we should check that the input is also the same. I'm sorry if my initial comment about the Rust API guidelines confused you.

crates/devices/virtio-console/src/console.rs
```diff
@@ -0,0 +1,516 @@
...
+
+    /// Create new console object with specified `capacity`.
+    ///
+    /// # Arguments
+    /// * `capacity` - The maximum capacity of the input buffer.
+    ///                    Passing 0 will return an `Error`.
+    /// * `output` - The output sink where all the data received from
+    ///              the transmitq will be written.
+    /// # Returns
+    /// The console object or an `Error` if the capacity is 0 or higher than `MAX_CAPACITY`.
```
andreeaflorescu, line 134: The users of the interface cannot know what `MAX_CAPACITY` is unless it is exported. We should mark it as public, and we can then use the following so that the value can easily be found while reading the documentation:
```rust
The console object or an `Error` if the capacity is 0 or higher than [`MAX_CAPACITY`].
```

crates/devices/virtio-console/src/console.rs
andreeaflorescu: We can also just use `unwrap` instead of assert for added code readability:

```rust
let console =
            Console::new_with_capacity(DEFAULT_CAPACITY, vec![0; DEFAULT_CAPACITY]).unwrap();
console.enqueue_data(&mut vec![INPUT_VALUE; DEFAULT_CAPACITY]).unwrap();
....
```

crates/devices/virtio-console/src/console.rs
```diff
@@ -0,0 +1,582 @@
...
+    fn test_default_and_partial_eq() {
+        // Default console
+        let console1 = Console::default();
+        assert_eq!(console1.capacity, DEFAULT_CAPACITY);
+
+        let console2 = Console::default();
+        assert_eq!(console2.capacity, DEFAULT_CAPACITY);
+
+        // Same size, no data
+        assert!(console1.eq(&console2));
```
lauralt, line 341: assert_eq! and assert_ne! are using the PartialEq implementation, so it would be better to use the idiomatic way to compare values instead of `assert!`s, you can find more details about this [here](https://doc.rust-lang.org/book/ch11-01-writing-tests.html#testing-equality-with-the-assert_eq-and-assert_ne-macros). But since these tests will probably disappear, it was more of an FYI.

andreeaflorescu (approved):

crates/devices/virtio-console/src/console.rs
```diff
@@ -0,0 +1,512 @@
...
+            return Ok(0);
+        }
+
+        // Break up the data from the "input_buffer" into chunks that fit into the buffers of the
+        // receive queue descriptor chain and write it to guest memory.
+        while let Some(desc) = desc_chain.next() {
+            if !desc.is_write_only() {
+                return Err(Error::UnexpectedReadOnlyDescriptor);
+            }
+            let len_write = cmp::min(desc.len() as usize, input_buffer.len());
```
lauralt, line 257: Since you are going to anyway push again to update the coverage value and to rebase, it would be clearer I think to convert input_buffer.len() here to u32, and have `len_write` as u32.

lauralt (dismissed):

PR #174, opened 2022-06-14
Update documentation of avail_idx & used_idx

andreeaflorescu:
### Summary of the PR

The `avail_idx` and `used_idx` methods panic if the ordering is not the right one. As checking the ordering incurs a performance degradation, the VMM will be responsible for calling these functions only with the right ordering. We're now making this panic explicit by adding it to the public documentation.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/crates/virtio-queue/src/lib.rs
+++ b/crates/virtio-queue/src/lib.rs
@@ -155,11 +155,19 @@ pub trait QueueStateT: for<'a> QueueStateGuard<'a> {
     fn set_event_idx(&mut self, enabled: bool);
 
     /// Read the `idx` field from the available ring.
+    ///
+    /// # Panics
+    ///
+    /// Panics if order is Release or AcqRel.
     fn avail_idx<M>(&self, mem: &M, order: Ordering) -> Result<Wrapping<u16>, Error>
     where
         M: GuestMemory + ?Sized;
 
     /// Read the `idx` field from the used ring.
+    ///
+    /// # Panics
+    ///
+    /// Panics if order is Release or AcqRel.
     fn used_idx<M: GuestMemory>(&self, mem: &M, order: Ordering) -> Result<Wrapping<u16>, Error>;
 
     /// Put a used descriptor head into the used ring.


crates/virtio-queue/src/queue.rs
```diff
@@ -741,4 +741,35 @@ mod tests {
...
+        let vq = MockSplitQueue::new(m, 16);
+
+        let q = vq.create_queue(m);
+        let supported_ordering = [Ordering::SeqCst, Ordering::Acquire, Ordering::Relaxed];
+        for ord in supported_ordering {
+            assert!(q.avail_idx(ord).is_ok());
+            assert!(q.used_idx(ord).is_ok());
+        }
+
+        // Let's store something dummy in `avail` & `used` so that `avail_idx` and `used_idx` can
```
lauralt, line 759: Shouldn't we do this before the calls of the two methods above? Also the comment says we store data in `avail` and `used`, but looks like we are only doing that for `avail`.

gsserge (approved):

crates/virtio-queue/src/queue.rs
andreeaflorescu: Scratch that. I deleted the stores because actually we don't need them. The `Ok(0)` was returned in my local branch because of a bug that I had in the test code. I removed the store calls.

lauralt (approved):

jiangliu:
How abouting using "debug_assert()" instead of runtime checking?

andreeaflorescu:
> How abouting using "debug_assert()" instead of runtime checking?

@jiangliu this will still cause a panic. In case we don't want to fix this by checking the operation, that we can just leave it as is and add a comment to the function specifying that the function panics in case it's called with unsupported ordering. WDYT?

jiangliu:
> > How abouting using "debug_assert()" instead of runtime checking?
> 
> @jiangliu this will still cause a panic. In case we don't want to fix this by checking the operation, than we can just leave it as is and add a comment to the function specifying that the function panics in case it's called with unsupported ordering. WDYT?

I feel it's a type of program error instead of a runtime error, so not sure about the runtime performance impact.

lauralt (approved):

jiangliu (approved):

PR #183, opened 2022-08-17
use add_desc_chains in tests

andreeaflorescu:
### Summary of the PR

We don't need to manually set the available ring in tests because the
newly added interface add_desc_chains does that for us. Update the
tests such that they are using the interface.

### Requirements

Before submitting your PR, please make sure you addressed the following
requirements:

- [x] All commits in this PR are signed (with `git commit -s`), and the commit
  message has max 60 characters for the summary and max 75 characters for each
  description line.
- [x] All added/changed functionality has a corresponding unit/integration
  test.
- [x] Any newly added `unsafe` code is properly documented.

PR #184, opened 2022-08-17
Update vm-memory to 0.9.0

vireshk:
This updates vm-memory to version 0.9.0.

### DIFF
--- a/crates/devices/virtio-blk/Cargo.toml
+++ b/crates/devices/virtio-blk/Cargo.toml
@@ -13,7 +13,7 @@ edition = "2018"
 backend-stdio = []
 
 [dependencies]
-vm-memory = ">=0.8.0"
+vm-memory = ">=0.9.0"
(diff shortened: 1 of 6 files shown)


rbradford (approved):

andreeaflorescu:
@vireshk I update all the requirements in #183. Do you mind if we merge that one instead? We also need to fix the way we are specifying dependencies as well because right now they are still specified as ">=" instead of using caret versions.

vireshk:
> @vireshk I update all the requirements in #183. Do you mind if we merge that one instead? We also need to fix the way we are specifying dependencies as well because right now they are still specified as ">=" instead of using caret versions.

Ahh sure, I missed that you already have a pull request for this. Thanks.

andreeaflorescu:
@vireshk I just pushed the commit that fixes the problem a few minutes ago because I noticed the main branch no longer builds. Sorry!

PR #192, opened 2022-09-12
virtio-bindings: drop bindings for older kernels 

alyssais:
### Summary of the PR

As [pointed out](https://github.com/rust-vmm/virtio-bindings/issues/35#issue-1128264449) by @sboeuf, the kernel headers are backwards compatible, so there's no need to maintain the bindings for multiple kernel versions.  And doing so will make updating the bindings easier, as it won't require adding a new feature each time.

But before we can do that, we actually have to re-add the kernel version features to the Cargo.toml, because they were lost when virtio-bindings was imported into vm-virtio, and this would break existing users of the crate next time a release is made.

It's unfortunately not possible for us to emit a deprecation warning for the features, as in Rust deprecation warnings have to be applied to functions/constants/etc. (not modules or re-exports](https://github.com/rust-lang/rust/issues/30827)), so any deprecation warning in this crate would mean editing the generated bindings files.  But it doesn't really do any harm to keep the feature declarations in the Cargo.toml around indefinitely.

### Requirements

requirements:

  description line.
  test.


alyssais:
The CI failures seem to be due to the bindings being from a very old bindgen.  That will be one of my next PRs, but it's not related to this one.  I could add that onto the end of this PR to make CI happy if that's required.

crates/virtio-bindings/src/lib.rs
```diff
@@ -1,24 +1,15 @@
 // Copyright 2019 Red Hat, Inc. All Rights Reserved.
 // SPDX-License-Identifier: (BSD-3-Clause OR Apache-2.0)
 
-#[cfg(all(feature = "virtio-v4_14_0", not(feature = "virtio-v5_0_0")))]
-mod bindings_v4_14_0;
-#[cfg(feature = "virtio-v5_0_0")]
-mod bindings_v5_0_0;
+#![allow(clippy::all)]
```
andreeaflorescu, line 4: I don't think this works as you intend it to work, which is also why the tests are failing. We want to allow clippy::all only in the bindings, now you are actually allowing it on the `virtio_blk` module which is the one that is immediately after the clippy macro.

alyssais: Ah, thank you!

alyssais: Actually I think this was correct.  My understanding is that `#!` applies to the whole of the current module, whereas `#` applies only to the next item.

The tests were failing with normal Rust warnings, not clippy ones, because I had accidentally dropped this line:

https://github.com/rust-vmm/vm-virtio/blob/c5895a61f6a4306d9167f33a7dba81ce29c7b001/crates/virtio-bindings/src/bindings_v5_0_0/mod.rs#L8-L9

Now that I've added it back as well, they're passing.

sboeuf (dismissed):
Looking good!

andreeaflorescu (dismissed):

andreeaflorescu (approved):

lauralt (approved):

PR #194, opened 2022-09-14
virtio-bindings: regenerate

alyssais:
### Summary of the PR

Here I've regenerated the bindings with modern versions of bindgen, Glibc, and Linux.  I put each tool update into its own commit to make it easy to see which upgrade caused which changes.

Something that worries me a little bit is that as part of the update to the kernel headers, some structs have gained members, and virtio_net_hdr_v1 has lost its Debug, Default, and PartialEq implementations due to the addition of a union member.  Both of these are technically breaking changes in the Rust sense, even if they're not considered so by the kernel.  But they're also minor, so I'm not sure if it makes sense to worry about them? cc @sboeuf

### Requirements

requirements:

  description line.
  test.


crates/virtio-bindings/src/virtio_net.rs
```diff
@@ -536,17 +607,308 @@ fn bindgen_test_layout_virtio_net_config() {
...
+    pub __bindgen_anon_1: virtio_net_hdr_v1__bindgen_ty_1,
+    pub num_buffers: __virtio16,
+}
+#[repr(C)]
+#[derive(Copy, Clone)]
+pub union virtio_net_hdr_v1__bindgen_ty_1 {
+    pub __bindgen_anon_1: virtio_net_hdr_v1__bindgen_ty_1__bindgen_ty_1,
+    pub csum: virtio_net_hdr_v1__bindgen_ty_1__bindgen_ty_2,
+    pub rsc: virtio_net_hdr_v1__bindgen_ty_1__bindgen_ty_3,
+}
```
sboeuf, line 679: @alyssais I think it's not acceptable to lose the traits implementations (Debug, Default and PartialEq) since this might introduce some regressions for consumers of this crate. The only alternative I can think of is by manually implementing these traits for `virtio_net_hdr_v1__bindgen_ty_1`.

sboeuf: Well the point about added members in structures doesn't really apply given we already agreed to follow what the kernel does, meaning it's acceptable to have new fields in a structure, and things like JSON encoding/decoding should be able to handle such situation from the consumer's code.
But in general I agree let's not make things too complex to maintain, meaning the second approach using minor (or major) version to introduce breaking changes (from a Rust perspective) would be the way to go IMO.

sboeuf: /cc @andreeaflorescu @jiangliu @slp wdyt?

andreeaflorescu: If I remember correctly PartialEq, Debug and Default were not added to all structures by bindgen because of an issue in how they're generating the bindings. We added them manually for older versions because otherwise the structures are pretty hard to use. I think we need to find a way to still have these traits implemented. Let me do some digging and figure out how we handle this in the past.

alyssais: Ah, that explains why I couldn't find a version of bindgen that exactly reproduced the bindings as they were before, so I suppose the generated bindings files were just edited?  It would be good to find a way around that, as it's useful for auditability if the generated files can be exactly reproduced.

andreeaflorescu: The way that we did this is to have one commit with the autogenerated bindings, and then separate commits for the changes. Than the changes were also added to the repository as patch files so that the process can be reproduced.

andreeaflorescu: If you have some more time to invest in this, we can also find a better way of generating them while still keeping the derives that we need. In any case, I don't think it's acceptable to not have them.

andreeaflorescu: @alyssais did you give this more thought? Are you still interested in regenerating the bindings?

alyssais: Hi! I'm still interested, but likely won't get a chance to look at it until I'm back from vacation in mid-October.

andreeaflorescu: Ok, no problem, thanks for letting me know!

alyssais: Quick update: after I came back from my vacation I immediately got a bad flu, so I'm just catching up now.  I'll look at this soon. :)

alyssais:
Updated to:

- Manually implement `Debug` and `PartialEq` for `virtio_net_hdr_v1`.  (bindgen still generates the `Default` impl, it's just not derived any more.)
- Use Linux 6.0 and bindgen 0.61.0.

andreeaflorescu:
@alyssais the build is failing because of some problems with how we were specifying dependencies. Once we merge #202 that shouldn't be a problem anymore.

alyssais:
Should be all sorted and ready for review. :)

crates/virtio-bindings/src/virtio_net/mod.rs
```diff
@@ -0,0 +1,48 @@
...
+pub use generated::*;
+
+use std::fmt::{Debug, Formatter, Result};
+use std::mem::{size_of, transmute};
+
+impl Debug for virtio_net_hdr_v1 {
+    fn fmt(&self, f: &mut Formatter) -> Result {
+        // SAFETY: As of Linux v6.0, all union fields have compatible types.
+        // https://doc.rust-lang.org/reference/items/unions.html#reading-and-writing-union-fields
+        let csum = unsafe { self.__bindgen_anon_1.csum };
```
andreeaflorescu, line 14: This will not really print correct values on big endian platforms, but I am not sure how many people would actually run this on big endian to begin with.

andreeaflorescu (commented):
Mostly looks good to me, I have 2 comments. Sorry for taking so long to review your code, it slipped my mind :(

crates/virtio-bindings/src/virtio_net/mod.rs
```diff
@@ -0,0 +1,48 @@
+// Copyright 2022 Unikie
+// SPDX-License-Identifier: BSD-3-Clause OR Apache-2.0
+
+mod generated;
+pub use generated::*;
+
+use std::fmt::{Debug, Formatter, Result};
+use std::mem::{size_of, transmute};
+
+impl Debug for virtio_net_hdr_v1 {
```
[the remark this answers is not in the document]
alyssais: I like this way much better.  It was confusing to both me and @vireshk in https://github.com/rust-vmm/vm-virtio/pull/200#issuecomment-1276378947 that we couldn't find a version of bindgen that matched the code in the crate.

Historically the bindings haven't been updated very often, and not needing to edit the bindings also makes it much easier (a single command) to update the bindings.  I think that will be good for the library. :) (I'm hoping I'll be able to put some work in to keeping the bindings more up to date going forward.)

crates/virtio-bindings/src/virtio_net/mod.rs
```diff
@@ -0,0 +1,48 @@
...
+    fn fmt(&self, f: &mut Formatter) -> Result {
+        // SAFETY: As of Linux v6.0, all union fields have compatible types.
```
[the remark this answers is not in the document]
alyssais: Same size, alignment, and permitted values.  It's a term I've seen used in Rust, but looking into it now it's perhaps not as common as I thougth.

crates/virtio-bindings/src/virtio_net/mod.rs
alyssais: That's also the case for all the generated Debug implementations too, right?  I could probably fix it here, but I think being inconsistent with all the other types would probably be even more confusing.

crates/virtio-bindings/src/virtio_net/mod.rs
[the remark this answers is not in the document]
alyssais: I see what you mean about e.g. `:#?` though — I'll see if I can fix the implementation to handle that.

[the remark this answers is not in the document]
alyssais: Fixed, and test added.  I think the implementation is nicer now too — I didn't know that the standard library had public stuff to help with debug implementations before!

alyssais:
AFAICT, CI failed because it timed out updating the crates.io index?  And also for the same reason as #206, i.e. https://github.com/bheisler/criterion.rs/issues/629.

lauralt:
> AFAICT, CI failed because it timed out updating the crates.io index? And also for the same reason as #206, i.e. [bheisler/criterion.rs#629](https://github.com/bheisler/criterion.rs/issues/629).

Yes, we need to ignore it as we already did here: https://github.com/rust-vmm/linux-loader/commit/2580d45f741988468e9b086adbcadae7cc7433a5.

lauralt:
And we need to fix and merge the rust-vmm-ci bump, the timeout problem you mentioned is (temporarily) fixed [here](https://github.com/rust-vmm/rust-vmm-ci/commit/aa33f19d6bf4b5eef838d127b09c28900cafbdd6). I will try to fix this asap, but don't quite have bandwidth today.

crates/virtio-bindings/src/virtio_net/mod.rs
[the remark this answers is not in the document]
andreeaflorescu: Maybe you could add this explicitly in the comment as well, I think it would clarify things.

andreeaflorescu:
> And we need to fix and merge the rust-vmm-ci bump, the timeout problem you mentioned is (temporarily) fixed [here](https://github.com/rust-vmm/rust-vmm-ci/commit/aa33f19d6bf4b5eef838d127b09c28900cafbdd6). I will try to fix this asap, but don't quite have bandwidth today.

I did the fix, you can just rebase on top of the main branch and there should be no more failures.

crates/virtio-bindings/src/virtio_net/mod.rs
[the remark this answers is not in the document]
alyssais: Done.

andreeaflorescu (approved):

lauralt (approved):

PR #195, opened 2022-09-27
Implement Fuzzing for Virtio Devices

andreeaflorescu:
### Summary of the PR

This PR adds fuzzing for the following crates: virtio-queue, virtio-queue-ser, and virtio-vsock.
Details about the design choices can be found in the associated read me.

**NOTE**: As not all commits are committed by myself, I'll address PR comments in separate commits.

### Requirements

requirements:

  description line.
  test.


rbradford:
@likebreath as you have been working on the fuzzing in Cloud Hypervisor I think it would be great if you reviewed this

fuzz/README.md
```diff
@@ -0,0 +1,102 @@
...
+## Running the Fuzzing Session
+
+Taking as an example fuzzing for the Virtio Queue, we are working with the following modules and
+directories:
+- `common/src/virtio_queue.rs` -> defines the structures used for fuzzing and a set of tests that
+  when ran are generating a corpus for the fuzzer.
+- `corpus/virtio_queue` -> the path of the corpus used when running the fuzzing session. This is
+  also the path where we are writing the files generated by running the test cases.
+- `fuzz_targets/virtio_queue.rs` -> the path for the Virtio Queue fuzz target.
+
```
stsquad, line 72: It might be worth adding a pre-requisites section here. I found I needed to run:

  cargo install cargo-fuzz
  
first

fuzz/common/src/vsock.rs
```diff
@@ -0,0 +1,409 @@
...
+    const SRC_CID: u64 = 1;
+    const DST_CID: u64 = 2;
+    const SRC_PORT: u32 = 3;
+    const DST_PORT: u32 = 4;
+    const LEN: u32 = 16;
+    const TYPE: u16 = 5;
+    const OP: u16 = 6;
+    const FLAGS: u32 = 7;
+    const FLAG: u32 = 8;
+    const BUF_ALLOC: u32 = 256;
```
stsquad, line 192: I'm slightly confused by the magic numbers for LEN and BUF_ALLOC when the comment implies these are all random values which have otherwise added sequentially. Does that mean there are some bounds for some of these values?

stsquad (approved):
Looks good to me - only minor nits for docs and a question about random numbers.

fuzz/common/src/vsock.rs
lauralt: No particular reason for having these values, other than they look more like sizes. We can update them if it creates confusion.

andreeaflorescu: If I remember correctly I just took these from the existing Vsock Packet tests. I can add some more documentation about it being rather arbitrary.

jiangliu (approved):

fuzz/README.md
andreeaflorescu: I noticed that there are some other outdated things in the README so I am thinking about opening a separate PR to address them along with this comment.

andreeaflorescu:
@liuw would you like to review this? Should I wait for your review?

.buildkite/custom-tests.json
```diff
@@ -24,6 +24,21 @@
       "test_name": "fuzz-build",
       "command": "cd fuzz && cargo +nightly build",
       "platform": ["x86_64", "aarch64"]
+    },
+    {
+      "test_name": "fuzz-virtio-queue",
+      "command": "cd fuzz && cargo test --package common --lib virtio_queue && cargo +nightly fuzz run virtio_queue -- -max_total_time=240",
```
likebreath: The the default timeout from cargo fuzz is 1200s (inherited from `libfuzzer`), and is too big to report any related issues. Making the timeout `60s` or so should be more suitable.

andreeaflorescu: Hmmm, I do not completely understand why would we make the timeout less. My goal is to run for 15 minutes because that's not delaying the merge of PRs too much and also reaches a pretty good coverage in the given amount of time. Why are you suggesting the 60s?

likebreath: @andreeaflorescu The value of `-max_total_time` (15 mins) looks fine to me. I was talking about reducing the default value of `-timeout`, which limits the amount of time for a single run before it considers that run a failure/hanging [1]. The default value (1200s) is just too big to report meaningful hanging issues. A timeout of `60s` is used by the `oss-fuzz` project. I hope it clarifies the confusions.

[1] https://llvm.org/docs/LibFuzzer.html#options

andreeaflorescu: Ohhh, I see what you mean. That's super interesting. I'll make this update in a future PR if that's ok with you.

PR #196, opened 2022-09-27
virtio-queue: Error out early on invalid available ring index

likebreath:
The number of descriptor chain heads to process should always be smaller or equal to the queue size, as the driver should never ask the VMM to process a available ring entry more than once. Checking and reporting such incorrect driver behavior can avoid potential hanging and Denial-of-Service from VMMs.

Signed-off-by: Bo Chen <chen.bo@intel.com>

### Summary of the PR

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/crates/virtio-queue/src/queue.rs
+++ b/crates/virtio-queue/src/queue.rs
@@ -612,7 +612,7 @@ impl QueueOwnedT for Queue {
         M::Target: GuestMemory,
     {
         self.avail_idx(mem.deref(), Ordering::Acquire)
-            .map(move |idx| AvailIter::new(mem, idx, self))
+            .map(move |idx| AvailIter::new(mem, idx, self))?
     }
 
     fn go_to_previous_position(&mut self) {
@@ -710,15 +710,25 @@ where
     ///           available descriptor chain.
     /// * `queue` - the `Queue` object from which the needed data to create the `AvailIter` can
     ///             be retrieved.
-    pub(crate) fn new(mem: M, idx: Wrapping<u16>, queue: &'b mut Queue) -> Self {
-        AvailIter {
+    pub(crate) fn new(mem: M, idx: Wrapping<u16>, queue: &'b mut Queue) -> Result<Self, Error> {
+        // The number of descriptor chain heads to process should always
+        // be smaller or equal to the queue size, as the driver should
+        // never ask the VMM to process a available ring entry more than
+        // once. Checking and reporting such incorrect driver behavior
+        // can prevent potential hanging and Denial-of-Service from
+        // happening on the VMM side.
+        if (idx - queue.next_avail).0 > queue.size {
+            return Err(Error::InvalidAvailRingIndex);
+        }
+
+        Ok(AvailIter {
             mem,
             desc_table: queue.desc_table,
             avail_ring: queue.avail_ring,
             queue_size: queue.size,
             last_index: idx,
             next_avail: &mut queue.next_avail,
-        }
+        })
     }
 
     /// Goes back one position in the available descriptor chain offered by the driver.
@@ -1238,18 +1248,8 @@ mod tests {
         // Decrement `idx` which should be forbidden. We don't enforce this thing, but we should
         // test that we don't panic in case the driver decrements it.
         vq.avail().idx().store(u16::to_le(1));
-
-        loop {
-            q.disable_notification(mem).unwrap();
-
-            while let Some(_chain) = q.iter(mem).unwrap().next() {
-                // In a real use case, we would do something with the chain here.
-            }
-
-            if !q.enable_notification(mem).unwrap() {
-                break;
-            }
-        }
+        // Invalid available ring index
+        assert!(q.iter(mem).is_err());
     }
 
     #[test]
--- a/crates/virtio-queue/src/queue.rs
+++ b/crates/virtio-queue/src/queue.rs
@@ -1253,6 +1253,51 @@ mod tests {
     }
 
     #[test]
+    fn test_iterator_and_avail_idx() {
+        // This test ensures constructing a descriptor chain iterator succeeds
+        // with valid available ring indexes while produces an error with invalid
+        // indexes.
+        let queue_size = 2;
+        let mem = &GuestMemoryMmap::<()>::from_ranges(&[(GuestAddress(0), 0x10000)]).unwrap();
+        let vq = MockSplitQueue::new(mem, queue_size);
+
+        let mut q: Queue = vq.create_queue().unwrap();
+
+        // q is currently valid.
+        assert!(q.is_valid(mem));
+
+        // Create descriptors to fill up the queue
+        let mut descs = Vec::new();
+        for i in 0..queue_size {
+            descs.push(Descriptor::new(
+                (0x1000 * (i + 1)) as u64,
+                0x1000,
+                0_u16,
+                i + 1,
+            ));
+        }
+        vq.add_desc_chains(&descs, 0).unwrap();
+
+        // Set the 'next_available' index to 'u16:MAX' to test the wrapping scenarios
+        q.set_next_avail(u16::MAX);
+
+        // When the number of chains exposed by the driver is equal to or less than the queue
+        // size, the available ring index is valid and constructs an iterator successfully.
+        let avail_idx = Wrapping(q.next_avail()) + Wrapping(queue_size);
+        vq.avail().idx().store(avail_idx.0);
+        assert!(q.iter(mem).is_ok());
+        let avail_idx = Wrapping(q.next_avail()) + Wrapping(queue_size - 1);
+        vq.avail().idx().store(avail_idx.0);
+        assert!(q.iter(mem).is_ok());
+
+        // When the number of chains exposed by the driver is larger than the queue size, the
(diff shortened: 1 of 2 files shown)


crates/virtio-queue/src/queue.rs
```diff
@@ -611,8 +611,18 @@ impl QueueOwnedT for Queue {
...
     {
-        self.avail_idx(mem.deref(), Ordering::Acquire)
-            .map(move |idx| AvailIter::new(mem, idx, self))
+        let idx = self.avail_idx(mem.deref(), Ordering::Acquire)?;
+
+        // The number of descriptor chain heads to process should always
+        // be smaller or equal to the queue size, as the driver should
+        // never ask the VMM to process a available ring entry more than
+        // once. Checking and reporting such incorrect driver behavior
+        // can avoid potential hanging and Denial-of-Service from VMMs.
```
andreeaflorescu, line 620: Do you have a reproduction for this DoS? When does this happen?

When you call `next` on the returned object from this function we already check that it is less then size because we're doing a modulo: https://github.com/rust-vmm/vm-virtio/blob/7e203db6ed044217acd80dec5967c15d979491f8/crates/virtio-queue/src/queue.rs#L750.

According to the spec the driver is allowed to pass a number that is larger than QUEUE_SIZE: https://docs.oasis-open.org/virtio/virtio/v1.1/csprd01/virtio-v1.1-csprd01.html#x1-380006
"idx field indicates where the driver would put the next descriptor entry in the ring (modulo the queue size)."

rbradford: Not sure the logic in the `if` block is correct (because of wrapping) but the goal is to avoid iterating and processing over the same descriptors over and over again. 

e.g. if the queue size was `n` and guest set the next descriptor  to `current + 1000 * n` it would force the virtio device to consider each descriptor in the array 1000 times. This would effectively DoS the thread processing the queue especially if the computation on the descriptor is costly. This was found with fuzzing.

rbradford: This manifests itself by `.next()` continuing to return descriptor chains until `next_avail` and `last_index` are equal: https://github.com/rust-vmm/vm-virtio/blob/main/crates/virtio-queue/src/queue.rs#L743

andreeaflorescu: @rbradford @likebreath do you think you can work on a unit test that exemplifies this scenario? We can use it as a regression test for the fix. I am trying to do that myself, but I am still not very sure how this can happen. The guest can flood the VMM with requests nevertheless, so some sort of a rate limiter needs to be implemented at the VMM side nevertheless. If we can improve the situation in any case at the device level, that would be great as well.

rbradford: > The guest can flood the VMM with requests nevertheless, so some sort of a rate limiter needs to be implemented at the VMM side nevertheless. If we can improve the situation in any case at the device level, that would be great as well.

A well behaved guest would never do this because they wouldn't want the same descriptors handled multiple times. I don't know if we could also use the used ring to identify this troublesome behaviour.

rbradford: > I don't know if we could also use the used ring to identify this troublesome behaviour.

I'm pretty sure we can't because it's only written by the device.

andreeaflorescu: Hmm, I think the fix here might be to do the mod when we create the `AvailIter` because then the if condition in `next` will correctly return None.

dianpopa: > Do you have a reproduction for this DoS? When does this happen?
> 
> When you call `next` on the returned object from this function we already check that it is less then size because we're doing a modulo:
> 
> https://github.com/rust-vmm/vm-virtio/blob/7e203db6ed044217acd80dec5967c15d979491f8/crates/virtio-queue/src/queue.rs#L750
> 
> .
> According to the spec the driver is allowed to pass a number that is larger than QUEUE_SIZE: https://docs.oasis-open.org/virtio/virtio/v1.1/csprd01/virtio-v1.1-csprd01.html#x1-380006 "idx field indicates where the driver would put the next descriptor entry in the ring (modulo the queue size)."

Indeed the idx field (which is read from guest memory  through the `self.avail_idx()` fn) can be larger than QUEUE_SIZE but my understanding is that the current fix puts a limit on the number of available descriptors (i.e idx - self.next_avail). 
If we allow the number of available descriptors to be higher than queue size, we end up processing some descriptors more than once.

rbradford: > If we allow the number of available descriptors to be higher than queue size, we end up processing some descriptors more than once.

Exactly. I can't think of any reason why a well behaved guest would want that.

andreeaflorescu: > Hmm, I think the fix here might be to do the mod when we create the `AvailIter` because then the if condition in `next` will correctly return None.

This is not actually correct because the spec says that the [idx grows and wraps around at u16::MAX](https://docs.oasis-open.org/virtio/virtio/v1.1/csprd01/virtio-v1.1-csprd01.html#x1-55003r55). I've mis-read the spec. The modulo was referring to writing elements in the vring.

I also looking how other projects are addressing this problem. We can't directly do `idx - self.next_avail` because that operation might underflow and cause a panic. So we must do a checked_sub if we want to check that the same descriptor is not processed multiple times.

rbradford: > We can't directly do `idx - self.next_avail` because that operation might underflow and cause a panic. 

Both these are `Wrapping<u16>` so I think that is it safe.

andreeaflorescu: Looks like QEMU has the same check in place: https://github.com/qemu/qemu/blob/master/hw/virtio/virtio.c#L966

rbradford: > Looks like QEMU has the same check in place: https://github.com/qemu/qemu/blob/master/hw/virtio/virtio.c#L966

Cool. Do you think this is the right place or should it be in the implementation of `.next()` ?

crates/virtio-queue/src/queue.rs
```diff
@@ -611,8 +611,18 @@ impl QueueOwnedT for Queue {
         M: Deref,
         M::Target: GuestMemory,
     {
-        self.avail_idx(mem.deref(), Ordering::Acquire)
-            .map(move |idx| AvailIter::new(mem, idx, self))
+        let idx = self.avail_idx(mem.deref(), Ordering::Acquire)?;
+
+        // The number of descriptor chain heads to process should always
```
rbradford, line 616: I don't think this comment or that in the commit message is quite right. The VMM isn't the part that can cause the DoS it's the guest.

crates/virtio-queue/src/queue.rs
andreeaflorescu: > > We can't directly do `idx - self.next_avail` because that operation might underflow and cause a panic.
> 
> Both these are `Wrapping<u16>` so I think that is it safe.

Right, that is safe because it will return u16::MAX which is going to be > queue size irrespective of the queue size set by the driver.

crates/virtio-queue/src/queue.rs
```diff
@@ -1238,18 +1248,8 @@ mod tests {
...
-
-            while let Some(_chain) = q.iter(mem).unwrap().next() {
-                // In a real use case, we would do something with the chain here.
-            }
-
-            if !q.enable_notification(mem).unwrap() {
-                break;
-            }
-        }
+        // Invalid available ring index
```
andreeaflorescu, line 1251: We should add some additional tests here:
1. A regression that fails before the fix and passes afterwards. We can write a unit test for the scenario described by Rob where the 
2. A test to check that you can still handle exactly queue size descriptor chains.

crates/virtio-queue/src/queue.rs
andreeaflorescu: > Cool. Do you think this is the right place or should it be in the implementation of `.next()` ?

I am checking to see if there is any other code path that could trigger this behavior outside of `iter`. It might make sense to have it in the constructor of `AvailIter`.

rbradford: So @likebreath looks like the actions that you need to do on this PR are address the comment I made about the commit message/comment and add a unit tests to ensure that invalid and only invalid ranges generate an error.

crates/virtio-queue/src/queue.rs
likebreath: Thank you for letting me know the message was misleading. That was not what I intended to say. 

Both the comments and commit message are updated. PTAL.

crates/virtio-queue/src/queue.rs
likebreath: A new unit test is added from a separate commit. Anything I am missing, please let me know. Thank you!

likebreath:
Thank you all the the feedbacks. All comments should be addressed. PTAL. 

A summary of changes:
1. Moved the check for invalid available ring index to the constructor of `AvailIter`;
2. Clarified the commit message and comments;
3. Added a related unit test;

lauralt (approved):

sboeuf (approved):

rbradford (approved):

andreeaflorescu (approved):

PR #200, opened 2022-10-10
virtio-bindings: Add bindings for more virtio files

vireshk:
Add auto-generated bindings for include/linux/virtio_mmio.h and include/linux/virtio_config.h from v5.0 kernel using bindgen 0.49.0.

While at it, also update existing bindings to match these versions.


vireshk:
These are required for the Xen-vhost-frontend crate we are working on. Thanks.

@andreeaflorescu @slp @jiangliu @alexandruag

crates/virtio-bindings/src/virtio_config.rs
```diff
@@ -0,0 +1,127 @@
+/* automatically generated by rust-bindgen 0.59.1 */
```
lauralt, line 1: The PR description says the bindings are generated with 0.6.0, but the comment says another thing. Also, shouldn't all the bindings in the repository be generated with the same version?

vireshk: Description is talking about Linux kernel version v6.0.

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tag/?h=v6.0

And yes, maybe all bindings should be generated together.

lauralt: Ah never mind, I confused the kernel version with the rust-bindgen one, but the question still stands, are all the bindings generated with the same rust-bindgen version?

vireshk: @lauralt I have updated all bindigns to bindgen 0.60.1 and kernel v6.0 now.

crates/virtio-bindings/CONTRIBUTING.md
```diff
@@ -6,7 +6,7 @@
 The bindings are currently generated using
 [bindgen](https://crates.io/crates/bindgen) version 0.49.0:
```
lauralt, line 7: Comments should be updated as well.

vireshk: Fixed. Thanks.

lauralt:
Please also see [this discussion](https://github.com/rust-vmm/vm-virtio/pull/194#discussion_r971694119). if we switch to these versions, looks like we are losing some trait derives/implementations.

vireshk:
> Please also see [this discussion](https://github.com/rust-vmm/vm-virtio/pull/194#discussion_r971694119). if we switch to these versions, looks like we are losing some trait derives/implementations.

Hmm, maybe I should just move back to 0.49 then and regenerate everything ? I am not particular about newer bindgen anyway, I though it would be better to use latest :)

vireshk:
> Yes let's use the one we had before until we figure out what to do with 0.60.

Done

lauralt (approved):

lauralt:
@vireshk can you check the CI failures?

vireshk:
> @vireshk can you check the CI failures?

This is strange, and sorry about not looking at them earier. I somehow expected these changes to never have any build issues since these are just bindgen changes :(

I am not sure why bindgen is adding Copy, Clone derives here, when it didn't do that earlier. I also tried with kernel v5.0 and still same issues. Will look more into it, but clueless at the moment.

And I hope I am not supposed to manually edit these files to make these issues go away, right ?

lauralt:
@vireshk The autogenerated bindings were indeed edited, that is also discussed in the thread I sent to you, e.g. https://github.com/rust-vmm/vm-virtio/pull/194#discussion_r974053605.

vireshk:
> @vireshk The autogenerated bindings were indeed edited, that is also discussed in the thread I sent to you, e.g. [#194 (comment)](https://github.com/rust-vmm/vm-virtio/pull/194#discussion_r974053605).

Ahh, I missed the comment for sure :(

Though I went through git history yesterday for virtio-bindings (previous individual crate too) and couldn't find any separate patches to make these manual changes. Though I see that Andreea did comment that it was done that way in the other thread, not sure why I don't see that.

Anyway, I know what to do now :)

vireshk:
> If the version that is used for the existing bindings is okay for you, I would suggest to use that one for the bindings that you are adding as well (since we already have a PR in review that is updating them). What do you think?

That is the version, 0.49.0, that I used to generate the last pull request. But I didn't update them manually to fix warnings.

crates/virtio-bindings/src/virtio_net.rs
```diff
@@ -583,7 +583,7 @@ fn bindgen_test_layout_virtio_net_config() {
     );
 }
 #[repr(C)]
-#[derive(Debug, Default, Copy, Clone, PartialEq)]
+#[derive(Default, Copy, Clone)]
```
lauralt, line 586: I think the discussion in the other PR was in favor of not removing these, but adding manual implementations.

vireshk: Hmm, I thought those issues will come only with a newer version of bindgen and 0.49.0 should be safe. And now I am really not sure what was done earlier when these bindings were generated and merged with 0.49.0. They should have kept a separate patch to let us know :)

lauralt: @slp can you help with some context here? I saw you added the bindings in the old repository.

vireshk: @lauralt I now moved back to kernel v5.0 too and those issues are gone.

What we have now is
kernel: v5.0
bindgen: 0.49.0

Hope this will work fine now and the other thread can conclude about the updates to newer kernel and bindgen.,

lauralt (approved):
👍

crates/virtio-bindings/src/virtio_mmio.rs
```diff
@@ -0,0 +1,31 @@
+/* automatically generated by rust-bindgen */
```
andreeaflorescu, line 1: Since we now have these defined here, should we actually use them in the virtio-device implementation instead of having magic offsets: https://github.com/rust-vmm/vm-virtio/blob/main/crates/virtio-device/src/mmio.rs#L72?

We will need to disambiguate between the offsets (which are defined in this file) and the actual values, defined in mmio.rs. It's a bit of bummer that these defines don't have `OFFSET` in the name :( but nothing we can do about that.

vireshk: > Since we now have these defined here, should we actually use them in the virtio-device implementation instead of having magic offsets: https://github.com/rust-vmm/vm-virtio/blob/main/crates/virtio-device/src/mmio.rs#L72?

Makes sense for sure. Should I do it in this series itself or do it once this is merged ?

> We will need to disambiguate between the offsets (which are defined in this file) and the actual values, defined in mmio.rs.

Well the offsets are prefixed with VIRTIO_, which the values aren't. Anything else we want to update in mmio.rs.

andreeaflorescu: Ah, I missed that, then we don't need to update anything else. I would say to add the change in the same PR because this is the PR that adds the values which make the change needed.

vireshk: Done.

andreeaflorescu (dismissed):

crates/virtio-device/src/mmio.rs
```diff
@@ -69,30 +74,32 @@ pub trait VirtioMmioDevice: WithDriverSelect {
...
                         .unwrap_or(false)
                         .into(),
-                    0x60 => self.interrupt_status().load(Ordering::SeqCst).into(),
-                    0x70 => self.device_status().into(),
-                    0xfc => self.config_generation().into(),
+                    VIRTIO_MMIO_INTERRUPT_STATUS => {
+                        self.interrupt_status().load(Ordering::SeqCst).into()
+                    }
+                    VIRTIO_MMIO_STATUS => self.device_status().into(),
+                    VIRTIO_MMIO_CONFIG_GENERATION => self.config_generation().into(),
```
lauralt, line 102: 0x100 below also has a definition.

crates/virtio-device/src/mmio.rs
```diff
@@ -69,30 +74,32 @@ pub trait VirtioMmioDevice: WithDriverSelect {
         match offset {
             // The standard specifies that accesses to configuration registers are 32-bit wide.
             0x00..=0xff if data.len() == 4 => {
-                let v = match offset {
-                    0x0 => MMIO_MAGIC_VALUE,
-                    0x04 => MMIO_VERSION,
-                    0x08 => self.device_type(),
-                    0x0c => VENDOR_ID,
-                    0x10 => match self.device_features_select() {
+                let v = match offset as u32 {
```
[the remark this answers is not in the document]
andreeaflorescu: A good explanation here is that we are not loosing any important information if the driver sent a value that is more than the accepted u32 because in that case "undefined behavior" is expected.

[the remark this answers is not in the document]
lauralt: This is not needed. if we got here, we know that offset is between 0x00 and 0xff which is less than u32::max.

lauralt (approved):

andreeaflorescu (approved):

Issue #204, opened 2022-11-24
virtio-vsock: spec violation about the number of descriptors per packet


stefano-garzarella:
In `crates/devices/virtio-vsock/src/packet.rs`  for both the TX and RX queue, we assume that each packet always consists of 2 descriptors (returning `DescriptorChainTooShort` error otherwise), but this is not in the [specification](https://docs.oasis-open.org/virtio/virtio/v1.2/cs01/virtio-v1.2-cs01.html).

It was an implementation detail true until Linux 6.1, but now it's about to change.
Testing [vhost-user-vsock](https://github.com/rust-vmm/vhost-device/tree/main/crates/vsock) (which uses this crate) with the [proposed Linux patch](https://lore.kernel.org/lkml/20221124060750.48223-1-bobby.eshleman@bytedance.com/), I noticed that it doesn't work, printing this error:
```
WARN  vhost_user_vsock::vhu_vsock_thread] vsock: RX queue error: DescriptorChainTooShort
```

In the vhost-vsock device available in Linux, we don't assume that, so I think we should do the same here to respect the virtio specification.

stefano-garzarella:
I would like to fix it before we release the new driver (not before Linux 6.2).
@lauralt can you take a look at it? If not, I'll try to see it next week.

lauralt:
Yes, the implementation was Linux specific at that point, it is mentioned in [a doc comment](https://github.com/rust-vmm/vm-virtio/blob/main/crates/devices/virtio-vsock/src/packet.rs#L15) as well. I agree this should be fixed, but other things were prioritised.
Having a look again at the code, it won't be super straightforward to fix this, and it will probably be ugly since we don't have the `Reader` and `Writer` abstractions that crosvm for example uses (there is a PR open for this in vm-virtio as well https://github.com/rust-vmm/vm-virtio/pull/33 for quite some time already). I can take a better look early next week.

Out of curiosity, does it fail with the error you mentioned when having multiple descriptors for the data or?

stefano-garzarella:
> Yes, the implementation was Linux specific at that point, it is mentioned in [a doc comment](https://github.com/rust-vmm/vm-virtio/blob/main/crates/devices/virtio-vsock/src/packet.rs#L15) as well. 

I didn't know that, thanks for pointing it out!

[...]

PR #207, opened 2022-12-19
vsock: support header and data on a single descriptor

stefano-garzarella:
### Summary of the PR

Starting from Linux 6.2 the virtio-vsock driver can use a single descriptor for both header and data:
https://lore.kernel.org/lkml/20221215043645.3545127-1-bobby.eshleman@bytedance.com/

So with this modification the virtio-vsock device can support header and data on a single descriptor or on two. This is just a Linux implementation detail though, for the spec it could be multiple descriptors as well.

For now let's add only the single descriptor support, since it doesn't involve an API change, but in the future we should change the API and support an array of VolatileSlices as suggested by Laura here: https://github.com/rust-vmm/vm-virtio/issues/204#issuecomment-1332511007

Let's make the tests work with this change (mainly by always using PKT_HEADER_SIZE for the header descriptor) and adding new tests to cover the single descriptor scenario.

Fixes: #204

### Requirements

requirements:

  description line.
  test.


lauralt:
> About the CI failure on the commit message, should I remove the links? 

No, we can merge the PR with admin rights when it'll be the time.

andreeaflorescu:
@stefano-garzarella The tests that we are now using for generating the fuzz input are failing, can you take a look?
https://buildkite.com/rust-vmm/vm-virtio-ci/builds/728#01852a4d-1e33-4d66-885b-006d851b7344

stefano-garzarella:
> @stefano-garzarella The tests that we are now using for generating the fuzz input are failing, can you take a look? https://buildkite.com/rust-vmm/vm-virtio-ci/builds/728#01852a4d-1e33-4d66-885b-006d851b7344

Ooops, I forgot to adjust the descriptors in the fuzz subfolder.

andreeaflorescu:
@stefano-garzarella if you do a rebase on top of main the `cargo audit` test should now pass, and we shouldn't see any more failures. Besides the commit message test all the other tests should pass to merge this PR.

crates/devices/virtio-vsock/src/packet.rs
```diff
@@ -460,19 +460,30 @@ impl<'a, B: BitmapSlice> VsockPacket<'a, B> {
...
         }
 
-        let data_desc = desc_chain.next().ok_or(Error::DescriptorChainTooShort)?;
+        // Starting from Linux 6.2 the virtio-vsock driver can use a single descriptor for both
+        // header and data.
+        let data_slice = if chain_head.len() >= PKT_HEADER_SIZE as u32 + pkt.len() {
+            mem.get_slice(
+                chain_head
+                    .addr()
+                    .checked_add(PKT_HEADER_SIZE.try_into().unwrap())
```
andreeaflorescu, line 469: nit: Since we are doing a `try_into()` followed by an `unwrap` I think we can just use `PKT_HEADER_SIZE as u64`. This is safe since the PKT_HEADER_SIZE is a usize. Unwraps tend to make you think that something potentially dangerous us going on 😆

crates/devices/virtio-vsock/src/packet.rs
```diff
@@ -460,19 +460,30 @@ impl<'a, B: BitmapSlice> VsockPacket<'a, B> {
             return Err(Error::InvalidHeaderLen(pkt.len()));
         }
 
-        let data_desc = desc_chain.next().ok_or(Error::DescriptorChainTooShort)?;
+        // Starting from Linux 6.2 the virtio-vsock driver can use a single descriptor for both
+        // header and data.
+        let data_slice = if chain_head.len() >= PKT_HEADER_SIZE as u32 + pkt.len() {
```
andreeaflorescu, line 465: The addition operation can overflow. Unfortunately not having a maximum (and reasonable) max packet size means that a wrongly initialized value can lead to an overflow here in case you're also dealing with a malicious user.

crates/devices/virtio-vsock/src/packet.rs
```diff
@@ -583,27 +594,39 @@ impl<'a, B: BitmapSlice> VsockPacket<'a, B> {
...
             .get_slice(chain_head.addr(), PKT_HEADER_SIZE)
             .map_err(Error::InvalidMemoryAccess)?;
-        let data_desc = desc_chain.next().ok_or(Error::DescriptorChainTooShort)?;
 
-        if !data_desc.is_write_only() {
-            return Err(Error::UnexpectedReadOnlyDescriptor);
-        }
+        // Starting from Linux 6.2 the virtio-vsock driver can use a single descriptor for both
+        // header and data.
+        let data_slice = if chain_head.len() as usize > PKT_HEADER_SIZE {
```
andreeaflorescu, line 603: We typically want to compare stuff.len() to the PKT_HEADER_SIZE. The `len` function uses an atypical return type for a `len` function which makes these conversions required. Maybe we can open an issue to fix this separately if this is not on purpose.

CC: @lauralt

crates/devices/virtio-vsock/src/packet.rs
andreeaflorescu: To reproduce:
```rust
        // On the TX path, it is allowed to not have a data descriptor.
        let v = vec![Descriptor::new(0x10_0000, PKT_HEADER_SIZE as u32, 0, 0)];
        let mut chain = queue.build_desc_chain(&v[..1]).unwrap();

        let header = PacketHeader {
            src_cid: SRC_CID.into(),
            dst_cid: DST_CID.into(),
            src_port: SRC_PORT.into(),
            dst_port: DST_PORT.into(),
            len: u32::MAX.into(),
            type_: 0.into(),
            op: 0.into(),
            flags: 0.into(),
            buf_alloc: 0.into(),
            fwd_cnt: 0.into(),
        };
        mem.write_obj(header, GuestAddress(0x10_0000)).unwrap();

        let packet = VsockPacket::from_tx_virtq_chain(&mem, &mut chain, u32::MAX).unwrap();
```

andreeaflorescu (changes requested):
Didn't look over all the code, but I think it needs some adjustments. Will take a look at the rest tomorrow.

crates/devices/virtio-vsock/src/packet.rs
lauralt: > The addition operation can overflow. Unfortunately not having a maximum (and reasonable) max packet size means that a wrongly initialized value can lead to an overflow here in case you're also dealing with a malicious user.

Which reminds me that I should at some point resume work (addressing comments actually) on https://lists.oasis-open.org/archives/virtio-comment/202206/msg00030.html 😄 .

crates/devices/virtio-vsock/src/packet.rs
lauralt: I think what we wanted here was to be consistent with the spec (len is defined as le32 there). We can also update the method to return an usize, but I would be in favor of keeping it as defined in the spec.

crates/devices/virtio-vsock/src/packet.rs
stefano-garzarella: ack, I'll change it!

crates/devices/virtio-vsock/src/packet.rs
stefano-garzarella: Okay, I'll change the check to `if chain_head.len() - PKT_HEADER_SIZE as u32 >= pkt.len()`, so every potential overflow will go to the `else` branch and handled as before this patch.

Is that okay for you?

crates/devices/virtio-vsock/src/packet.rs
stefano-garzarella: I don't have a strong opinion on that, but I agree that we should change it in another PR.

crates/devices/virtio-vsock/src/packet.rs
andreeaflorescu: I think this will still make the fuzzer fail because overflows are panics in debug mode, and if I am not mistaken you can still have an underflow? Will double check.

crates/devices/virtio-vsock/src/packet.rs
andreeaflorescu: Ok, it looks like it's on purpose like this, so I will resolve this.

crates/devices/virtio-vsock/src/packet.rs
stefano-garzarella: At that point, we already checked that `chain_head.len()` is  `>= PKT_HEADER_SIZE`, so I think we can't have underflow or overflow. Or am I forgetting something? (My mind is already in the holiday mood 🎄).

andreeaflorescu: Yeah, I know what you mean, same here 🤶. Let me take a better look after some more ☕

andreeaflorescu:
@stefano-garzarella I created a longer running fuzzing pipeline for PRs that runs fuzzing for 1 day, but doesn't block the merge of PRs. Do you mind if I do an update branch on your PR to trigger the run?

stefano-garzarella:
> @stefano-garzarella I created a longer running fuzzing pipeline for PRs that runs fuzzing for 1 day, but doesn't block the merge of PRs. Do you mind if I do an update branch on your PR to trigger the run?

It's fine, but the current branch doesn't include the change that I proposed to avoid the overflow.
Let me push it and IIUC the 1day pipeline will start, right?

stefano-garzarella:
> > @stefano-garzarella I created a longer running fuzzing pipeline for PRs that runs fuzzing for 1 day, but doesn't block the merge of PRs. Do you mind if I do an update branch on your PR to trigger the run?
> 
> It's fine, but the current branch doesn't include the change that I proposed to avoid the overflow. Let me push it and IIUC the 1day pipeline will start, right?

ah no, I thought was on main. So I'll send the new version and then you can update it.

andreeaflorescu:
> > @stefano-garzarella I created a longer running fuzzing pipeline for PRs that runs fuzzing for 1 day, but doesn't block the merge of PRs. Do you mind if I do an update branch on your PR to trigger the run?
> 
> It's fine, but the current branch doesn't include the change that I proposed to avoid the overflow. Let me push it and IIUC the 1day pipeline will start, right?

Yap, that's right! Once it is pushed once, then we can also retrigger it from github. But since the webhook was yet never called, GitHub doesn't know of its existence.

crates/devices/virtio-vsock/src/packet.rs
```diff
@@ -867,6 +920,33 @@ mod tests {
...
+
+        // Let's also test a valid descriptor chain, with both header and data on a single
+        // descriptor.
+        let v = vec![Descriptor::new(
+            0x5_0000,
+            PKT_HEADER_SIZE as u32 + 0x100,
+            VRING_DESC_F_WRITE as u16,
+            0,
+        )];
+        let mut chain = queue.build_desc_chain(&v[..1]).unwrap();
```
andreeaflorescu, line 932: nit: you can pass `&v` directly.

andreeaflorescu (approved):
We can probably get rid of some of the duplication in the tx and rx path, but I don't think that makes a lot of sense since we want to do a more generic refactoring anyway.

LGTM.

crates/devices/virtio-vsock/src/packet.rs
stefano-garzarella: oh right, copy&past from other tests, I think we can pass `&v` everywhere.
I can send a cleanup on top of this.

crates/devices/virtio-vsock/src/packet.rs
andreeaflorescu: This overflow is indeed no longer possible. Resolving this conversation.

crates/devices/virtio-vsock/src/packet.rs
andreeaflorescu: Yes, right. I noticed that in the other tests as well, but didn't want to burden you to change it all 😆. On the other hand I also don't want to propagate the mistake to other places.

stefano-garzarella: No problem at all.
I'll fix the tests added here and fix the other in a separate commit.
Should I integrate that commit in this PR, or better a new one?

andreeaflorescu: Works either way as long as it's a separate commit. Whichever makes it easier for you.

stefano-garzarella: Okay, I'll include here, less noise for others ;-)

andreeaflorescu (dismissed):

lauralt:
[This documentation](https://github.com/rust-vmm/vm-virtio/blob/main/crates/devices/virtio-vsock/src/packet.rs#L15) needs to be updated.

lauralt:
We should probably also add a CHANGELOG entry for this, even if it's not an API change.

stefano-garzarella:
new version:
- fixed [documentation](https://github.com/rust-vmm/vm-virtio/blob/main/crates/devices/virtio-vsock/src/packet.rs#L15)
- added CHANGELOG entries

crates/devices/virtio-vsock/src/packet.rs
```diff
@@ -583,27 +595,39 @@ impl<'a, B: BitmapSlice> VsockPacket<'a, B> {
...
             .get_slice(chain_head.addr(), PKT_HEADER_SIZE)
             .map_err(Error::InvalidMemoryAccess)?;
-        let data_desc = desc_chain.next().ok_or(Error::DescriptorChainTooShort)?;
 
-        if !data_desc.is_write_only() {
-            return Err(Error::UnexpectedReadOnlyDescriptor);
-        }
+        // Starting from Linux 6.2 the virtio-vsock driver can use a single descriptor for both
+        // header and data.
+        let data_slice = if chain_head.len() as usize > PKT_HEADER_SIZE {
```
lauralt, line 604: I'm wondering if, for the time being (until we change to multiple `VolatileSlice`s), we should also check here if there's no next descriptor in the chain. If there is another descriptor in the chain, we keep the same behaviour as before, ignore the bytes after the header size in the first descriptor, and use only the second one. This kind of translates to: what faulty behaviour do we want to allow?

lauralt (approved):
Do we publish this as a patch release, in that case we should follow [this guide](https://github.com/rust-vmm/community/blob/main/docs/maintainers/crate_release.md#patch-release) which involves creating a new branch etc.

andreeaflorescu:
> Do we publish this as a patch release, in that case we should follow [this guide](https://github.com/rust-vmm/community/blob/main/docs/maintainers/crate_release.md#patch-release) which involves creating a new branch etc.

I don't think we need a patch release on a new branch unless we want to backport this fix to previous releases. We can just do patch releases from main as well as long as the history is linear, which is what I think is the case here.

crates/devices/virtio-vsock/src/packet.rs
stefano-garzarella: It makes sense to me in the sense of keeping the same behavior as before, but in my opinion we should not have ignored the bytes after the header in the first descriptor. In theory those are part of the payload.

So for now I agree to try to change as little as possible and check both the length and the not having a next. (So I can restore all the tests I changed as well).

In the future let's fix that and handle the various descriptors simply as a sequence of bytes.

What do you think?

lauralt: > It makes sense to me in the sense of keeping the same behavior as before, but in my opinion we should not have ignored the bytes after the header in the first descriptor. In theory those are part of the payload.
> 
> So for now I agree to try to change as little as possible and check both the length and the not having a next. (So I can restore all the tests I changed as well).
> 
> In the future let's fix that and handle the various descriptors simply as a sequence of bytes.
> 
> What do you think?

Yup, I agree. I don't think we can achieve that without the multiple VolatileSlices API change, so we should prioritise this refactoring.

lauralt: I will open a separate issue for that, I see your PR will close the one opened by you, which is okay I think.

stefano-garzarella:
new version:
- checked also that `chain_head.next()` is false as @lauralt suggested

crates/devices/virtio-vsock/src/packet.rs
```diff
@@ -460,22 +461,34 @@ impl<'a, B: BitmapSlice> VsockPacket<'a, B> {
...
+                mem.get_slice(
+                    chain_head
+                        .addr()
+                        .checked_add(PKT_HEADER_SIZE as u64)
+                        .ok_or(Error::DescriptorLengthTooSmall)?,
+                    pkt.len() as usize,
+                )
+                .map_err(Error::InvalidMemoryAccess)?
+            } else {
+                let data_desc = desc_chain.next().ok_or(Error::DescriptorChainTooShort)?;
```
lauralt, line 477: nit: I would check before is the next bit of the head is set (with has_next()).

lauralt (approved):

crates/devices/virtio-vsock/src/packet.rs
stefano-garzarella: Okay, I think it was pre-existing since I just moved the code in the else branch, but since I'm touching this code, make sense to update it ;-)

andreeaflorescu (approved):

PR #220, opened 2023-01-19
virtio-queue: fix alignment check for indirect descriptor

zhuangel:
see also: https://gitlab.com/virtio-fs/virtiofsd/-/issues/74

With specific guest kernel config, virtiofsd run panic with cloud-hypervisor, because address of indirect descriptor table is not proper aligned.

After read the spec, there should no restrict on address of indirect descriptor:

    2.6.5 The Virtqueue Descriptor Table
    ...
    To increase ring capacity the driver can store a table
    of indirect descriptors anywhere in memory, and insert
    a descriptor in main virtqueue (with
    flags&VIRTQ_DESC_F_INDIRECT on) that refers to memory
    buffer containing this indirect descriptor table.

### Summary of the PR

### Requirements

requirements:

  description line.
  test.
- [ ] All added/changed public-facing functionality has entries in the "Upcoming 

### DIFF
--- a/crates/virtio-queue/src/chain.rs
+++ b/crates/virtio-queue/src/chain.rs
@@ -104,10 +104,9 @@ where
             return Err(Error::InvalidIndirectDescriptor);
         }
 
-        // Check the target indirect descriptor table is correctly aligned.
-        if desc.addr().raw_value() & (VRING_DESC_ALIGN_SIZE as u64 - 1) != 0
-            || desc.len() & (VRING_DESC_ALIGN_SIZE - 1) != 0
-        {
+        // Alignment requirements for vring elements start from virtio 1.0,
+        // but this is not necessary for address of indirect descriptor.
+        if desc.len() & (VRING_DESC_ALIGN_SIZE - 1) != 0 {
             return Err(Error::InvalidIndirectDescriptorTable);
         }
 
@@ -386,24 +385,50 @@ mod tests {
     }
 
     #[test]
+    fn test_indirect_descriptor_address_noaligned() {
+        // Alignment requirements for vring elements start from virtio 1.0,
+        // but this is not necessary for address of indirect descriptor.
+        let m = &GuestMemoryMmap::<()>::from_ranges(&[(GuestAddress(0), 0x10000)]).unwrap();
+        let vq = MockSplitQueue::new(m, 16);
+        let dtable = vq.desc_table();
+
+        // Create a chain with a descriptor pointing to an indirect table with unaligned address.
+        let desc = Descriptor::new(
+            0x7001,
+            0x1000,
+            (VRING_DESC_F_INDIRECT | VRING_DESC_F_NEXT) as u16,
+            2,
+        );
+        dtable.store(0, desc).unwrap();
+
+        let mut c: DescriptorChain<&GuestMemoryMmap> = DescriptorChain::new(m, vq.start(), 16, 0);
+
+        // Create an indirect table with 4 chained descriptors.
+        let idtable = DescriptorTable::new(m, GuestAddress(0x7001), 4);
+        for i in 0..4u16 {
+            let desc: Descriptor = if i < 3 {
+                Descriptor::new(0x1000 * i as u64, 0x1000, VRING_DESC_F_NEXT as u16, i + 1)
+            } else {
+                Descriptor::new(0x1000 * i as u64, 0x1000, 0, 0)
+            };
+            idtable.store(i, desc).unwrap();
+        }
+
+        // Try to iterate through the indirect descriptor chain.
+        for i in 0..4 {
+            let desc = c.next().unwrap();
+            assert!(c.is_indirect);
+            if i < 3 {
+                assert_eq!(desc.flags(), VRING_DESC_F_NEXT as u16);
+                assert_eq!(desc.next(), i + 1);
+            }
+        }
+    }
+
+    #[test]
     fn test_indirect_descriptor_err() {
         // We are testing here different misconfigurations of the indirect table. For these error
         // case scenarios, the iterator over the descriptor chain won't return a new descriptor.
-        {
-            let m = &GuestMemoryMmap::<()>::from_ranges(&[(GuestAddress(0), 0x10000)]).unwrap();
-            let vq = MockSplitQueue::new(m, 16);
-
-            // Create a chain with a descriptor pointing to an invalid indirect table: addr not a
-            // multiple of descriptor size.
-            let desc = Descriptor::new(0x1001, 0x1000, VRING_DESC_F_INDIRECT as u16, 0);
-            vq.desc_table().store(0, desc).unwrap();
-
-            let mut c: DescriptorChain<&GuestMemoryMmap> =
-                DescriptorChain::new(m, vq.start(), 16, 0);
-
-            assert!(c.next().is_none());
-        }
-
         {
             let m = &GuestMemoryMmap::from_ranges(&[(GuestAddress(0), 0x10000)]).unwrap();
             let vq = MockSplitQueue::new(m, 16);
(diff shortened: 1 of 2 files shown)


crates/virtio-queue/src/chain.rs
```diff
@@ -105,9 +105,7 @@ where
         }
 
         // Check the target indirect descriptor table is correctly aligned.
-        if desc.addr().raw_value() & (VRING_DESC_ALIGN_SIZE as u64 - 1) != 0
-            || desc.len() & (VRING_DESC_ALIGN_SIZE - 1) != 0
-        {
+        if desc.len() & (VRING_DESC_ALIGN_SIZE - 1) != 0 {
```
stsquad, line 108: I note in the kernel the structures specifically point out this is pre-1.0 behaviour:

```
/* Alignment requirements for vring elements.
 * When using pre-virtio 1.0 layout, these fall out naturally.
 */
#define VRING_AVAIL_ALIGN_SIZE 2
#define VRING_USED_ALIGN_SIZE 4
#define VRING_DESC_ALIGN_SIZE 16
```

However we need to fix the comment as well because we don't check alignment - really all you are doing is checking the descriptor length is a multiple of VRING_DESC_ALIGN_SIZE. This is now a poorly named constant because we really just want to show we have an integer number of struct vring_descs.

stsquad (changes requested):

rbradford:
I see you're using a very old kernel (relative to virtio-fs) - 5.4. Perhaps it's just the old kernel that is responsible for this issue. Since the descriptors, etc, are provided by the kernel.

Can you retry with a modern kernel?

crates/virtio-queue/src/chain.rs
zhuangel: Thanks for reviewing, I have update the comments, could you please help to check it, thanks.

zhuangel:
> I see you're using a very old kernel (relative to virtio-fs) - 5.4. Perhaps it's just the old kernel that is responsible for this issue. Since the descriptors, etc, are provided by the kernel.
> 
> Can you retry with a modern kernel?

Thanks, @rbradford, I run the same test with guest kernel 5.19, the same problem still exists.

But for 6.0 and 6.1, I got a boot failure like `VmBoot(KernelMissingPvhHeader)`, I just change the SLUB into SLOB base recommend configuration [https://github.com/cloud-hypervisor/cloud-hypervisor/blob/v28.0/resources/linux-config-x86_64](url), and I checked `CONFIG_PVH` is enabled in config file.

rbradford:
> > I see you're using a very old kernel (relative to virtio-fs) - 5.4. Perhaps it's just the old kernel that is responsible for this issue. Since the descriptors, etc, are provided by the kernel.
> > Can you retry with a modern kernel?
> 
> Thanks, @rbradford, I run the same test with guest kernel 5.19, the same problem still exists.
> 
> But for 6.0 and 6.1, I got a boot failure like `VmBoot(KernelMissingPvhHeader)`, I just change the SLUB into SLOB base recommend configuration [https://github.com/cloud-hypervisor/cloud-hypervisor/blob/v28.0/resources/linux-config-x86_64](url), and I checked `CONFIG_PVH` is enabled in config file.

You might need to compiler newer kernels like this: ```CFLAGS="-Wa,-mx86-used-note=no" make bzImage -j `nproc` ```

zhuangel:
> > > I see you're using a very old kernel (relative to virtio-fs) - 5.4. Perhaps it's just the old kernel that is responsible for this issue. Since the descriptors, etc, are provided by the kernel.
> > > Can you retry with a modern kernel?
> > 
> > 
> > Thanks, @rbradford, I run the same test with guest kernel 5.19, the same problem still exists.
> > But for 6.0 and 6.1, I got a boot failure like `VmBoot(KernelMissingPvhHeader)`, I just change the SLUB into SLOB base recommend configuration [https://github.com/cloud-hypervisor/cloud-hypervisor/blob/v28.0/resources/linux-config-x86_64](url), and I checked `CONFIG_PVH` is enabled in config file.
> 
> You might need to compiler newer kernels like this: `` CFLAGS="-Wa,-mx86-used-note=no" make bzImage -j `nproc` ``

Thanks for suggestion, the problem still exists.

zhuangel:
> > > > I see you're using a very old kernel (relative to virtio-fs) - 5.4. Perhaps it's just the old kernel that is responsible for this issue. Since the descriptors, etc, are provided by the kernel.
> > > > Can you retry with a modern kernel?
> > > 
> > > 
> > > Thanks, @rbradford, I run the same test with guest kernel 5.19, the same problem still exists.
> > > But for 6.0 and 6.1, I got a boot failure like `VmBoot(KernelMissingPvhHeader)`, I just change the SLUB into SLOB base recommend configuration [https://github.com/cloud-hypervisor/cloud-hypervisor/blob/v28.0/resources/linux-config-x86_64](url), and I checked `CONFIG_PVH` is enabled in config file.
> > 
> > 
> > You might need to compiler newer kernels like this: `` CFLAGS="-Wa,-mx86-used-note=no" make bzImage -j `nproc` ``
> 
> Thanks for the suggestion, the problem still exists.

After hack the linux-loader code (force change the alignment of elf note section from 8 into 4), I could load v6.1 kernel successfully, I not sure the root cause of this issue, but this should have no relation with current PR, because I reproduce the same issue (virtiofsd panic), after change guest kernel into v6.1.

FYI, the hack code on linux-loader, [https://github.com/rust-vmm/linux-loader/blob/v0.6.0/src/loader/x86_64/elf/mod.rs#L346](url) [https://github.com/rust-vmm/linux-loader/blob/v0.6.0/src/loader/x86_64/elf/mod.rs#L347](url) [https://github.com/rust-vmm/linux-loader/blob/v0.6.0/src/loader/x86_64/elf/mod.rs#L381](url)

stsquad (dismissed):

crates/virtio-queue/src/chain.rs
```diff
@@ -389,21 +388,6 @@ mod tests {
     fn test_indirect_descriptor_err() {
         // We are testing here different misconfigurations of the indirect table. For these error
         // case scenarios, the iterator over the descriptor chain won't return a new descriptor.
-        {
```
lauralt, line 392: I think we should keep the test and assert that it is okay to have an address that is not aligned. We should also move the test somewhere else since this test function is for indirect descriptor errors.

crates/virtio-queue/src/chain.rs
```diff
@@ -104,10 +104,9 @@ where
             return Err(Error::InvalidIndirectDescriptor);
         }
 
-        // Check the target indirect descriptor table is correctly aligned.
-        if desc.addr().raw_value() & (VRING_DESC_ALIGN_SIZE as u64 - 1) != 0
-            || desc.len() & (VRING_DESC_ALIGN_SIZE - 1) != 0
-        {
+        // Alignment requirements for vring elements start from virtio 1.0,
```
lauralt, line 107: Can you also add an entry in the virtio-queue changelog for this? It's not an API change, but it's an important bug fix.

zhuangel: Sorry, I am not familiar with the release process, I check the commit history of CHANGELOG.md, the changelog will be updated when new release comes, this PR should be an issue fix for next release?

crates/virtio-queue/src/chain.rs
zhuangel: Thanks for reviewing, I add a new case for indirect descriptor with unaligned address.

crates/virtio-queue/src/chain.rs
lauralt: Yes, we can update the changelog either right before the release with all new changes, or incrementally in each PR. We are trying to go with the second option, that's why we added it as a check in the PR template. You should just add a Fixed section in the changelog and one entry for your fix. Thanks!

zhuangel: OK, I have added this change into the changelog.

lauralt (approved):

stsquad (approved):

PR #223, opened 2023-02-08
virtio-queue: fix endianess issues in mock queue

slp:
Fix a couple of new endianess issues in the mock queue that impact big endian hosts (namely, s390x).

Signed-off-by: Sergio Lopez <slp@redhat.com>

### DIFF
--- a/crates/virtio-queue/src/mock.rs
+++ b/crates/virtio-queue/src/mock.rs
@@ -464,7 +464,8 @@ impl<'a, M: GuestMemory> MockSplitQueue<'a, M> {
         let mut new_entries = 0;
         let avail_idx: u16 = self
             .mem
-            .read_obj(self.avail_addr().unchecked_add(2))
+            .read_obj::<u16>(self.avail_addr().unchecked_add(2))
+            .map(u16::from_le)
             .unwrap();
 
         for (idx, desc) in descs.iter().enumerate() {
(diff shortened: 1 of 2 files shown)


rbradford (commented):
So, I think that for virtio 1.0 the spec does specify little endian but for legacy/pre 1.0 it is native endian. Does this crate support legacy/transitional devices?

lauralt:
> So, I think that for virtio 1.0 the spec does specify little endian but for legacy/pre 1.0 it is native endian. Does this crate support legacy/transitional devices?

No, it doesn't, it is specified in the first note [here](https://github.com/rust-vmm/vm-virtio#note).

lauralt (approved):

rbradford (approved):
Great! I looked through the source code but missed it in the README!

andreeaflorescu:
I wonder if we could get the CI running on a big endian host so that we don't introduce these breaking changes all the time :( are there any easy hosts that we could set up?

slp:
> I wonder if we could get the CI running on a big endian host so that we don't introduce these breaking changes all the time :( are there any easy hosts that we could set up?

That would be great, but I'm not aware of any CI service providing s390x support :-(

andreeaflorescu:
Buildkite allows us to run the CI on any host, we just need a s390x host on which we can install the buildkite agent.

PR #226, opened 2023-02-20
release virtio-queue v0.7.1

andreeaflorescu:
### Summary of the PR

This is a patch release that fixes endianess and the problems with the iterator. In this release we did not update to the latest virtio-bindings version as that would result in a big change. This could cause other problems that would make it hard to consume this otherwise important patch release.

Note that when building locally virtio-queue the latests virtio-bindings will be used, but in production use cases this will not be the case because the v0.1.0 virtio-bindings will be used from crates.io.

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/crates/virtio-queue-ser/CHANGELOG.md
+++ b/crates/virtio-queue-ser/CHANGELOG.md
@@ -1,3 +1,7 @@
+# v0.4.1
+- Update the virtio-queue dependency to v0.7.1. This release contains no
+  breaking changes.
+
 # v0.4.0
 
 ## Added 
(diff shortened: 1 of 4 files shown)


rbradford (dismissed):

crates/virtio-queue-ser/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# v0.4.1
+- Update the virtio-queue dependency to v0.7.1. This has not breaking changes.
```
stsquad, line 2: s/This has not breaking changes./This has no API breaking changes/?

andreeaflorescu: Done.

andreeaflorescu:
@rbradford @stsquad please re-review.

stsquad (approved):

rbradford (approved):

PR #278, opened 2023-11-15
virtio_queue: Add descriptors_utils.rs

MatiasVara:
### Summary of the PR
This commit adds `descriptors_utils.rs` from virtiofsd (daa4e32). This module provides the Reader/Writer traits to iterate over descriptors. Devices shall rely on this interface instead of manipulate descriptors. The reason to this is that the distribution is driver-dependent, i.e., virtio does not force any particular way to distribute the data over descriptors. The version in this commit removes `read_to_at()` and `write_from_at()` functions that are required only for virtiofsd. This commit is on top of 0.9 tag. To try this module, I am adding support for it in the virtio-sound device at https://github.com/MatiasVara/vhost-device/tree/use-descriptor-utils-virtio-sound. 

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/virtio-queue/src/descriptor_utils.rs
+++ b/virtio-queue/src/descriptor_utils.rs
@@ -1,0 +1,879 @@
+// Portions Copyright 2019 The Chromium OS Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE-BSD-3-Clause file.
+//
+// Copyright (C) 2024 Red Hat, Inc. All rights reserved.
+//
+// SPDX-License-Identifier: Apache-2.0 AND BSD-3-Clause
+
+use std::collections::VecDeque;
+use std::io::{self, Read, Write};
+use std::mem::{size_of, MaybeUninit};
+use std::ops::Deref;
+use std::ptr::copy_nonoverlapping;
+use std::{cmp, result};
+
+use crate::{DescriptorChain, Error};
+use vm_memory::bitmap::{BitmapSlice, WithBitmapSlice};
+use vm_memory::{
+    Address, ByteValued, GuestMemory, GuestMemoryRegion, MemoryRegionAddress, VolatileSlice,
+};
+
+pub type Result<T> = result::Result<T, Error>;
+
+#[derive(Clone)]
+struct DescriptorChainConsumer<'a, B> {
+    buffers: VecDeque<VolatileSlice<'a, B>>,
+    bytes_consumed: usize,
+}
+
+impl<'a, B: BitmapSlice> DescriptorChainConsumer<'a, B> {
+    fn available_bytes(&self) -> usize {
+        // This is guaranteed not to overflow because the total length of the chain
+        // is checked during all creations of `DescriptorChainConsumer` (see
+        // `Reader::new()` and `Writer::new()`).
+        self.buffers
+            .iter()
+            .fold(0usize, |count, vs| count + vs.len())
+    }
+
+    fn bytes_consumed(&self) -> usize {
+        self.bytes_consumed
+    }
+
+    /// Consumes at most `count` bytes from the `DescriptorChain`. Callers must provide a function
+    /// that takes a `&[VolatileSlice]` and returns the total number of bytes consumed. This
+    /// function guarantees that the combined length of all the slices in the `&[VolatileSlice]` is
+    /// less than or equal to `count`.
+    ///
+    /// # Errors
+    ///
+    /// If the provided function returns any error then no bytes are consumed from the buffer and
+    /// the error is returned to the caller.
+    fn consume<F>(&mut self, count: usize, f: F) -> io::Result<usize>
+    where
+        F: FnOnce(&[&VolatileSlice<B>]) -> io::Result<usize>,
+    {
+        let mut buflen = 0;
+        let mut bufs = Vec::with_capacity(self.buffers.len());
+        for vs in &self.buffers {
+            if buflen >= count {
+                break;
+            }
+
+            bufs.push(vs);
+
+            let rem = count - buflen;
+            if rem < vs.len() {
+                buflen += rem;
+            } else {
+                buflen += vs.len();
+            }
+        }
+
+        if bufs.is_empty() {
+            return Ok(0);
+        }
+
+        let bytes_consumed = f(&bufs)?;
+
+        // This can happen if a driver tricks a device into reading/writing more data than
+        // fits in a `usize`.
+        let total_bytes_consumed =
+            self.bytes_consumed
+                .checked_add(bytes_consumed)
+                .ok_or_else(|| {
+                    io::Error::new(io::ErrorKind::InvalidData, Error::DescriptorChainOverflow)
+                })?;
+
+        let mut rem = bytes_consumed;
+        while let Some(vs) = self.buffers.pop_front() {
+            if rem < vs.len() {
+                // Split the slice and push the remainder back into the buffer list. Safe because we
+                // know that `rem` is not out of bounds due to the check and we checked the bounds
+                // on `vs` when we added it to the buffer list.
+                self.buffers.push_front(vs.offset(rem).unwrap());
+                break;
+            }
+
+            // No need for checked math because we know that `vs.size() <= rem`.
+            rem -= vs.len();
+        }
+
+        self.bytes_consumed = total_bytes_consumed;
+
+        Ok(bytes_consumed)
+    }
+
+    fn split_at(&mut self, offset: usize) -> Result<DescriptorChainConsumer<'a, B>> {
+        let mut rem = offset;
+        let pos = self.buffers.iter().position(|vs| {
+            if rem < vs.len() {
+                true
+            } else {
+                rem -= vs.len();
+                false
+            }
+        });
+
+        if let Some(at) = pos {
+            let mut other = self.buffers.split_off(at);
+
+            if rem > 0 {
+                // There must be at least one element in `other` because we checked
+                // its `size` value in the call to `position` above.
+                let front = other.pop_front().expect("empty VecDeque after split");
+                self.buffers
+                    .push_back(front.subslice(0, rem).map_err(Error::VolatileMemoryError)?);
+                other.push_front(front.offset(rem).map_err(Error::VolatileMemoryError)?);
+            }
+
+            Ok(DescriptorChainConsumer {
+                buffers: other,
+                bytes_consumed: 0,
+            })
+        } else if rem == 0 {
+            Ok(DescriptorChainConsumer {
+                buffers: VecDeque::new(),
+                bytes_consumed: 0,
+            })
+        } else {
+            Err(Error::SplitOutOfBounds(offset))
+        }
+    }
+}
+
+/// Provides high-level interface over the sequence of memory regions
+/// defined by readable descriptors in the descriptor chain.
+///
+/// Note that virtio spec requires driver to place any device-writable
+/// descriptors after any device-readable descriptors (2.6.4.2 in Virtio Spec v1.1).
+/// Reader will skip iterating over descriptor chain when first writable
+/// descriptor is encountered.
+#[derive(Clone)]
+pub struct Reader<'a, B = ()> {
+    buffer: DescriptorChainConsumer<'a, B>,
+}
+
+impl<'a, B: BitmapSlice> Reader<'a, B> {
+    /// Construct a new Reader wrapper over `desc_chain`.
+    pub fn new<M, T>(mem: &'a M, desc_chain: DescriptorChain<T>) -> Result<Reader<'a, B>>
+    where
+        M: GuestMemory,
+        <<M as GuestMemory>::R as GuestMemoryRegion>::B: WithBitmapSlice<'a, S = B>,
+        T: Deref,
+        T::Target: GuestMemory + Sized,
+    {
+        let mut total_len: usize = 0;
+        let buffers = desc_chain
+            .readable()
+            .map(|desc| {
+                // Verify that summing the descriptor sizes does not overflow.
+                // This can happen if a driver tricks a device into reading more data than
+                // fits in a `usize`.
+                total_len = total_len
+                    .checked_add(desc.len() as usize)
+                    .ok_or(Error::DescriptorChainOverflow)?;
+
+                let region = mem
+                    .find_region(desc.addr())
+                    .ok_or(Error::FindMemoryRegion)?;
+                let offset = desc
+                    .addr()
+                    .checked_sub(region.start_addr().raw_value())
+                    .unwrap();
+                region
+                    .get_slice(MemoryRegionAddress(offset.raw_value()), desc.len() as usize)
+                    .map_err(Error::GuestMemoryError)
+            })
+            .collect::<Result<VecDeque<VolatileSlice<'a, B>>>>()?;
+        Ok(Reader {
+            buffer: DescriptorChainConsumer {
+                buffers,
+                bytes_consumed: 0,
+            },
+        })
+    }
+
+    /// Reads an object from the descriptor chain buffer.
+    pub fn read_obj<T: ByteValued>(&mut self) -> io::Result<T> {
+        let mut obj = MaybeUninit::<T>::uninit();
+
+        // SAFETY: `MaybeUninit` guarantees that the pointer is valid for
+        // `size_of::<T>()` bytes.
+        let buf = unsafe {
+            ::std::slice::from_raw_parts_mut(obj.as_mut_ptr() as *mut u8, size_of::<T>())
+        };
+
+        self.read_exact(buf)?;
+
+        // SAFETY: any type that implements `ByteValued` can be considered initialized
+        // even if it is filled with random data.
+        Ok(unsafe { obj.assume_init() })
+    }
+
+    /// Returns number of bytes available for reading.  May return an error if the combined
+    /// lengths of all the buffers in the DescriptorChain would cause an integer overflow.
+    pub fn available_bytes(&self) -> usize {
+        self.buffer.available_bytes()
+    }
+
+    /// Returns number of bytes already read from the descriptor chain buffer.
+    pub fn bytes_read(&self) -> usize {
+        self.buffer.bytes_consumed()
+    }
+
+    /// Splits this `Reader` into two at the given offset in the `DescriptorChain` buffer.
+    /// After the split, `self` will be able to read up to `offset` bytes while the returned
+    /// `Reader` can read up to `available_bytes() - offset` bytes.  Returns an error if
+    /// `offset > self.available_bytes()`.
+    pub fn split_at(&mut self, offset: usize) -> Result<Reader<'a, B>> {
+        self.buffer.split_at(offset).map(|buffer| Reader { buffer })
+    }
+}
+
+impl<'a, B: BitmapSlice> io::Read for Reader<'a, B> {
+    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+        self.buffer.consume(buf.len(), |bufs| {
+            let mut rem = buf;
+            let mut total = 0;
+            for vs in bufs {
+                let copy_len = cmp::min(rem.len(), vs.len());
+
+                // SAFETY: Safe because we verify that we do not read outside
+                // of the slice's bound. The slice guard will only get dropped
+                // after the function returns. This will keep the pointer valid
+                // while reads are happening.
+                unsafe {
+                    copy_nonoverlapping(vs.ptr_guard().as_ptr(), rem.as_mut_ptr(), copy_len);
+                }
+                rem = &mut rem[copy_len..];
+                total += copy_len;
+            }
+            Ok(total)
+        })
+    }
+}
+
+/// Provides high-level interface over the sequence of memory regions
+/// defined by writable descriptors in the descriptor chain.
+///
+/// Note that virtio spec requires driver to place any device-writable
+/// descriptors after any device-readable descriptors (2.6.4.2 in Virtio Spec v1.1).
+/// Writer will start iterating the descriptors from the first writable one and will
+/// assume that all following descriptors are writable.
+#[derive(Clone)]
+pub struct Writer<'a, B = ()> {
+    buffer: DescriptorChainConsumer<'a, B>,
+}
+
+impl<'a, B: BitmapSlice> Writer<'a, B> {
+    /// Construct a new Writer wrapper over `desc_chain`.
+    pub fn new<M, T>(mem: &'a M, desc_chain: DescriptorChain<T>) -> Result<Writer<'a, B>>
+    where
+        M: GuestMemory,
+        <<M as GuestMemory>::R as GuestMemoryRegion>::B: WithBitmapSlice<'a, S = B>,
+        T: Deref,
+        T::Target: GuestMemory + Sized,
+    {
+        let mut total_len: usize = 0;
+        let buffers = desc_chain
+            .writable()
+            .map(|desc| {
+                // Verify that summing the descriptor sizes does not overflow.
+                // This can happen if a driver tricks a device into writing more data than
+                // fits in a `usize`.
+                total_len = total_len
+                    .checked_add(desc.len() as usize)
+                    .ok_or(Error::DescriptorChainOverflow)?;
+
+                let region = mem
+                    .find_region(desc.addr())
+                    .ok_or(Error::FindMemoryRegion)?;
+                let offset = desc
+                    .addr()
+                    .checked_sub(region.start_addr().raw_value())
+                    .unwrap();
+                region
+                    .get_slice(MemoryRegionAddress(offset.raw_value()), desc.len() as usize)
+                    .map_err(Error::GuestMemoryError)
+            })
+            .collect::<Result<VecDeque<VolatileSlice<'a, B>>>>()?;
+
+        Ok(Writer {
+            buffer: DescriptorChainConsumer {
+                buffers,
+                bytes_consumed: 0,
+            },
+        })
+    }
+
+    /// Writes an object to the descriptor chain buffer.
+    pub fn write_obj<T: ByteValued>(&mut self, val: T) -> io::Result<()> {
+        self.write_all(val.as_slice())
+    }
+
+    /// Returns number of bytes available for writing.  May return an error if the combined
+    /// lengths of all the buffers in the DescriptorChain would cause an overflow.
+    pub fn available_bytes(&self) -> usize {
+        self.buffer.available_bytes()
+    }
+
+    /// Returns number of bytes already written to the descriptor chain buffer.
+    pub fn bytes_written(&self) -> usize {
+        self.buffer.bytes_consumed()
+    }
+
+    /// Splits this `Writer` into two at the given offset in the `DescriptorChain` buffer.
+    /// After the split, `self` will be able to write up to `offset` bytes while the returned
+    /// `Writer` can write up to `available_bytes() - offset` bytes.  Returns an error if
+    /// `offset > self.available_bytes()`.
+    pub fn split_at(&mut self, offset: usize) -> Result<Writer<'a, B>> {
+        self.buffer.split_at(offset).map(|buffer| Writer { buffer })
+    }
+}
+
+impl<'a, B: BitmapSlice> io::Write for Writer<'a, B> {
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        self.buffer.consume(buf.len(), |bufs| {
+            let mut rem = buf;
+            let mut total = 0;
+            for vs in bufs {
+                let copy_len = cmp::min(rem.len(), vs.len());
+
+                // SAFETY: Safe because we ensure that we do not write over the
+                // slice's bounds. The slice guard will only get dropped after
+                // the function returns. This will keep the pointer valid while
+                // writes are happening.
+                unsafe {
+                    copy_nonoverlapping(rem.as_ptr(), vs.ptr_guard_mut().as_ptr(), copy_len);
+                }
+                vs.bitmap().mark_dirty(0, copy_len);
+                rem = &rem[copy_len..];
+                total += copy_len;
+            }
+            Ok(total)
+        })
+    }
+
+    fn flush(&mut self) -> io::Result<()> {
+        // Nothing to flush since the writes go straight into the buffer.
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::{Descriptor, Queue, QueueOwnedT, QueueT};
+    use vm_memory::{GuestAddress, GuestMemoryMmap, Le32};
+
+    use crate::mock::MockSplitQueue;
+    use virtio_bindings::bindings::virtio_ring::{VRING_DESC_F_NEXT, VRING_DESC_F_WRITE};
+
+    const MAX_QUEUE_SIZE: u16 = 16;
+
+    #[derive(Copy, Clone, PartialEq, Eq)]
+    pub enum DescriptorType {
+        Readable,
+        Writable,
+    }
+
+    /// Test utility function to create a descriptor chain in guest memory.
+    pub fn create_descriptor_chain(
+        memory: &GuestMemoryMmap,
+        descriptor_array_addr: GuestAddress,
+        descriptors: Vec<(DescriptorType, u32)>,
+        spaces_between_regions: u32,
+    ) -> Result<DescriptorChain<&GuestMemoryMmap>> {
+        let descriptors_len = descriptors.len();
+        let mut descs = vec![];
+
+        let queue = MockSplitQueue::create(memory, descriptor_array_addr, MAX_QUEUE_SIZE);
+
+        let mut buffers_start_addr = queue.end();
+
+        for (index, (type_, size)) in descriptors.into_iter().enumerate() {
+            let mut flags = 0;
+            if let DescriptorType::Writable = type_ {
+                flags |= VRING_DESC_F_WRITE;
+            }
+            if index + 1 < descriptors_len {
+                flags |= VRING_DESC_F_NEXT;
+            }
+
+            descs.push(Descriptor::new(
+                buffers_start_addr.raw_value(),
+                size,
+                flags as u16,
+                (index + 1) as u16,
+            ));
+
+            let offset = size + spaces_between_regions;
+            buffers_start_addr = buffers_start_addr
+                .checked_add(u64::from(offset))
+                .ok_or(Error::InvalidChain)?;
+        }
+
+        queue.build_desc_chain(&descs).unwrap();
+
+        let avail_ring = queue.avail_addr();
+
+        let mut queue: Queue = Queue::new(MAX_QUEUE_SIZE).unwrap();
+        queue
+            .try_set_desc_table_address(descriptor_array_addr)
+            .unwrap();
+        queue.try_set_avail_ring_address(avail_ring).unwrap();
+        queue.set_ready(true);
+
+        let chain = queue.iter(memory).unwrap().next().unwrap();
+
+        Ok(chain.clone())
+    }
+
+    #[test]
+    fn reader_test_inv_desc_addr() {
+        let memory: GuestMemoryMmap =
+            GuestMemoryMmap::from_ranges(&[(GuestAddress(0x0), 0x1000)]).unwrap();
+
+        let queue = MockSplitQueue::create(&memory, GuestAddress(0x0), MAX_QUEUE_SIZE);
+
+        // set addr out of memory
+        let descriptor = Descriptor::new(0x1001, 1, 0, 1_u16);
+        queue.build_desc_chain(&[descriptor]).unwrap();
+
+        let avail_ring = queue.avail_addr();
+
+        let mut queue: Queue = Queue::new(MAX_QUEUE_SIZE).unwrap();
+        queue.try_set_desc_table_address(GuestAddress(0x0)).unwrap();
+        queue.try_set_avail_ring_address(avail_ring).unwrap();
+        queue.set_ready(true);
+
+        let chain = queue.iter(&memory).unwrap().next().unwrap();
+
+        assert!(Reader::new(&memory, chain).is_err());
+    }
+
+    #[test]
+    fn reader_test_simple_chain() {
+        use DescriptorType::*;
+
+        let memory_start_addr = GuestAddress(0x0);
+        let memory = GuestMemoryMmap::from_ranges(&[(memory_start_addr, 0x10000)]).unwrap();
+
+        let chain = create_descriptor_chain(
+            &memory,
+            GuestAddress(0x0),
+            vec![
+                (Readable, 8),
+                (Readable, 16),
+                (Readable, 18),
+                (Readable, 64),
+            ],
+            0,
+        )
+        .expect("create_descriptor_chain failed");
+        let mut reader = Reader::new(&memory, chain).expect("failed to create Reader");
+        assert_eq!(reader.available_bytes(), 106);
+        assert_eq!(reader.bytes_read(), 0);
+
+        let mut buffer = [0_u8; 64];
+        if let Err(e) = reader.read_exact(&mut buffer) {
+            panic!("read_exact should not fail here: {:?}", e);
+        }
+
+        assert_eq!(reader.available_bytes(), 42);
+        assert_eq!(reader.bytes_read(), 64);
+
+        match reader.read(&mut buffer) {
+            Err(e) => panic!("read should not fail here: {:?}", e),
+            Ok(length) => assert_eq!(length, 42),
+        }
+
+        assert_eq!(reader.available_bytes(), 0);
+        assert_eq!(reader.bytes_read(), 106);
+    }
+
+    #[test]
+    fn writer_test_simple_chain() {
+        use DescriptorType::*;
+
+        let memory_start_addr = GuestAddress(0x0);
+        let memory = GuestMemoryMmap::from_ranges(&[(memory_start_addr, 0x10000)]).unwrap();
+
+        let chain = create_descriptor_chain(
+            &memory,
+            GuestAddress(0x0),
+            vec![
+                (Writable, 8),
+                (Writable, 16),
+                (Writable, 18),
+                (Writable, 64),
+            ],
+            0,
+        )
+        .expect("create_descriptor_chain failed");
+        let mut writer = Writer::new(&memory, chain).expect("failed to create Writer");
+        assert_eq!(writer.available_bytes(), 106);
+        assert_eq!(writer.bytes_written(), 0);
+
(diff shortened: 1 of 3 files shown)


MatiasVara:
@germag feel free to comment.

Ablu:
@MatiasVara thanks a lot for updating this! This is something which we really need. I will review this tomorrow in more detail!

Just a couple of minor comments before diving into a review:
- Would be great to add a reference from which virtiofsd version this was adapted
- Has someone done a comparison whether there have been new changes on this over on the crosvm project?

MatiasVara:
> @MatiasVara thanks a lot for updating this! This is something which we really need. I will review this tomorrow in more detail!
> 
> Just a couple of minor comments before diving into a review:
> 
> * Would be great to add a reference from which virtiofsd version this was adapted
> * Has someone done a comparison whether there have been new changes on this over on the crosvm project?

I missed those two points from our previous discussion. I will add a reference to the virtiofsd version. In regards to the second point, I just did a quick review of the crosvm version and found that it doesn't require a lifetime parameter when instantiated. However, I did not go any further. The differences may need to be better understood.

stsquad:
Is virtiofsd's read/write_at not likely to be useful for other backends? Is it a very special case only virtiofsd benefits from?

crates/virtio-queue/src/descriptor_utils.rs
```diff
@@ -0,0 +1,939 @@
...
+// Safe because it only has data and has no implicit padding.
+unsafe impl ByteValued for virtq_desc {}
+
+#[derive(Copy, Clone, Debug, Default)]
+#[repr(C)]
+struct virtq_avail {
+    flags: Le16,
+    idx: Le16,
+    ring: Le16,
+}
```
Ablu, line 423: virtio-bindings has definitions for these, but without the ByteValued impl... Do we maybe want to define these as transparent wrapper around the bindgen generated ones? :thinking:.

crates/virtio-queue/src/descriptor_utils.rs
```diff
@@ -0,0 +1,939 @@
...
+unsafe impl ByteValued for virtq_avail {}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::{Queue, QueueOwnedT, QueueT};
+    use vm_memory::{Bytes, GuestAddress};
+
+    const VIRTQ_DESC_F_NEXT: u16 = 0x1;
+    const VIRTQ_DESC_F_WRITE: u16 = 0x2;
```
Ablu, line 435: These should probably use the `VRING_DESC_F_*` definitions from virtio-bindings.

crates/virtio-queue/src/descriptor_utils.rs
```diff
@@ -0,0 +1,939 @@
...
+
+        let mut queue: Queue = Queue::new(MAX_QUEUE_SIZE).unwrap();
+        queue
+            .try_set_desc_table_address(descriptor_array_addr)
+            .unwrap();
+        queue.try_set_avail_ring_address(avail_ring).unwrap();
+        queue.set_ready(true);
+        let desc = queue.iter(memory).unwrap().next().unwrap();
+        Ok(desc.clone())
+    }
```
Ablu, line 497: This feels like it duplicates a lot of what is available at https://github.com/rust-vmm/vm-virtio/blob/d739c5e/crates/virtio-queue/src/mock.rs#L445 (also see the other mock functions there). Is there something that this does fundamentally different?

crates/virtio-queue/src/descriptor_utils.rs
```diff
@@ -0,0 +1,939 @@
...
+                (Readable, 64),
+            ],
+            0,
+        )
+        .expect("create_descriptor_chain failed");
+        let mut reader = Reader::new(&memory, chain).expect("failed to create Reader");
+        assert_eq!(reader.available_bytes(), 106);
+        assert_eq!(reader.bytes_read(), 0);
+
+        let mut buffer = [0 as u8; 64];
```
Ablu, line 523: nit: I do not think this case is required here. I think the read_exact should define this as byte just as well. Even if not, `0_u8` would probably be shorter.

Ablu (commented):
Some random comments, but nothing extraordinary. One thing I wonder about is whether the separation into Reader and Writer makes a lot of sense for the user.

Does a user of this really care that much about it? I wonder whether it could not be a single object and the API could be used something like:

```
for desc_chain in requests {
    let header = desc_chain.read_obj::<SomeHeader>()?;
    let response = ...;
    desc_chain.write_obj(response)?;
}
```
(maybe overly simplified, but should show the idea)

With the proposed API here that has to be done manually like `writter_status.split_at(size_of::<VirtioSoundHeader>())`. While this certainly bring great improvements for the user, I wonder whether the ergonomics could be even better with a single object? :thinking: Sure, some special case may require "seeking" in the chain, but I guess 90%+ of the uses will just be sequential processing/parsing followed by writing the answer?

MatiasVara:
> Is virtiofsd's read/write_at not likely to be useful for other backends? Is it a very special case only virtiofsd benefits from?

Probably, I can add them if they are required for other backends. If not, I believe each device can "adapt or extend" current traits to suit its particular needs.

MatiasVara:
@Ablu Thanks for the review! I will try to answer some of the comments during today. 

> Some random comments, but nothing extraordinary. One thing I wonder about is whether the separation into Reader and Writer makes a lot of sense for the user.
> 
> Does a user of this really care that much about it? I wonder whether it could not be a single object and the API could be used something like:
> 
> ```
> for desc_chain in requests {
>     let header = desc_chain.read_obj::<SomeHeader>()?;
>     let response = ...;
>     desc_chain.write_obj(response)?;
> }
> ```
> 
> (maybe overly simplified, but should show the idea)
> 
> With the proposed API here that has to be done manually like `writter_status.split_at(size_of::<VirtioSoundHeader>())`. While this certainly bring great improvements for the user, I wonder whether the ergonomics could be even better with a single object? 🤔 Sure, some special case may require "seeking" in the chain, but I guess 90%+ of the uses will just be sequential processing/parsing followed by writing the answer?

Having a single object would simplify things. My guess about this design is that the chain of descriptors could be divided into two areas: a chain with read-only descriptors followed by a chain with write-only descriptors. This is the only requirement from the virtio-spec about the descriptors distribution (2.6.4.2 in Virtio Spec v1.1). As far as I understand it, when a Reader and Writer are instantiated, they "point" to the beginning of the area they correspond to. A single object shall automatically take into account such an offset. This is not a problem at all, and such a logic should be added to `write/read_obj()`.

Ablu:
> Having a single object would simplify things. My guess about this design is that the chain of descriptors could be divided into two areas: a chain with read-only descriptors followed by a chain with write-only descriptors. This is the only requirement from the virtio-spec about the descriptors distribution (2.6.4.2 in Virtio Spec v1.1). As far as I understand it, when a Reader and Writer are instantiated, they "point" to the beginning of the area they correspond to. A single object shall automatically take into account such an offset. This is not a problem at all, and such a logic should be added to `write/read_obj()`.

Ah. Did not realize that readable first, writeable after that even was a global rule. Not sure whether we need specific logic to verify this (they flags on the descriptor sound sufficient to me). But it confirms that read, then write is the only case :).

@MatiasVara: Want to try to adapt it to a single object? Eventually, I think we should target for this API being the default way to work with descriptors. So probably the descriptor chain iterator on the queue should in the end just return something that implements this high-level API :).

MatiasVara:
> > Having a single object would simplify things. My guess about this design is that the chain of descriptors could be divided into two areas: a chain with read-only descriptors followed by a chain with write-only descriptors. This is the only requirement from the virtio-spec about the descriptors distribution (2.6.4.2 in Virtio Spec v1.1). As far as I understand it, when a Reader and Writer are instantiated, they "point" to the beginning of the area they correspond to. A single object shall automatically take into account such an offset. This is not a problem at all, and such a logic should be added to `write/read_obj()`.
> 
> Ah. Did not realize that readable first, writeable after that even was a global rule. Not sure whether we need specific logic to verify this (they flags on the descriptor sound sufficient to me). But it confirms that read, then write is the only case :).
> 
> @MatiasVara: Want to try to adapt it to a single object? Eventually, I think we should target for this API being the default way to work with descriptors. So probably the descriptor chain iterator on the queue should in the end just return something that implements this high-level API :).

Yes, I can try to adapt it to a single object.

MatiasVara:
> Some random comments, but nothing extraordinary. One thing I wonder about is whether the separation into Reader and Writer makes a lot of sense for the user.
> 
> Does a user of this really care that much about it? I wonder whether it could not be a single object and the API could be used something like:
> 
> ```
> for desc_chain in requests {
>     let header = desc_chain.read_obj::<SomeHeader>()?;
>     let response = ...;
>     desc_chain.write_obj(response)?;
> }
> ```
> 
> (maybe overly simplified, but should show the idea)
> 
> With the proposed API here that has to be done manually like `writter_status.split_at(size_of::<VirtioSoundHeader>())`. While this certainly bring great improvements for the user, I wonder whether the ergonomics could be even better with a single object? 🤔 Sure, some special case may require "seeking" in the chain, but I guess 90%+ of the uses will just be sequential processing/parsing followed by writing the answer?

Would it make sense an interface like this? Note that this is just a draft and names will be different:
```rust
pub struct Walker<'a>{
    r: Reader<'a>,
    w: Writer<'a>,
    payload: Writer<'a>
}

impl <'a> Walker<'a>{
    pub fn new<M>(
        mem: &'a GuestMemoryMmap,
        desc_chain: DescriptorChain<M>,
        offset: usize
    ) -> Result<Walker<'a>>
    where
       M: Deref,
       M: Clone,
       M::Target: GuestMemory + Sized,
    {
        let reader = Reader::new(mem, desc_chain.clone()).unwrap();
        let mut writer = Writer::new(mem, desc_chain.clone()).unwrap();
        let payload = writer.split_at(offset).unwrap();
        Ok(Walker {
            r: reader,
            w: writer,
            payload
        })
    }

    pub fn read_obj<T: ByteValued>(&mut self) -> io::Result<T> {
        self.r.read_obj()
    }
    
    pub fn write_obj<T: ByteValued>(&mut self, val: T) -> io::Result<()> {
        self.w.write_obj(val)
    }

    pub fn write_payload_obj<T: ByteValued>(&mut self, val: T) -> io::Result<()> {
        self.payload.write_obj(val)
    }
}
```
I do not think payload is required for all the devices. In those cases, I would set `offset = 0`. WDYT?

stefano-garzarella:
> I do not think payload is required for all the devices. In those cases, I would set `offset = 0`. WDYT?

What is the `payload` use case?

MatiasVara:
> > I do not think payload is required for all the devices. In those cases, I would set `offset = 0`. WDYT?
> 
> What is the `payload` use case?

I used to use the `split_at()` function on the writer to get a new writer that begins in the offset in which the payload of the request starts. In the case of virtio-sound, on write-only descriptors, there is first the status and then the payload but we first write the payload and then the status. I would require a sort of seek() to first write the payload and then the status. Since all request have a status, I proposed to `split_at` when Walker is instantiated and then provide a method to explicitly write in the area that corresponds with the payload which is after the status header.

Ablu:
I wonder: Why do we need separate Reader/Writer types at all? Couldn't we just use one that implements read/write mechanism directly (a bit like Cursor from the standard lib)?

MatiasVara:
> I wonder: Why do we need separate Reader/Writer types at all? Couldn't we just use one that implements read/write mechanism directly (a bit like Cursor from the standard lib)?

Can you elaborate?

stefano-garzarella:
> > > I do not think payload is required for all the devices. In those cases, I would set `offset = 0`. WDYT?
> > 
> > 
> > What is the `payload` use case?
> 
> I used to use the `split_at()` function on the writer to get a new writer that begins in the offset in which the payload of the request starts. In the case of virtio-sound, on write-only descriptors, there is first the status and then the payload but we first write the payload and then the status. I would require a sort of seek() to first write the payload and then the status (or just change the way we currently do it). Since all requests have a status, I proposed to `split_at` when Walker is instantiated and then provide a method to explicitly write in the area that corresponds with the payload which is after the status header.

mmm, `payload` it looks like too specific for virtio-sound use case, for example in virtio-blk is it the opposite (payload is before the status).
What about having a vector of Writers/Readers, then using an index?

Ablu:
> > I wonder: Why do we need separate Reader/Writer types at all? Couldn't we just use one that implements read/write mechanism directly (a bit like Cursor from the standard lib)?
> 
> Could you elaborate?, Thanks.

Sorry for not being super clear... What I meant was I see no pressing need for having `Writer` and `Reader` types for walking the descriptor chain.

Going back to what I sketched before:

```
for desc_chain in requests {
    let header = desc_chain.read_obj::<SomeHeader>()?;
    let response = ...;
    desc_chain.write_obj(response)?;
}
```

Here, descriptor_chain could directly be what this PR currently calls `DescriptorChainConsumer` (probably merged into or containing the current `desc_chain` type). Then read_obj()/write_obj could be defined on that while keeping some internal cursor.

I am not exactly sure whether I understand why the payload differentiation is needed though. Why can't I just do:

```
desc_chain.read_obj::<SomeHeader>()?;
desc_chain.write_obj(payload)?;
desc_chain.write_obj(status)?;
```
:thinking:

crates/virtio-queue/src/descriptor_utils.rs
```diff
@@ -0,0 +1,939 @@
...
+                let copy_len = cmp::min(rem.len(), vs.len());
+
+                // SAFETY: Safe because we verify that we do not read outside
+                // of the slice's bound. The slice guard will only get dropped
+                // after the function returns. This will keep the pointer valid
+                // while reads are happening.
+                unsafe {
+                    copy_nonoverlapping(vs.ptr_guard().as_ptr(), rem.as_mut_ptr(), copy_len);
+                }
+                rem = &mut rem[copy_len..];
```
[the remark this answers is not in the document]
roypat: We have some very similar APIs over at Firecracker (https://github.com/firecracker-microvm/firecracker/blob/main/src/vmm/src/devices/virtio/iovec.rs). We tried coming up with a way to integrate something like this into `{Read,Write}Volatile`, but also couldnt find anything nice (but those thoughts were also mostly around adding some `WriteVolatile::write_vectored_volatile(&[VolatileSlice])`, which would have to convert the `VolatileSlice`s to `iovec`s, and thus allocate some `Vec` somewhere). 

I guess a "non-contiguous VolatileSlice" might be the right abstract? E.g. instead VolatileSlice, which is essentially iovec + dirty bitmap, it would be Vec<iovec> + dirty bitmap. And if that was the core primitive of vm-memory, we could define `{Read,Write}Volatile` in terms of that, to get `readv` and `writev` for "free"

MatiasVara:
> I am not exactly sure whether I understand why the payload differentiation is needed though. Why can't I just do:
> 
> ```
> desc_chain.read_obj::<SomeHeader>()?;
> desc_chain.write_obj(payload)?;
> desc_chain.write_obj(status)?;
> ```
> 



> > > > I do not think payload is required for all the devices. In those cases, I would set `offset = 0`. WDYT?
> > > 
> > > 
> > > What is the `payload` use case?
> > 
> > 
> > I used to use the `split_at()` function on the writer to get a new writer that begins in the offset in which the payload of the request starts. In the case of virtio-sound, on write-only descriptors, there is first the status and then the payload but we first write the payload and then the status. I would require a sort of seek() to first write the payload and then the status (or just change the way we currently do it). Since all requests have a status, I proposed to `split_at` when Walker is instantiated and then provide a method to explicitly write in the area that corresponds with the payload which is after the status header.
> 
> mmm, `payload` it looks like too specific for virtio-sound use case, for example in virtio-blk is it the opposite (payload is before the status). What about having a vector of Writers/Readers, then using an index?

Could you elaborate? I proposed payload to work around `split_at()` which is not simply to implement if both Reader and Writer are in the same object. I think `split_at()` is useful because it allows you write to the write-only descriptor chain in a independent way.

stefano-garzarella:
> > mmm, `payload` it looks like too specific for virtio-sound use case, for example in virtio-blk is it the opposite (payload is before the status). What about having a vector of Writers/Readers, then using an index?
> 
> Could you elaborate? I proposed payload to work around `split_at()` which is not simply to implement if both Reader and Writer are in the same object.

Yeah, but this is meaningful only for virtio-sound, because for example for virito-blk the `payload` is before the `status`.

For some others maybe `payload` doesn't mean much, so IMHO we should provide more generic API. What I was suggesting was for example provide a vector of offsets to the constructor, where to split the Writers or Readers, then use the index in the API to write a specific section we split.

> I think `split_at()` is useful because it allows you write to the write-only descriptor chain in a independent way.

If we want a single Walker, then maybe we need to provide methods to split the Walker, but I don't think we should split into header and payload in it, because that's really device-specific and the user has to do the split.

MatiasVara:
> > > I wonder: Why do we need separate Reader/Writer types at all? Couldn't we just use one that implements read/write mechanism directly (a bit like Cursor from the standard lib)?
> > 
> > 
> > Could you elaborate?, Thanks.
> 
> Sorry for not being super clear... What I meant was I see no pressing need for having `Writer` and `Reader` types for walking the descriptor chain.
> 
> Going back to what I sketched before:
> 
> ```
> for desc_chain in requests {
>     let header = desc_chain.read_obj::<SomeHeader>()?;
>     let response = ...;
>     desc_chain.write_obj(response)?;
> }
> ```
> 
I think you can't use the same `desc_chain` for reading and writing but I may be wrong. I think a single object would still have two `desc_chain` one for reading and another for writing. In such a case, `split_at` may be hard to implement.  


> Here, descriptor_chain could directly be what this PR currently calls `DescriptorChainConsumer` (probably merged into or containing the current `desc_chain` type). Then read_obj()/write_obj could be defined on that while keeping some internal cursor.
> 
> I am not exactly sure whether I understand why the payload differentiation is needed though. Why can't I just do:
> 
> ```
> desc_chain.read_obj::<SomeHeader>()?;
> desc_chain.write_obj(payload)?;
> desc_chain.write_obj(status)?;
> ```
> 
> 🤔
I used payload to work around `split_at()`. The status is not always at the end of the chain. In that case, you would need a sort of seek before the `write_obj()`. I used `split_at()` instead so I can write the payload and the status independently.

Ablu:
> I used payload to work around `split_at()`. The status is not always at the end of the chain. In that case, you would need a sort of seek before the `write_obj()`. I used `split_at()` instead so I can write the payload and the status independently.

I wonder: Couldn't one just `.clone()` at the appropriate state in order to detach one reader/writer instance from another?

Ablu:
> I think it is still required to separate the `desc_chain` for reading and for writing.

Why is that? :thinking: Chances are that I misunderstand something, but I have not found the reason spelled out so far.

> I think we have to modify current implementation only if changes improve the way it is currently used. It is also true that those changes may raise after some utilization in the future. The other way to think would be to let each device add descriptor_utils.rs (as it is done for virtiofsd) and modify it as needed. In the future, we can move descriptor_utils.rs to virtio-queue crate when it is more clear how other devices use it. It is not clear to me how to modify Reader/Writer in a single object in a way that produces any benefits. I am not saying that that does not exist. A third option would be to provide something much more simpler, e.g., what we proposed for virtio-sound, in which some API simply allows reading and writing over descriptors in a continuous way so the device is allowed to support any distribution. This would require rewriting descriptor_utils from scratch, which is not the goal of this PR.    

I think unless we have some upstream solution, people probably won't put the brain cycles into creating a one-fits-all solutions that suits everyone equally well. I was hoping that one could turn it into a simpler to use API with just a few tweaks. I will try to play with this somewhen during this week and see how it feels.

Ablu:
> > > I think it is still required to separate the `desc_chain` for reading and for writing.
> > 
> > 
> > Why is that? 🤔 Chances are that I misunderstand something, but I have not found the reason spelled out so far.
> 
> I did not explain myself very well, I meant that Reader relies on `desc_chain.readable()` whereas Writer relies on `desc_chain.writable()`.

Ah! Got it now! I guess it makes sense with the read-first, write-then specification. In that case, would it maybe make sense to make those readable/writable calls return these reader/write objects? It would require to also do an iterator implementation on them, but it feels like a simpler API to me in the end?

I am slightly worried by forcing the user to manually create the reader / writers, they may not realize that this is possible and instead keep manually iterating things... Or am I too pessimistic here?

MatiasVara:
> I think unless we have some upstream solution, people probably won't put the brain cycles into creating a one-fits-all solutions that suits everyone equally well. I was hoping that one could turn it into a simpler to use API with just a few tweaks. I will try to play with this somewhen during this week and see how it feels.

I agree. The following code sketches what we did for virtio-sound although it was not merged. I am probably missing something here and also this could be refactored. In overall, the `read()` function is used to read in an continuous way:   

```rust
pub struct Request {
    pub desc_chain: DescriptorChain,
    pos_curr_desc: usize,
    idx_curr_desc: usize,
}

impl Request {
    pub fn new(len: usize, desc_chain: DescriptorChain) -> Self {
        Self {
            pos_curr_desc: 0,
            idx_curr_desc: 0,
            desc_chain: desc_chain,
        }
    }

    pub fn read(&mut self, buf: &mut [u8]) -> Result<u32> {
        let mut count = buf.len();
        let descriptors: Vec<_> = self.desc_chain.clone().collect();
        let mut buf_pos = 0;

        while count > 0 {
            if self.idx_curr_desc > descriptors.len()  {
                break;
            }
            let avail = descriptors[self.idx_curr_desc].len() - self.pos_curr_desc as u32;
            let mut end = buf.len() as u32;

            if avail < (buf.len() - buf_pos) as u32 {
                end = buf_pos as u32 + avail;
            }

            let len = self
                .desc_chain
                .memory()
                .read(
                    &mut buf[buf_pos..end as usize],
                    descriptors[self.idx_curr_desc]
                        .addr()
                        .checked_add(self.pos_curr_desc as u64)
                        .ok_or(Error::DescriptorReadFailed)?,
                )
                .map_err(|_| Error::DescriptorReadFailed)?;

            buf_pos += len;
            self.pos_curr_desc += len;
            count -= len;

            if self.pos_curr_desc >= descriptors[self.idx_curr_desc].len() as usize {
                self.idx_curr_desc += 1;
                self.pos_curr_desc = 0;
            }
        }
        Ok((buf.len() - count) as u32)
    }
}
```
Do you think this could be simpler than using Reader/Writer?

MatiasVara:
> I am slightly worried by forcing the user to manually create the reader / writers, they may not realize that this is possible and instead keep manually iterating things... Or am I too pessimistic here?

I think the idea is that user does not use Reader/Writer as iterators. The user will simple issue .read() and .write() from/to other objects like a header.

Ablu:
I will play with the API a bit later this week. Will respond then :)

Ablu:
> I will play with the API a bit later this week. Will respond then :)

I tried to iterate on it, but did not make as much progress as I hoped while fighting the GuestMemory abstraction layers... Will try to continue tomorrow or Monday...

crates/virtio-queue/src/descriptor_utils.rs
```diff
@@ -0,0 +1,939 @@
...
+/// descriptor is encountered.
+#[derive(Clone)]
+pub struct Reader<'a, B = ()> {
+    buffer: DescriptorChainConsumer<'a, B>,
+}
+
+impl<'a, B: Bitmap + BitmapSlice + 'static> Reader<'a, B> {
+    /// Construct a new Reader wrapper over `desc_chain`.
+    pub fn new<M>(
+        mem: &'a GuestMemoryMmap<B>,
```
Ablu, line 191: This should ideally just depend on the GuestMemory trait. Unfortunately, that becomes a bit messy with querying the regions later... I got some ideas to clean that up, but we probably need to do it step by step...

Ablu: Yep, but ideally, we could stay generic over the concrete GuestMemory implementation. So it should probably be a `GuestMemory` here (`GuestMemoryMmap` is one implementation of `GuestMemory`). However, it looks like the generics in vm-memory are a bit weird. I _think_ I found a way to clean that up. It mostly seems to be related to some weird higher-rank lifetime generic constraints. But I need a bit more time to figure out the best solution.

MatiasVara:
> Ok. I gave up on the generics for now... It needs some larger refactoring in vm-memory.
> 
> So back to the actual topic:
> 
> I played with the API a bit. The two things that annoyed me with the reader/writer separation is that code has to manually create the reader/writer and has to forward to the "cursor" to the writable section. Though, I wonder whether we can fix that with two fairly simple tweaks:
> 
> 1. adding a `.reader()` and `.writer()` function on `DescriptorChain` that returns instances of these helpers

Do you mean the Reader and Writer helpers from `descriptor_utils.rs`? Or the user would be able to use these functions over a `DescriptorChain` to just read and write over the chain? `write()` could already take into account the offset due to the read-only descriptors at the begin of the chain.

Ablu:
> Do you mean the Reader and Writer helpers from `descriptor_utils.rs`?

Yes, I was suggesting new `.reader()` and `.writer()` functions on the DescriptorChain that return the helpers added in this PR.

> write() could already take into account the offset due to the read-only descriptors at the begin of the chain.

My idea would be the just seek the descriptor chain forward (by simpler calling .next() until we have a writable descriptor) before calling the constructors of the `Writer`.

Does that make more sense?

MatiasVara:
> > Do you mean the Reader and Writer helpers from `descriptor_utils.rs`?
> 
> Yes, I was suggesting new `.reader()` and `.writer()` functions on the DescriptorChain that return the helpers added in this PR.
> 
> > write() could already take into account the offset due to the read-only descriptors at the begin of the chain.
> 
> My idea would be the just seek the descriptor chain forward (by simpler calling .next() until we have a writable descriptor) before calling the constructors of the `Writer`.
> 
> Does that make more sense?

Yes, I think you do not need to call .next(), the constructor of Writer already does it.

Ablu:
> Yes, it does, I think you do not need to call .next(), the constructor of Writer already does it.

ah. You are right. I misread the code and thought it yields None if there is a mismatch, but it will just keep iterating and only yields for matching write/read permissions. That looks good then. Not sure if the helper functions are required then...

Shall we try to port a couple more devices to see how it feels?

Ablu:
> yes, I think so. I would like to try to add the instantiation of Reader/Writer when a DescriptorChain is instantiated (I do not know if you already has this in some repo) and then play with `write()` and `read()` in virtio-sound.

I fear tying the reader/writer creation to the constructor of a DescriptorChain may be quite costly. It involves mapping all the buffers even if one then does not need the reader/writer. Also DescriptorChain iterates by yielding instances of itself. We probably do not want to create reader/writers during each iteration step. So I assume we want to keep it "on demand". But we should certainly make it as easy to use as possible.

MatiasVara:
> > yes, I think so. I would like to try to add the instantiation of Reader/Writer when a DescriptorChain is instantiated (I do not know if you already has this in some repo) and then play with `write()` and `read()` in virtio-sound.
> 
> I fear tying the reader/writer creation to the constructor of a DescriptorChain may be quite costly. It involves mapping all the buffers even if one then does not need the reader/writer. Also DescriptorChain iterates by yielding instances of itself. We probably do not want to create reader/writers during each iteration step. So I assume we want to keep it "on demand". But we should certainly make it as easy to use as possible.

Thanks, I think I misunderstood this sentence:
> adding a .reader() and .writer() function on DescriptorChain that returns instances of these helpers

I thought that `DescriptorChain` would instantiate the helpers and `.read()` and `.write()` would hide the details about the helpers. But you propose to return something like this in `.reader()`:
```rust
Reader::new(&mem, desc_chain.clone())
```

Ablu:
> I thought that `DescriptorChain` would instantiate the helpers and `.read()` and `.write()` would hide the details about the helpers. But you propose to return something like this in `.reader()`:
> 
> ```rust
> Reader::new(&mem, desc_chain.clone())
> ```

yep :)

crates/virtio-queue/src/chain.rs
```diff
@@ -88,6 +88,28 @@ where
         }
     }
 
+    /// Return a new instance of Writer
+    pub fn writer(self, mem: &'a GuestMemoryMmap) -> Result<Writer<'a>, Error>
+    where
+        M::Target: GuestMemory + Sized,
```
Ablu, line 94: Hm... Why is this needed here? M is not really used in this function, right?

crates/virtio-queue/src/chain.rs
```diff
@@ -32,7 +32,7 @@ pub struct DescriptorChain<M> {
     is_indirect: bool,
 }
 
-impl<M> DescriptorChain<M>
+impl<'a, M> DescriptorChain<M>
```
Ablu, line 35: I would move the lifetime down to the two functions that require it. Or does anything require it on this top-level?

crates/virtio-queue/src/chain.rs
```diff
@@ -88,6 +88,28 @@ where
...
+
+    /// Return a new instance of Reader
+    pub fn reader(self, mem: &'a GuestMemoryMmap) -> Result<Reader<'a>, Error>
+    where
+        M::Target: GuestMemory + Sized,
+    {
+        let Ok(reader) = Reader::new(&mem, self) else {
+            return Err(Error::InvalidChain);
+        };
+        Ok(reader)
```
Ablu, line 110: This could be a `.map_err()`.

crates/virtio-queue/src/descriptor_utils.rs
```diff
@@ -0,0 +1,939 @@
...
+#[derive(Debug)]
+pub enum Error {
+    DescriptorChainOverflow,
+    FindMemoryRegion,
+    GuestMemoryError(GuestMemoryError),
+    InvalidChain,
+    IoError(io::Error),
+    SplitOutOfBounds(usize),
+    VolatileMemoryError(VolatileMemoryError),
+}
```
Ablu, line 29: The rest of the crate seems to mostly use `crate::Error`. Can we reuse that here too (maybe by adding an error or two to the crate-wide Error enum?)

crates/virtio-queue/src/chain.rs
MatiasVara: I can reduce it to:
```rust
    where
        M::Target: Sized,
```
But when removing the M I get:
```
the size for values of type `<M as Deref>::Target` cannot be known at compilation time
the trait `Sized` is not implemented for `<M as Deref>::Target`
chain.rs(94, 26): required by a bound introduced by this call
descriptor_utils.rs(310, 34): required by a bound in `Writer::<'a, B>::new`
chain.rs(92, 79): consider further restricting the associated type: ` where <M as Deref>::Target: Sized`
```

crates/virtio-queue/src/chain.rs
MatiasVara: I'll move it to the two functions that require it.

crates/virtio-queue/src/chain.rs
MatiasVara: Do you mean something like:
```diff
-        let Ok(reader) = Reader::new(&mem, self) else {
-            return Err(Error::InvalidChain);
-        };
-        Ok(reader)
+        let reader = Reader::new(&mem, self).map_err(|_| Error::InvalidChain);
+        reader
```

Ablu: yes, though you can skip the assignment to the reader and just return the expression directly.

crates/virtio-queue/src/descriptor_utils.rs
MatiasVara: No, there is not. I could rewrite `create_descriptor_chain()` by using `build_desc_chain()`.

MatiasVara: I fixed the tests by using `build_desc_chain()`.

crates/virtio-queue/src/descriptor_utils.rs
MatiasVara: Do you mean that `writer_test_simple_chain()` test is not required?

Ablu: > I do not think this case is required here.

I think this was meant to mean: "cast" here... I was suggesting to drop the `as u8`. Sorry for the confusion...

crates/virtio-queue/src/descriptor_utils.rs
MatiasVara: I removed this duplication.

crates/virtio-queue/src/descriptor_utils.rs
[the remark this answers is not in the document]
MatiasVara: Could I address this after merging?

virtio-queue/src/descriptor_utils.rs
```diff
@@ -0,0 +1,850 @@
+// Copyright 2019 The Chromium OS Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
```
Ablu, line 3: We currently claim dual licensing of Apache + BSD. So this probably requires us to tweak the documentation and Cargo.toml if that is no longer universally true?

Ablu (commented):
Overal,, I think I am happy with this. Not sure how to handle the license difference here...

virtio-queue/src/descriptor_utils.rs
MatiasVara: Hello, how the documentation shall be tweaked? Do you think I need to add something to the PR?

Ablu: If the license of virtio-queue changed, it would at least have to be reflected in the Cargo.toml file. Probably the `OR` there has to become and `AND`?

MatiasVara:
Hello, what would be best approach to fix the CI regarding the coverage? The report states:
```
ValueError: Current code coverage (86.03%) deviates by -0.95% from the previous code coverage 86.98%.
```
Thanks.

Ablu:
> Hello, what would be best approach to fix the CI regarding the coverage? The report states:

We should attempt to figure out where it is coming from.

```
cargo llvm-cov --html
```
should be able to generate a html report where coverage is annotated line by line. Does that reveal any corner cases that would be useful to cover? Since this is a pure code addition, I assume some lines are simply not covered. If you think there is no reasonable test to fix the coverage then the percentage in the configuration file may be lowered (while making clear in the commit message why there is no point in fixing the coverage).

MatiasVara:
> cargo llvm-cov --html

Thanks, I spotted some functions and corner cases that were not covered. Some of them are:
* split() for writer
* writing more than its size
* flush()

I will add tests for them.

MatiasVara:
Hello, why `rust-vmm-fuzz` (bin "virtio_queue_ser") cannot be compiled?

virtio-queue/src/descriptor_utils.rs
MatiasVara: Is changing the `OR` to `AND` enough to fix the licensing? Is this change needed at all?

crates/virtio-queue/src/descriptor_utils.rs
MatiasVara: @Ablu, after discussing with @stefano-garzarella, if using GuestMemory is not possible, do you think it is correct to define these helpers in DescriptorChain?

Ablu: @MatiasVara do you mean the helpers to create the reader or the whole `descriptor_utils.rs` file?

MatiasVara: yes, i.e., the writer and reader which are defined in the context of DescriptorChain. Shall we remove them from there?

Ablu: > yes, i.e., the writer and reader which are defined in the context of DescriptorChain. Shall we remove them from there?

I got no opinion on that. I am fine with either. Will leave it to CODEOWNERs to comment.

stefano-garzarella: After a second look, I think it's fine for now (if the maintainers agree) to leave them in DescriptorChain.

virtio-queue/src/descriptor_utils.rs
MatiasVara: I reverted the changes and submit https://github.com/rust-vmm/vm-virtio/issues/283.

virtio-queue/src/descriptor_utils.rs
```diff
@@ -0,0 +1,880 @@
...
```
stefano-garzarella, line 3: Maybe we should follow the other files.

What about this changes?

```suggestion
// Portions Copyright 2019 The Chromium OS Authors. All rights reserved.
// Use of this source code is governed by a BSD-style license that can be
// found in the LICENSE-BSD-3-Clause file.
```

stefano-garzarella: Feel free to add also `// Copyright (C) 2024 Red Hat, Inc. All rights reserved.`

MatiasVara: Done, thanks.

stefano-garzarella (approved):

Ablu (approved):

virtio-queue/src/descriptor_utils.rs
```diff
@@ -0,0 +1,884 @@
...
+/// Note that virtio spec requires driver to place any device-writable
+/// descriptors after any device-readable descriptors (2.6.4.2 in Virtio Spec v1.1).
+/// Reader will skip iterating over descriptor chain when first writable
+/// descriptor is encountered.
+#[derive(Clone)]
+pub struct Reader<'a, B = ()> {
+    buffer: DescriptorChainConsumer<'a, B>,
+}
+
+impl<'a, B: Bitmap + BitmapSlice + 'static> Reader<'a, B> {
```
stefano-garzarella, line 159: why we need `'static` here?

virtio-queue/src/descriptor_utils.rs
```diff
@@ -0,0 +1,884 @@
...
+                    .checked_add(desc.len() as usize)
+                    .ok_or(Error::DescriptorChainOverflow)?;
+
+                let region = mem
+                    .find_region(desc.addr())
+                    .ok_or(Error::FindMemoryRegion)?;
+                let offset = desc
+                    .addr()
+                    .checked_sub(region.start_addr().raw_value())
+                    .unwrap();
```
stefano-garzarella, line 186: Is this safe or we should return an error?

stefano-garzarella:
I was trying to use `Reader` and `Writer` in https://github.com/rust-vmm/vm-virtio/blob/main/virtio-vsock/src/packet.rs, but I needed it to accept a generic `GuestMemory`, so I gave it a try and after struggling with it, I found a few things that seem to work:

https://github.com/stefano-garzarella/vm-virtio/commit/98497a00fb3b247bd9df70d76576a100cd3dec96

@Ablu @MatiasVara What do you think?

virtio-queue/src/descriptor_utils.rs
Ablu: Short answer: Because abstractions in vm-memory are broken.

Long answer: The slack discussion on this is currently still readable: https://rust-vmm.slack.com/archives/CFH3R8PFV/p1701337335710629.
Summary: vm-memory uses higher-ranked-traits to seemingly cheat out of needing to leak lifetimes into the GuestMemory trait. But in reality this basically just causes everything needing 'static lifetime. I tried to fix it when @MatiasVara posted this PR, but the refactoring attempt exploded and started leaking lifetimes into a lot of public interfaces: https://github.com/Ablu/vm-memory/commit/b22bdbdc953778b1693de6c95044beda4293d05c.

Ablu:
> @Ablu @MatiasVara What do you think?

Ah, I tried the same before but missed being able to set the lifetime via `WithBitmapSlice`. That certainly fixes the problems that I had where region.find_region() would result in something tied to a lifetime to the region (and thus not being able to add it to the Writer instance).

So it looks good to me. Yet, this interface in vm-memory is just way too complex to use without loosing sanity. I also dislike the Deref use in DescriptorChain. It disallows properly using lifetimes through it. IMHO it should just be a `&GuestMemory` there. But I failed refactoring this in sensible steps as everything is so intertwined.

virtio-queue/src/descriptor_utils.rs
stefano-garzarella: @Ablu I see, but for example we can remove `Bitmap` since it is implicit in `BitmapSlice`, no?

BTW in the patch I posted I left only BitmapSlice and the compiler seems happy, I need to investigate that more.

stefano-garzarella:
> > @Ablu @MatiasVara What do you think?
> 
> Ah, I tried the same before but missed being able to set the lifetime via `WithBitmapSlice`. That certainly fixes the problems that I had where region.find_region() would result in something tied to a lifetime to the region (and thus not being able to add it to the Writer instance).

Yeah, I was inspired by virtio-vsock API: https://github.com/rust-vmm/vm-virtio/blob/d219302978ddaee67bd341433e8ac219da0302b8/virtio-vsock/src/packet.rs#L418

BTW, in the end I think I'll change the virtio-vsock API to accept Reader and Writer from the caller, so I'm not sure if I really need the changes I proposed. But if you think they can improve a bit this PR, feel free to incorporate.

> 
> So it looks good to me. Yet, this interface in vm-memory is just way too complex to use without loosing sanity. I also dislike the Deref use in DescriptorChain. It disallows properly using lifetimes through it. IMHO it should just be a `&GuestMemory` there. But I failed refactoring this in sensible steps as everything is so intertwined

Okay, so I'm not the only one who almost went crazy with it....

virtio-queue/src/descriptor_utils.rs
stefano-garzarella: > @Ablu I see, but for example we can remove `Bitmap` since it is implicit in `BitmapSlice`, no?
> 
> BTW in the patch I posted I left only BitmapSlice and the compiler seems happy, I need to investigate that more.

Okay, I can do that in my patch because of setting the lifetime via WithBitmapSlice. I'm resolving this conversation.

Ablu: Yes, BitmapSlice is just a "special" Bitmap version. So we can drop the `Bitmap`.

virtio-queue/src/descriptor_utils.rs
MatiasVara: I think it is safe because in the precedent line, we ensure that `desc.addr()` is in `region`, i.e., `desc.addr()` is >= `region.start_addr()`.

MatiasVara:
> I was trying to use `Reader` and `Writer` in https://github.com/rust-vmm/vm-virtio/blob/main/virtio-vsock/src/packet.rs, but I needed it to accept a generic `GuestMemory`, so I gave it a try and after struggling with it, I found a few things that seem to work:
> 
> [stefano-garzarella@98497a0](https://github.com/stefano-garzarella/vm-virtio/commit/98497a00fb3b247bd9df70d76576a100cd3dec96)
> 
> @Ablu @MatiasVara What do you think?

I think is it is a good idea. I just pushed a version with the changes.

stefano-garzarella (approved):

slp (approved):
LGTM, thanks!

stefano-garzarella:
I'm going to merge this, since we want to use it in vhost-devices.
We have 3 acks, so it should be okay.

PR #291, opened 2024-03-15
virtio-device: extend VirtioDeviceActions Trait

joaopeixoto13:
### Summary of the PR

This PR enhances the VirtioDeviceActions Trait to accommodate Vhost and Vhost-User devices. This way, this crate can be used not only to implement virtio abstractions for devices that are implemented inside the VMM, but it can also serve as a frontend virtio device abstraction for Vhost and Vhost-user type devices. This patch introduces four new methods to the VirtioDeviceActions:

- `read_config` and `write_config`: These methods are invoked when the driver intends to read from or write to the device configuration space. Given that the device configuration space can be managed by various handlers outside of the VMM, such as vhost or vhost-user, dedicated logic is necessary to handle these operations.
- `negotiate_driver_features`:  This method is called when the driver finishes the negotiation of the driver features with the frontend device (selecting page 0). This method is crucial when the device handler is implemented outside of the VMM since the frontend device needs to negotiate the features with the backend device. Otherwise, the device will not be prepared to support some driver features.
- `interrupt_status`: When the driver requires reading the interrupt status from the device, this method is invoked. Since the responsibility for managing interrupt status lies with the frontend device, specialized logic is needed to update the interrupt status appropriately (Used Buffer Notification or Configuration Change Notification). If the device is implemented within the VMM, the interrupt status is directly managed and updated by the device.

### Requirements

requirements:

  test.


joaopeixoto13:
Update: I already have a raw [implementation](https://github.com/joaopeixoto13/bao-virtio) that allows me to verify the [PR](https://github.com/rust-vmm/vm-virtio/pull/291) to use the VirtioDeviceActions trait to accommodate not only VirtIO backends but also Vhost and Vhost-user backends. In this case, it was tested and proven to work for: VirIO-Net and VirtIO-Block, Vhost-user-fs and Vhost-vsock.

virtio-device/src/virtio_config.rs
```diff
@@ -150,41 +163,19 @@ where
...
-            return;
-        }
-
-        // TODO: Are partial reads ok?
-        let end = cmp::min(offset.saturating_add(data.len()), config_len);
-        let read_len = end - offset;
-        // Cannot fail because the lengths are identical and we do bounds checking beforehand.
-        data[..read_len].copy_from_slice(&config_space[offset..end])
+        <Self as VirtioDeviceActions>::read_config(self, offset, data)
     }
```
epilys, line 175: Now `read_config` is delegated to `VirtioDeviceActions`, but that means `T` must implement it now, whereas it did it not have to before. Or did I miss something? It seems like you're taking away functionality here, not adding.

joaopeixoto13: @epilys You are right. I have already placed the default implementations within the VirtioDeviceActions trait. Only the devices that need to invoke some additional logic must implement it. For example, for the vhost-user devices and when the driver intends to read from or write to the device configuration space, if the protocol feature CONFIG is negotiated, dedicated logic is necessary to handle these operations (e.g. GET_CONFIG/SET_CONFIG requests).

virtio-device/src/virtio_config.rs
```diff
@@ -130,6 +182,10 @@ where
             1 => ((features << 32) >> 32) + (v << 32),
             // Accessing an unknown page has no effect.
             _ => features,
+        };
+
+        if page == 0 {
```
stefano-garzarella, line 187: Why only on page 0?

Is it fine if for some strange reason the device negotiates only features on page 1 to not call `negotiate_driver_features()`?

joaopeixoto13: > Why only on page 0?
> 
> Is it fine if for some strange reason the device negotiates only features on page 1 to not call `negotiate_driver_features()`?


In accordance with Linux, the frontend driver negotiates the device features recognized and enabled by the driver by first selecting page 1, performing a write operation, and then selecting page 0 to perform a subsequent write (check the `vm_finalize_features` function).

However, if for some unexpected reason, the driver only negotiates features on page 1, how can we detect this scenario? Would it be feasible to negotiate the driver features and protocol features (e.g., vhost-user) with the backend device twice?

It seems more logical to negotiate the features only once, ensuring that all driver features have been written and are fully prepared for negotiation (by verifying page 0). 

Additionally, we could delegate the check for page 0 to the VMM itself by invoking the `driver_features_select` method to determine which page is currently selected. However, it seems more efficient to integrate this logic directly into the `negotiate_driver_features` function.

stefano-garzarella: > > Why only on page 0?
> > Is it fine if for some strange reason the device negotiates only features on page 1 to not call `negotiate_driver_features()`?
> 
> In accordance with Linux, the frontend driver negotiates the device features recognized and enabled by the driver by first selecting page 1, performing a write operation, and then selecting page 0 to perform a subsequent write (check the `vm_finalize_features` function).

Is it defined by the spec, or it is an implementation detail?

> 
> However, if for some unexpected reason, the driver only negotiates features on page 1, how can we detect this scenario? 

If the spec defined the behaviour you mentioned, it is fine. Just add a comment on the code with that.

>Would it be feasible to negotiate the driver features and protocol features (e.g., vhost-user) with the backend device twice?

Maybe it's supported, but not ideally.

> 
> It seems more logical to negotiate the features only once, ensuring that all driver features have been written and are fully prepared for negotiation (by verifying page 0).

Yeah, I see.

> 
> Additionally, we could delegate the check for page 0 to the VMM itself by invoking the `driver_features_select` method to determine which page is currently selected. However, it seems more efficient to integrate this logic directly into the `negotiate_driver_features` function.

Agree.

joaopeixoto13: > Is it defined by the spec, or it is an implementation detail?

According to the [VirtIO specification](https://docs.oasis-open.org/virtio/virtio/v1.3/csd01/virtio-v1.3-csd01.html), I was unable to find specific guidance regarding how the driver should negotiate the features it supports. However, upon reviewing other VirtIO driver implementations, such as the [VirtIO guest drivers in Rust](https://crates.io/crates/virtio-drivers), I discovered that they implement the negotiation process in the reverse order compared to Linux. Specifically, they first select page 0 to address the lower 32 bits of features before moving to the higher 32 bits on page 1, as seen [here](https://github.com/rcore-os/virtio-drivers/blob/master/src/transport/mmio.rs#L338).
However, in [Linux](https://github.com/torvalds/linux/blob/master/drivers/virtio/virtio_mmio.c#L124), this happens in reverse order.
So, this indicates that the negotiation order can be implementation-specific.

A potential solution could be to verify if the feature negotiation is complete by checking the `FEATURES_OK` bit in the `set_device_status` method. 
If the bit is set, we can proceed to invoke the hook-up. 
This approach would provide abstraction from the guest side, allowing it to write to either page 0 or page 1, while ensuring that the hook-up is implemented only after all features have been successfully negotiated between the frontend and backend. 
What are your thoughts on this approach? If you agree, I can do that.

stefano-garzarella: > > Is it defined by the spec, or it is an implementation detail?
> 
> According to the [VirtIO specification](https://docs.oasis-open.org/virtio/virtio/v1.3/csd01/virtio-v1.3-csd01.html), I was unable to find specific guidance regarding how the driver should negotiate the features it supports. However, upon reviewing other VirtIO driver implementations, such as the [VirtIO guest drivers in Rust](https://crates.io/crates/virtio-drivers), I discovered that they implement the negotiation process in reverse order compared to [Linux](https://github.com/torvalds/linux/blob/master/drivers/virtio/virtio_mmio.c#L124). Specifically, they first select page 0 to address the lower 32 bits of features before moving to the higher 32 bits on page 1, as seen [here](https://github.com/rcore-os/virtio-drivers/blob/master/src/transport/mmio.rs#L338). So, this indicates that the negotiation order can be implementation-specific.

That was exactly my concern!

> 
> A potential solution could be to verify if the feature negotiation is complete by checking the `FEATURES_OK` bit in the `set_device_status` method. If the bit is set, we can proceed to invoke the hook-up. This approach would provide abstraction from the guest side, allowing it to write to either page 0 or page 1, while ensuring that the hook-up is implemented only after all features have been successfully negotiated between the frontend and backend. What are your thoughts on this approach? If you agree, I can do that.

Yeah, I completely agree!

joaopeixoto13: > > A potential solution could be to verify if the feature negotiation is complete by checking the `FEATURES_OK` bit in the `set_device_status` method. If the bit is set, we can proceed to invoke the hook-up. This approach would provide abstraction from the guest side, allowing it to write to either page 0 or page 1, while ensuring that the hook-up is implemented only after all features have been successfully negotiated between the frontend and backend. What are your thoughts on this approach? If you agree, I can do that.
> 
> Yeah, I completely agree!

Done [here](https://github.com/rust-vmm/vm-virtio/pull/291/commits/39eeae40e89870872786841977b2731d049343bb)

stefano-garzarella: > > > A potential solution could be to verify if the feature negotiation is complete by checking the `FEATURES_OK` bit in the `set_device_status` method. If the bit is set, we can proceed to invoke the hook-up. This approach would provide abstraction from the guest side, allowing it to write to either page 0 or page 1, while ensuring that the hook-up is implemented only after all features have been successfully negotiated between the frontend and backend. What are your thoughts on this approach? If you agree, I can do that.
> > 
> > 
> > Yeah, I completely agree!
> 
> Done [here](https://github.com/rust-vmm/vm-virtio/pull/291/commits/39eeae40e89870872786841977b2731d049343bb)

Cool, LGTM!
I think you can merge that change with the first commit and I'll approve it ;-)

joaopeixoto13: > > > > A potential solution could be to verify if the feature negotiation is complete by checking the `FEATURES_OK` bit in the `set_device_status` method. If the bit is set, we can proceed to invoke the hook-up. This approach would provide abstraction from the guest side, allowing it to write to either page 0 or page 1, while ensuring that the hook-up is implemented only after all features have been successfully negotiated between the frontend and backend. What are your thoughts on this approach? If you agree, I can do that.
> > > 
> > > 
> > > Yeah, I completely agree!
> > 
> > 
> > Done [here](https://github.com/rust-vmm/vm-virtio/pull/291/commits/39eeae40e89870872786841977b2731d049343bb)
> 
> Cool, LGTM! I think you can merge that change with the first commit and I'll approve it ;-)

Done :)

stefano-garzarella (approved):
LGTM, please rebase on current main.

joaopeixoto13:
> LGTM, please rebase on current main.

Sorry, I thought I was in the latest version. I believe everything is now in order.

stefano-garzarella:
@joaopeixoto13 are you still interested in getting this PR merged? If yes, could you please re-open it in https://github.com/rust-vmm/rust-vmm/. We moved `vm-virtio` crates there.

PR #297, opened 2024-06-19
use virtio-queue from crates.io to prevent dupes in Cargo.lock

aghecenco:
### Summary of the PR

Use the published virtio-queue to prevent unwanted duplicates in Cargo.lock when consuming both am unpublished device and a released virtio-queue.

### Requirements

requirements:

  test.


epilys:
@aghecenco looking at the CI failures, and it seems `virtio-queue` contains unpublished changes (we bump the version when we make a new release). So unfortunately this might not be possible as long as the crates are developed in the same repository.

Can you give more info re: Cargo.lock and duplicates? perhaps there's another solution to your problem.

PR #323, opened 2024-12-17
New round of clippy warning fixups 

epilys:
Fixes for new minor lint warnings that cause error in CI.

Signed-off-by: Manos Pitsidianakis <manos.pitsidianakis@linaro.org>

### DIFF
--- a/virtio-blk/src/request.rs
+++ b/virtio-blk/src/request.rs
@@ -275,6 +275,7 @@ mod tests {
     }
 
     impl Request {
+        /// Manually create mock `Request` instances.
         pub fn new(
             request_type: RequestType,
             data: Vec<(GuestAddress, u32)>,
--- a/virtio-queue/src/descriptor_utils.rs
+++ b/virtio-queue/src/descriptor_utils.rs
@@ -232,7 +232,7 @@ impl<'a, B: BitmapSlice> Reader<'a, B> {
     }
 }
 
-impl<'a, B: BitmapSlice> io::Read for Reader<'a, B> {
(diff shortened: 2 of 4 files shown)


virtio-blk/src/request.rs
```diff
@@ -130,6 +130,22 @@ pub struct Request {
 unsafe impl ByteValued for RequestHeader {}
 
 impl Request {
+    /// Manually create mock `Request` instances.
+    #[cfg(test)]
+    pub fn new(
```
stefano-garzarella, line 135: What warning are we fixing with this patch?

For the future I suggest reporting the warning in the commit message, it's helpful to understand how to improve future code as well.

epilys: The move was not lint-related, the method triggered the `missing_docs` lint and I noticed the method wasn't called in the `requests.rs` `tests` module, but from other files. Should I move it back or mention this in the commit message, you think?

stefano-garzarella: I would leave it in the test, I'm not familiar with this code, but maybe the idea was to instantiate it that way only in the test and not expose it to the user. It's true that it's conditional on the test here, but maybe it can be misinterpreted and accidentally enabled in the future.

In summary, I would make as little change as possible for now :-)

epilys: Will do, thanks! :)

epilys: Done

stefano-garzarella (approved):

PR #324, opened 2025-01-15
Add proof for conformance to 2.7.7.2 section

MatiasVara:
### Summary of the PR
This is WIP in which I try to add a kany proof to meet the requirements outlined in 2.7.7.2 of the virtio specification regarding the notification suppression mechanism. I sketched this proof from the same proof defined for the queue implemented in firecraker. This commit adds the verify_spec_2_7_7_2() proof to verify that the implementation of queue meets 2.7.7.2 requirement. The proof relies on whether the EVENT_IDX feature has been negotiated. Conversely with `test_needs_notification()` test, this proof `tests` for all possible values of queue. To run the proof, you can run:

```bash
cargo kani
```
The proof currently passes:
```
SUMMARY:
 ** 0 of 4802 failed (280 unreachable)

 ** 1 of 1 cover properties satisfied


VERIFICATION:- SUCCESSFUL
Verification Time: 41.5208s

Complete - 1 successfully verified harnesses, 0 failures, 1 total.
```

If you have any comment, feel free to let me know.

Thanks.

### Requirements

requirements:

  test.

### DIFF
--- a/virtio-queue/src/queue.rs
+++ b/virtio-queue/src/queue.rs
@@ -269,6 +269,259 @@ impl Queue {
     }
 }
 
+#[cfg(kani)]
+#[allow(dead_code)]
+mod verification {
+    use std::mem::ManuallyDrop;
+    use vm_memory::MmapRegion;
+
+    use std::num::Wrapping;
+    use vm_memory::FileOffset;
+
+    use vm_memory::guest_memory::GuestMemoryIterator;
+    use vm_memory::{GuestMemoryRegion, MemoryRegionAddress};
+
+    use super::*;
+
+    /// A made-for-kani version of `vm_memory::GuestMemoryMmap`. Unlike the real
+    /// `GuestMemoryMmap`, which manages a list of regions and then does a binary
+    /// search to determine which region a specific read or write request goes to,
+    /// this only uses a single region. Eliminating this binary search significantly
+    /// speeds up all queue proofs, because it eliminates the only loop contained herein,
+    /// meaning we can use `kani::unwind(0)` instead of `kani::unwind(2)`. Functionally,
+    /// it works identically to `GuestMemoryMmap` with only a single contained region.
+    pub struct ProofGuestMemory {
+        the_region: vm_memory::GuestRegionMmap,
+    }
+
+    impl<'a> GuestMemoryIterator<'a, vm_memory::GuestRegionMmap> for ProofGuestMemory {
+        type Iter = std::iter::Once<&'a vm_memory::GuestRegionMmap>;
+    }
+
+    impl GuestMemory for ProofGuestMemory {
+        type R = vm_memory::GuestRegionMmap;
+        type I = Self;
+
+        fn num_regions(&self) -> usize {
+            1
+        }
+
+        fn find_region(&self, addr: GuestAddress) -> Option<&Self::R> {
+            self.the_region
+                .to_region_addr(addr)
+                .map(|_| &self.the_region)
+        }
+
+        fn iter(&self) -> <Self::I as GuestMemoryIterator<Self::R>>::Iter {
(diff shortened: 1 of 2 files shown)


yinkaolotin:
@MatiasVara i have gone through the VIRTIO spec, but while i understand how the queue is managed, i was trying out a simple proof for add_used the other day; I expected this to work but didn't 
```
    #[kani::proof]
    #[kani::unwind(0)]
    fn verify_add_used() {
        let ProofContext(mut queue, mem) = kani::any();
        let used_idx = queue.next_used;
        let used_desc_table_index = kani::any();
        let len = kani::any();
        
         match queue.add_used(&mem, used_desc_table_index, len) {
            Ok(_) => {
                assert_eq!(queue.next_used, used_idx + Wrapping(1));
            }
            Err(e) => match e {
                Error::InvalidDescriptorIndex => {
                    assert!(used_desc_table_index >= queue.size);
                }
                _ => {
                   // i thought this should be true in case failure was due to AddressOverFLow or GuestMemory error
                    assert!(used_desc_table_index < queue.size);
                }
            },
        }
      }
```

the firecracker version looks like this:
```
if queue.add_used(used_desc_table_index, kani::any()).is_ok() {
            assert_eq!(queue.next_used, used_idx + Wrapping(1));
} else {
            assert_eq!(queue.next_used, used_idx);

            // Ideally, here we would want to actually read the relevant values from memory and
            // assert they are unchanged. However, kani will run out of memory if we try to do so,
            // so we instead verify the following "proxy property": If an error happened, then
            // it happened at the very beginning of add_used, meaning no memory accesses were
            // done. This is relying on implementation details of add_used, namely that
            // the check for out-of-bounds descriptor index happens at the very beginning of the
            // function.
            assert!(used_desc_table_index >= queue.actual_size());
}
```
 which i think wouldn't work due to how `add_used` is implemented in rust-vmm. In rust-vmm, if add used fails, it appears next_used could still have been updated in between the first failure point and the other, so that was why i checked for just `assert!(used_desc_table_index < queue.size);` which would be guaranteed, incase it's not a descriptor index error.
```
    if head_index >= self.size {
            error!(
                "attempted to add out of bounds descriptor to used ring: {}",
                head_index
            );
            return Err(Error::InvalidDescriptorIndex);
        }

        let next_used_index = u64::from(self.next_used.0 % self.size);
        // This can not overflow an u64 since it is working with relatively small numbers compared
        // to u64::MAX.
        let offset = VIRTQ_USED_RING_HEADER_SIZE + next_used_index * VIRTQ_USED_ELEMENT_SIZE;
        let addr = self
            .used_ring
            .checked_add(offset)
            .ok_or(Error::AddressOverflow)?;
        mem.write_obj(VirtqUsedElem::new(head_index.into(), len), addr)
            .map_err(Error::GuestMemory)?;

        self.next_used += Wrapping(1);
        self.num_added += Wrapping(1);

        mem.store(
            u16::to_le(self.next_used.0),
            self.used_ring
                .checked_add(2)
                .ok_or(Error::AddressOverflow)?,
            Ordering::Release,
        )
```

I assume i might be missing an important detail, but not sure what that might be

virtio-queue/src/queue.rs
```diff
@@ -271,6 +271,259 @@ impl Queue {
...
+    impl kani::Arbitrary for ProofContext {
+        fn any() -> Self {
+            let mem = setup_kani_guest_memory();
+
+            let mut queue = Queue::new(MAX_QUEUE_SIZE).unwrap();
+
+            queue.ready = true;
+
+            queue.set_desc_table_address(
+                Some(QUEUE_BASE_ADDRESS as u32),
```
MatiasVara, line 456: I think we need to use kani::any() to set the address of the different rings.

PR #327, opened 2025-01-29
virtio-bindings: use bindgen library from build.rs

stefanha:
Since commit 3e51e2d ("virtio-bindings: regenerate with bindgen 0.70.1") there are compile-time alignment checks in virtio-bindings code. These checks are specific to the architecture where bindgen was run and they can break builds on other architectures.

Here is an example i686 build failure after upgrading from virtio-bindings 0.2.2 to 0.2.4:
https://koji.fedoraproject.org/koji/taskinfo?taskID=127997322

The bindgen User Guide recommends generating bindings from build.rs because it avoids portability issues and the need to maintain a copy of the bindings for each target:
https://rust-lang.github.io/rust-bindgen/library-usage.html

Introduce an import-linux-headers.sh script that copies required virtio headers from a Linux headers tree into include/. Do not distribute the full Linux headers tree because it is large and has falls under various software licenses that are not compatible with this crate's license.

Generate the actual bindings at compile-time in build.rs and then include!() the output from src/lib.rs.

You can verify that the generated bindings have not changed significantly for x86_64 by diffing them against the previous commit. For example:
```diff
--- orig/virtio_gpu.rs
+++ new/virtio_gpu.rs
@@ -1,4 +1,4 @@
-/* automatically generated by rust-bindgen 0.70.1 */ +/* automatically generated by rust-bindgen 0.71.1 */

 #[repr(C)]
 #[derive(Default)]
@@ -58,7 +58,7 @@
 pub const VIRTIO_GPU_MAP_CACHE_WC: u32 = 3;
 pub type __u8 = ::std::os::raw::c_uchar;
 pub type __u32 = ::std::os::raw::c_uint;
-pub type __u64 = ::std::os::raw::c_ulonglong;
+pub type __u64 = ::std::os::raw::c_ulong;
 pub type __le32 = __u32;
 pub type __le64 = __u64;
 pub const virtio_gpu_ctrl_type_VIRTIO_GPU_UNDEFINED: virtio_gpu_ctrl_type = 0
 ```

The __u64 change is due to how <linux/types.h> defined __u64. We should avoid using that non-BSD licensed header. The result is equivalent now that the bindings are generated at compile-time.

Fixes: #326

### CODE REVIEW

File: virtio-bindings/import-linux-headers.sh [hunk 1]
```diff
@@ -0,0 +1,33 @@
...
+import_header() {
+	# Use our local header files rather than system headers
+	sed -e 's%#include <\([^>]*\)>%#include "\1"%' "$1" >"$2"
+}
+
+mkdir -p "$dst"
+
+# If you change this list, remember to update build.rs
+for header in \
+	virtio_blk \
```
stefano-garzarella, line 22: What about importing also version.h?
So we can easily figure out which linux version these headers belong to. (I also suggest writing it in the commit, I guess they are v6.12 as before, but I couldn't find references).

I mean just this change:

```suggestion
	version \
	virtio_blk \
```

stefanha: <linux/version.h> does not have a license header indicating it can be distributed under the BSD license.

Instead I updated the commit description to mention explicitly that the Linux 6.12 headers were used. Past updates to kernel headers were accompanied with a commit message that includes the kernel version. Let's keep doing that.

stefano-garzarella: @stefanha I see, makes sense to me.

BTW not sure if it was done by mistake, but in your last version I see `virtio-bindings/include/linux/version.h`

stefanha: Yes, it was a mistake. I had already implemented your idea when I noticed the licensing situation is not clear and forgot to remove version.h. I have pushed a fixed version now.

stefano-garzarella (approved):

virtio-bindings/build.rs
```diff
@@ -0,0 +1,39 @@
+// Copyright Red Hat, Inc.
+// SPDX-License-Identifier: (BSD-3-Clause OR Apache-2.0)
+
+use std::env;
+use std::path::PathBuf;
+
+fn generate_bindings(header: &str) {
```
rbradford, line 7: I think generating the bindings on crate build is imposing quite a heavy requirement on the host as bindgen has clang dependencies that are not easily resolvable. We probably didn't see it here since the rust-vmm container has a working combination of clang toolchain and bindgen tool but we're seeing that when we try to compile using GH actions using the "regular" rust action script.

PR #346, opened 2025-06-18
unit proof for add_used operation

priyasiddharth:
### Summary of the PR

feat(kani): unit proof for add_used method (sec 2.8.22)

Verify that the add_used method updates the queue on success and does not update on failure.
For this (and subsequent proofs), implement a stub region that supports stateful proofs.
We add a unit proof to make sure that the stub region itself is working as intended.
Note that `unsafe` code is added only for StubRegion that will run in Kani.
Kani verifies all `unsafe` accesses do not cause undefined behaviour (in the context of unit proof execution).

### Requirements

requirements:

- [X] All commits in this PR have Signed-Off-By trailers (with
- [X] All added/changed functionality has a corresponding unit/integration
  test.
- [X] All added/changed public-facing functionality has entries in the "Upcoming 
- [X] Any newly added `unsafe` code is properly documented.

### DIFF
--- a/virtio-queue/src/queue/verification.rs
+++ b/virtio-queue/src/queue/verification.rs
@@ -2,9 +2,290 @@
 use std::mem::ManuallyDrop;
 use std::num::Wrapping;
 
-use vm_memory::{FileOffset, GuestMemoryRegion, MemoryRegionAddress, MmapRegion};
+use vm_memory::{AtomicAccess, GuestMemoryError, GuestMemoryRegion, MemoryRegionAddress};
+
+use std::io::{Read, Write};
+use std::mem::MaybeUninit;
+use vm_memory::ByteValued;
 
 use super::*;
+
+pub struct StubRegion {
+    buffer: *mut u8,
+    region_len: usize,
+    region_start: GuestAddress,
+}
+
+// A stub region that implements the GuestMemoryRegion and Bytes traits. This is
+// used to create a single memory region for testing purposes in Kani. It allows
+// us to simulate a memory region without needing a full memory management
+// implementation. The region is backed by a raw pointer to a buffer, which is
+// allocated and leaked to ensure a stable address. The region supports reading
+// and writing bytes, objects, and slices. It also provides methods to convert
+// between guest addresses and memory region addresses, and to check offsets
+// within the region.
+impl StubRegion {
+    pub fn new(buf_ptr: *mut u8, buf_len: usize, start_offset: u64) -> Self {
+        Self {
+            buffer: buf_ptr,
+            region_len: buf_len,
+            region_start: GuestAddress(start_offset),
+        }
+    }
+
+    fn to_region_addr(&self, addr: GuestAddress) -> Option<MemoryRegionAddress> {
+        let offset = addr
+            .raw_value()
+            .checked_sub(self.region_start.raw_value())?;
+        if offset < self.region_len as u64 {
+            Some(MemoryRegionAddress(offset))
+        } else {
+            None
+        }
+    }
+
+    fn checked_offset(
+        &self,
+        addr: MemoryRegionAddress,
+        count: usize,
+    ) -> Option<MemoryRegionAddress> {
+        let end = addr.0.checked_add(count as u64)?;
+        if end <= self.region_len as u64 {
+            Some(MemoryRegionAddress(end))
+        } else {
+            None
+        }
+    }
+}
+
+impl GuestMemoryRegion for StubRegion {
+    type B = ();
+
+    fn len(&self) -> <GuestAddress as vm_memory::AddressValue>::V {
+        self.region_len.try_into().unwrap()
+    }
+
+    fn start_addr(&self) -> GuestAddress {
+        self.region_start
+    }
+
+    fn bitmap(&self) -> &Self::B {
+        // For Kani, we do not need a bitmap, so we return an empty tuple.
+        &()
+    }
+}
+
+// Implements the Bytes trait for StubRegion, allowing it to read and write
+// bytes and objects. This is used to simulate memory operations in Kani proofs.
+impl Bytes<MemoryRegionAddress> for StubRegion {
+    type E = GuestMemoryError;
+
+    fn write(&self, buf: &[u8], addr: MemoryRegionAddress) -> Result<usize, Self::E> {
+        let offset = addr.0 as usize;
+        let end = offset
+            .checked_add(buf.len())
+            .ok_or(GuestMemoryError::InvalidGuestAddress(GuestAddress(addr.0)))?;
+        if end > self.region_len as usize {
+            return Err(GuestMemoryError::InvalidGuestAddress(GuestAddress(addr.0)));
+        }
+        unsafe {
+            std::ptr::copy_nonoverlapping(buf.as_ptr(), self.buffer.add(offset), buf.len());
+        }
+        Ok(buf.len())
+    }
+
+    fn read(&self, buf: &mut [u8], addr: MemoryRegionAddress) -> Result<usize, Self::E> {
+        let offset = addr.0 as usize;
+        let end = offset
+            .checked_add(buf.len())
+            .ok_or(GuestMemoryError::InvalidGuestAddress(GuestAddress(addr.0)))?;
+        if end > self.region_len as usize {
+            return Err(GuestMemoryError::InvalidGuestAddress(GuestAddress(addr.0)));
+        }
+        unsafe {
+            std::ptr::copy_nonoverlapping(self.buffer.add(offset), buf.as_mut_ptr(), buf.len());
+        }
+        Ok(buf.len())
+    }
+
+    fn write_slice(&self, buf: &[u8], addr: MemoryRegionAddress) -> Result<(), Self::E> {
+        self.write(buf, addr).map(|_| ())
+    }
+
+    fn read_slice(&self, buf: &mut [u8], addr: MemoryRegionAddress) -> Result<(), Self::E> {
+        self.read(buf, addr).map(|_| ())
+    }
+
+    fn read_from<F: Read>(
+        &self,
+        addr: MemoryRegionAddress,
+        src: &mut F,
+        count: usize,
+    ) -> Result<usize, Self::E> {
+        let mut temp = vec![0u8; count];
+        src.read_exact(&mut temp)
+            .map_err(|_| GuestMemoryError::PartialBuffer {
+                expected: count,
+                completed: 0,
+            })?;
+        self.write(&temp, addr)
+    }
+
+    fn read_exact_from<F: Read>(
+        &self,
+        addr: MemoryRegionAddress,
+        src: &mut F,
+        count: usize,
+    ) -> Result<(), Self::E> {
+        let mut temp = vec![0u8; count];
+        src.read_exact(&mut temp)
+            .map_err(|_| GuestMemoryError::PartialBuffer {
+                expected: count,
+                completed: 0,
+            })?;
+        self.write_slice(&temp, addr)
+    }
+
+    fn read_obj<T: ByteValued>(&self, addr: MemoryRegionAddress) -> Result<T, Self::E> {
+        let size = std::mem::size_of::<T>();
+        let offset = addr.0 as usize;
+        let end = offset
+            .checked_add(size)
+            .ok_or(GuestMemoryError::InvalidGuestAddress(GuestAddress(addr.0)))?;
+        if end > self.region_len as usize {
+            return Err(GuestMemoryError::InvalidGuestAddress(GuestAddress(addr.0)));
+        }
+        let mut result: T = unsafe { MaybeUninit::<T>::zeroed().assume_init() };
+        unsafe {
+            std::ptr::copy_nonoverlapping(
+                self.buffer.add(offset),
+                (&mut result as *mut T) as *mut u8,
+                size,
+            );
+        }
+        Ok(result)
+    }
+
+    fn write_to<F: Write>(
+        &self,
+        addr: MemoryRegionAddress,
+        dst: &mut F,
+        count: usize,
+    ) -> Result<usize, Self::E> {
+        let offset = addr.0 as usize;
+        let end = offset
+            .checked_add(count)
+            .ok_or(GuestMemoryError::InvalidGuestAddress(GuestAddress(addr.0)))?;
+        if end > self.region_len as usize {
+            return Err(GuestMemoryError::InvalidGuestAddress(GuestAddress(addr.0)));
+        }
+        unsafe {
+            let slice = std::slice::from_raw_parts(self.buffer.add(offset), count);
+            dst.write_all(slice)
+                .map_err(|_| GuestMemoryError::PartialBuffer {
+                    expected: count,
+                    completed: 0,
+                })?;
+        }
+        Ok(count)
+    }
+
+    fn write_obj<T: ByteValued>(&self, val: T, addr: MemoryRegionAddress) -> Result<(), Self::E> {
+        let size = std::mem::size_of::<T>();
+        let offset = addr.0 as usize;
+        let end = offset
+            .checked_add(size)
+            .ok_or(GuestMemoryError::InvalidGuestAddress(GuestAddress(addr.0)))?;
+        if end > self.region_len as usize {
+            return Err(GuestMemoryError::InvalidGuestAddress(GuestAddress(addr.0)));
+        }
+        let bytes = val.as_slice();
+        unsafe {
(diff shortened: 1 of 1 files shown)


priyasiddharth:
@roypat Hi Patrick,
I am looking for some guidance on how best to fake `GuestMemory` read/writes for `kani`. The PR contains a detailed description of the problem. Hoping you can help.

thanks
Siddharth

virtio-queue/src/queue.rs
```diff
@@ -269,6 +269,314 @@ impl Queue {
...
+        _used_ring: GuestAddress,
+        _next_used: u16,) -> Result<(), Error> {
+        Ok(())
+    }
+
+    #[kani::proof]
+    #[kani::unwind(0)]
+    #[kani::stub(write_elem, stub_write_elem)]
+    #[kani::stub(store_elem, stub_store_elem)]
+    fn verify_add_used() {
```
priyasiddharth, line 559: This is the proof

virtio-queue/src/queue.rs
```diff
@@ -466,20 +774,12 @@ impl QueueT for Queue {
             .used_ring
             .checked_add(offset)
             .ok_or(Error::AddressOverflow)?;
-        mem.write_obj(VirtqUsedElem::new(head_index.into(), len), addr)
-            .map_err(Error::GuestMemory)?;
+        write_elem(mem, addr, head_index, len)?;
```
priyasiddharth, line 777: delegating to local functions that can be stubbed by kani

virtio-queue/src/queue.rs
```diff
@@ -466,20 +774,12 @@ impl QueueT for Queue {
...
 
-        mem.store(
-            u16::to_le(self.next_used.0),
-            self.used_ring
-                .checked_add(2)
-                .ok_or(Error::AddressOverflow)?,
-            Ordering::Release,
-        )
-        .map_err(Error::GuestMemory)
+        store_elem(mem, self.used_ring, self.next_used.0)
```
priyasiddharth, line 782: ditto

virtio-queue/src/queue.rs
priyasiddharth: There can be multiple problems:
1. State space can explode
2. There can be implicit loops (depending on semantics of copy) ... since we are writing a many-byte buffer .. and in the proof we say don't unroll/unroll once so all bytes may not be written, or other problems

Likely Flow:
The write_obj method delegates to write_slice :https://sourcegraph.com/github.com/rust-vmm/vm-memory/-/blob/src/mmap/mod.rs?L190-192
This copies bytes from src to dst: https://sourcegraph.com/github.com/rust-vmm/vm-memory/-/blob/src/bytes.rs?L538
Both problems can occur with this code.
See the commit from firecracker: https://sourcegraph.com/github.com/firecracker-microvm/firecracker/-/commit/1e680c64ae6a6f86f207d3b4e7be3a72fb95fc76
In firecracker it looks like problem 2 does not occur since we still have loops in the fake implementation.

virtio-queue/src/queue.rs
```diff
@@ -269,6 +269,314 @@ impl Queue {
...
+    /// this only uses a single region. Eliminating this binary search significantly
+    /// speeds up all queue proofs, because it eliminates the only loop contained herein,
+    /// meaning we can use `kani::unwind(0)` instead of `kani::unwind(2)`. Functionally,
+    /// it works identically to `GuestMemoryMmap` with only a single contained region.
+    pub struct ProofGuestMemory {
+        the_region: vm_memory::GuestRegionMmap,
+    }
+
+    impl GuestMemory for ProofGuestMemory {
+        type R = vm_memory::GuestRegionMmap;
```
MatiasVara, line 319: I replaced this with something like `type R = MyOptimizedRegion;` and then this leaded me to implement `Bytes<MemoryRegionAddress>` and `GuestMemoryRegion` for `MyOptimizedRegion`. I also struggled to implemented `new()`, which looks like:
```rust
    impl MyOptimizedRegion {
        pub fn new(start: GuestAddress, len: u64) -> Self {
            // Allocate the memory region and bitmap
            Self {
                region_start: start,
                region_len: len,
                data: vec![0; len as usize],
                region_bitmap: AtomicBitmap::new(len as usize, NonZero::new(4096).unwrap()),
            }
        }
    }
```
I am worried that this approach would add a lot of boiler plate code and thus code duplication. Any comments @roypat ? Thanks

roypat: Can you just use `()` as the bitmap type? That'll turn all the bitmap stuff into noops, pretty sure that's what we're doing over in Firecracker. For the memory allocation, try just doing `std::alloc::alloc` (or kani::any_array() with mem::forget). No need to ever free memory, because the concept of a memory leak doesnt exist in kani

virtio-queue/src/queue/verification.rs
```diff
@@ -269,3 +504,205 @@ fn verify_spec_2_7_7_2() {
...
+    let used_desc_table_index = kani::any();
+    if queue
+        .add_used(&memory, used_desc_table_index, kani::any())
+        .is_ok()
+    {
+        assert_eq!(queue.next_used, used_idx + Wrapping(1));
+    } else {
+        assert_eq!(queue.next_used, used_idx);
+
+        // Ideally, here we would want to actually read the relevant values from memory and
```
MatiasVara, line 523: I think this comment is true in firecraker but is it still valid in the rust-vmm implementation?

virtio-queue/src/queue/verification.rs
```diff
@@ -2,10 +2,286 @@
...
+    let region = StubRegion::new(buffer.as_mut_ptr(), buffer.len(), 0);
+
+    // Arbitrary bytes to write
+    let bytes: [u8; 16] = kani::any();
+    // Write bytes to region at offset 0
+    assert!(region.write(&bytes, MemoryRegionAddress(0)).is_ok());
+
+    // Read back into a new buffer
+    let mut readback = kani::vec::exact_vec::<u8, 16>();
+    assert!(region.read(&mut readback, MemoryRegionAddress(0)).is_ok());
```
MatiasVara, line 277: I wonder if the offset can also use `kani::any()?`

virtio-queue/src/queue/verification.rs
```diff
@@ -269,3 +502,57 @@ fn verify_spec_2_7_7_2() {
...
+/// Reads the idx field from the used ring in guest memory, as stored by Queue::add_used.
+/// Returns the value as u16.
+fn get_used_idx(
+    queue: &Queue,
+    mem: &SingleRegionGuestMemory,
+) -> Result<u16, vm_memory::GuestMemoryError> {
+    let addr =
+        queue
+            .used_ring
+            .checked_add(2)
```
stefano-garzarella, line 515: why 2?

virtio-queue/src/queue/verification.rs
```diff
@@ -2,10 +2,286 @@
...
+// used to create a single memory region for testing purposes in Kani. It allows
+// us to simulate a memory region without needing a full memory management
+// implementation. The region is backed by a raw pointer to a buffer, which is
+// allocated and leaked to ensure a stable address. The region supports reading
+// and writing bytes, objects, and slices. It also provides methods to convert
+// between guest addresses and memory region addresses, and to check offsets
+// within the region.
+impl StubRegion {
+    pub fn new(buf_ptr: *mut u8, buf_len: usize, start_offset: u64) -> Self {
+        // Allocate the buffer and leak it to get a stable pointer.
```
stefano-garzarella, line 29: Is this comment updated?

virtio-queue/src/queue/verification.rs
```diff
@@ -221,6 +453,7 @@ impl kani::Arbitrary for ProofContext {
     }
 }
 
+/// Kani proof harness for VirtIO 1.2 Section 2.7.7.2: Device-to-driver notification suppression.
```
MatiasVara, line 456: Since this is for a different proof we could just add this comment in a new PR.

virtio-queue/src/queue/verification.rs
```diff
@@ -14,11 +290,11 @@ use super::*;
 /// meaning we can use `kani::unwind(0)` instead of `kani::unwind(2)`. Functionally,
 /// it works identically to `GuestMemoryMmap` with only a single contained region.
 pub struct SingleRegionGuestMemory {
-    the_region: vm_memory::GuestRegionMmap,
+    the_region: StubRegion,
```
stefano-garzarella, line 293: I'd split this patch in 2 patches, one where you introduce StubRegion (explaining why we need to do that) using it in the existing proofs and another patch where you add the new proof.

stefano-garzarella (commented):
Please for the commit description, follow the style of other commits, using the crate as prefix, something like this:

`virtio-queue: add unit proof for add_used method (sec 2.8.22)`

virtio-queue/src/queue/verification.rs
```diff
@@ -269,3 +502,57 @@ fn verify_spec_2_7_7_2() {
...
+            .ok_or(vm_memory::GuestMemoryError::InvalidGuestAddress(
+                queue.used_ring,
+            ))?;
+    let val = mem.load(addr, Ordering::Acquire)?;
+    Ok(u16::from_le(val))
+}
+
+/// Kani proof harness for verifying the behavior of the `add_used` method of the `Queue`.
+///
+/// # Specification (VirtIO 1.3, Section 2.8.22: "Receiving Used Buffers From The Device")
```
MatiasVara, line 525: I think section 2.8 is for packed virtqueues whereas here we are only dealing with split virtqueues. I think the section should be 2.7.14 (see https://docs.oasis-open.org/virtio/virtio/v1.3/csd01/virtio-v1.3-csd01.html#x1-7100014)

MatiasVara: I think what we try to enforce here is `2.7.8.2 Device Requirements: The Virtqueue Used Ring`. I think we may need a comment like in firecraker in which we state all the things we do not verify, e.g., the order in which the used-ring indexes are updated.

virtio-queue/src/queue/verification.rs
priyasiddharth: done

virtio-queue/src/queue/verification.rs
priyasiddharth: done

virtio-queue/src/queue/verification.rs
priyasiddharth: This is the offset of the `idx` field. 

https://docs.oasis-open.org/virtio/virtio/v1.3/csd01/virtio-v1.3-csd01.html#x1-540008
Following the convention of using bare numbers as in production code.

virtio-queue/src/queue/verification.rs
priyasiddharth: done

virtio-queue/src/queue/verification.rs
priyasiddharth: Not relevant here since proof moved to another PR

virtio-queue/src/queue/verification.rs
```diff
@@ -269,3 +502,57 @@ fn verify_spec_2_7_7_2() {
...
+        assert_eq!(queue.next_used.0, get_used_idx(&queue, &memory).unwrap());
+        assert_eq!(old_num_added + Wrapping(1), queue.num_added);
+        kani::cover!();
+    } else {
+        // On failure, we expect the next_used to remain unchanged
+        // and the used_desc_table_index to be out of bounds.
+        assert_eq!(queue.next_used, used_idx);
+        assert!(used_desc_table_index >= queue.size());
+        // The old value should still be the same as before the add_used call.
+        assert_eq!(old_val, get_used_idx(&queue, &memory).unwrap());
```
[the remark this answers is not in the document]
priyasiddharth: resolving for now

virtio-queue/src/queue/verification.rs
stefano-garzarella: @priyasiddharth I meant 2 patches in this PR, why removing the other changes from this PR?

virtio-queue/src/queue/verification.rs
```diff
@@ -269,3 +502,57 @@ fn verify_spec_2_7_7_2() {
...
+}
+
+/// Kani proof harness for verifying the behavior of the `add_used` method of the `Queue`.
+///
+/// # Specification (VirtIO 1.3, Section 2.8.22: "Receiving Used Buffers From The Device")
+///
+/// When the device has finished processing a buffer, it must add an element to the used ring,
+/// indicating which descriptor chain was used and how many bytes were written. The device must
+/// increment the used index after writing the element. If the descriptor index is out of bounds,
+/// the operation must fail and the used index must not be incremented.
```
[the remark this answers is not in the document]
stefano-garzarella: @priyasiddharth sorry, just to be clear, in my mind commit = patch + description.
PR (or series) is a set of commits/patches.

virtio-queue/src/queue/verification.rs
priyasiddharth: We are **not** enforcing any of the following in the current proof therefore I don't think `2.7.8.2` applies:
1. The device MUST set len prior to updating the used idx.
2. The device MUST write at least len bytes to descriptor, beginning at the first device-writable buffer, prior to updating the used idx.
3. The device MAY write more than len bytes to descriptor. Note: There are potential error cases where a device might not know what parts of the buffers have been written. This is why len is permitted to be an underestimate: that’s preferable to the driver believing that uninitialized memory has been overwritten when it has not.

virtio-queue/src/queue/verification.rs
priyasiddharth: fixed

virtio-queue/src/queue/verification.rs
priyasiddharth: added what we do not verify in doc

stefano-garzarella (commented):
minor comments from my side. LGTM but I'd like an approval from @MatiasVara and @roypat

virtio-queue/src/queue/verification.rs
```diff
@@ -507,3 +507,61 @@ fn verify_spec_2_7_7_2() {
...
+                queue.used_ring,
+            ))?;
+    let val = mem.load(addr, Ordering::Acquire)?;
+    Ok(u16::from_le(val))
+}
+
+/// Kani proof harness for verifying the behavior of the `add_used` method of
+/// the `Queue`.
+///
+/// # Specification (VirtIO 1.3, Section 2.7.14: "Receiving Used Buffers From The Device")
```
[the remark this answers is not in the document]
priyasiddharth: done

virtio-queue/src/queue/verification.rs
```diff
@@ -458,6 +458,7 @@ impl kani::Arbitrary for ProofContext {
     }
 }
 
+/// Kani proof harness for VirtIO 1.2 Section 2.7.7.2: Device-to-driver notification suppression.
```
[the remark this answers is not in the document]
priyasiddharth: fixed spec version number. I am leaving the function name as-is since we have a consistent doc naming scheme for spec version. Also adding two many digits to the function name is confusing imo.

virtio-queue/src/queue/verification.rs
```diff
@@ -458,19 +458,22 @@ impl kani::Arbitrary for ProofContext {
     }
 }
 
+/// # Specification (VirtIO 1.3, Section 2.7.7.2: "Device-to-driver notification suppression")
```
stefano-garzarella, line 461: In virtio 1.3 (https://docs.oasis-open.org/virtio/virtio/v1.3/csd01/virtio-v1.3-csd01.html#x1-510007) the section 2.7.7.2 is "2.7.7.2 Device Requirements: Used Buffer Notification Suppression", so I guess here we should update the title. (Is the 2.7.7.2 correct?)

virtio-queue/src/queue/verification.rs
```diff
@@ -2,10 +2,291 @@
...
+    kani::assume(write_offset <= buffer.len() - 16);
+    assert!(region
+        .write(&bytes, MemoryRegionAddress(write_offset as u64))
+        .is_ok());
+
+    // Read back into a new buffer
+    let mut readback = kani::vec::exact_vec::<u8, 16>();
+    assert!(region
+        .read(&mut readback, MemoryRegionAddress(write_offset as u64))
+        .is_ok());
```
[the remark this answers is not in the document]
epilys: Thanks!

epilys (approved):

stefano-garzarella (approved):
LGTM. Since it's the last week for @priyasiddharth GSoC, I'm going to merge this.
@MatiasVara @roypat I know you're busy, so if you will find something to fix, we can do later in a follow up PR.

Issue #347, opened 2025-06-25
virtio-bindings: Don't generate bindings at build time


jinankjain:
The current virtio-bindings crate generates bindings at build time, which introduces a dependency on bindgen and the Clang libraries. These dependencies can be challenging to satisfy in build server environments. Ideally, similar to other rust-vmm crates, we should pre-generate these bindings and include them in the repository. This issue has also been encountered in the Cloud Hypervisor project, as seen in [PR #7123](https://github.com/cloud-hypervisor/cloud-hypervisor/pull/7123). I’d like to hear the community’s thoughts on how best to approach this.

phip1611:
Sounds reasonable to decouple this from the build step. A quick workaround might be to get a reproducible and pinned toolchain via a Nix shell, working everywhere. Similar as [done here](https://github.com/rust-osdev/uefi-rs/blob/ac54da4e8ed9b3c86c9a27fefd44f43e54759640/.github/workflows/developer_productivity.yml#L28).

alyssais:
We used to check in the bindings until https://github.com/rust-vmm/vm-virtio/pull/327, but then discovered that they aren't architecture-independent, so generating them at build time fixed a real issue.  The alternative would be reverting that change, and pre-generating and checking in different bindings for each architecture we want to support.

A Nix shell could be useful for that, because we could take advantage of Nix's very good cross compilation support to get all the relevant headers.

jinankjain:
I think we can generate architecture dependent bindings that's what we have been doing in rust-vmm/mshv crate.

alyssais:
I've opened https://github.com/rust-vmm/vm-virtio/pull/348 to check in the bindings per architecture, with a script to update them going forward.

jinankjain:
@alyssais are you planning to work on the new release of crate as well?

PR #348, opened 2025-06-25
virtio-bindings: check in generated bindings again

alyssais:
### Summary of the PR

This reverts 2c21f25 ("virtio-bindings: use bindgen library from build.rs"), and goes back to including generated bindings in the repo, because users have found the requirement to have an up to date libclang at build time to be onerous.  We started generating the bindings at build time because we discovered they vary per architecture, so I've added a new generate-bindings.sh script that will generate bindings for a list of architectures, which will be easy to change in future.

Fortunately endianness doesn't seem to make a difference to the generated bindings.

Fixes: https://github.com/rust-vmm/vm-virtio/issues/347

### Requirements

requirements:

  test.

### DIFF
--- a/virtio-bindings/CHANGELOG.md
+++ b/virtio-bindings/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Bindings are now once again distributed with the crate, rather than
   generated at build time.
+- Regenerate bindings with Linux 6.15.
 
 # v0.2.5
 
--- a/virtio-bindings/src/bindings/aarch64/virtio_net.rs
+++ b/virtio-bindings/src/bindings/aarch64/virtio_net.rs
@@ -466,6 +466,49 @@ const _: () = {
 };
 #[repr(C)]
 #[derive(Debug, Default)]
+pub struct virtio_net_rss_config_hdr {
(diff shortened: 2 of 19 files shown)


epilys:
@alyssais Could we keep the `build.rs` solution easily with your approach behind an opt-in `feature`? I think both approaches are useful, and while not requiring clang for building seems to be the most convenient choice here, if we can provide both alternatives easily we probably should. (With the caveat that even if it's easy to make it opt-in, we still need to add tests for the CI to make sure it doesn't break in the future)

rbradford (approved):

jinankjain (approved):

stefanha-ai:
@stefano-garzarella As long as the bindings generation exhaustively covers all targets then it should be fine. The previous issue was using bindings generated for a single target across all targets, which doesn't work due to portability issues.

stefanha:
@stefano-garzarella As long as bindings are generated exhaustively for all targets it should work. The previous issue was that bindings generated for one target were being used across all targets - this doesn't work for portability reasons.

epilys (approved):

PR #362, opened 2025-09-05
virtio-bindings: Fix powerpc64 bindings

slp:
Change generate-bindings.sh to generate bindings for both powerpc64 and powerpc64le (let's drop 32-bit powerpc) and fix src/lib.rs to import to right headers for each powerpc64 variant according to its endianness.


epilys (dismissed):

epilys:
@stefano-garzarella ah you're right, they need to be `git rm`'ed.

slp:
> Should we also remove the `virtio-bindings/src/bindings/powerpc` folder?

Done, thanks @stefano-garzarella

stefano-garzarella (changes requested):
@slp thanks for fixing, but please update also the `virtio-bindings/CHANGELOG.md`

slp:
> @slp thanks for fixing, but please update also the `virtio-bindings/CHANGELOG.md`

I don't know how much detail do we want in the changelog, please let me know if you need it to be more verbose.

PR #363, opened 2025-09-15
virtio-queue: kani proofs for virtio queue

priyasiddharth:
### Summary of the PR

This PR adds new Kani proofs for virtio queue mainly around notification logic.

### Requirements

requirements:

- [X] All commits in this PR have Signed-Off-By trailers (with
- [X] All added/changed functionality has a corresponding unit/integration
  test.
- [X] All added/changed public-facing functionality has entries in the "Upcoming 
- [X] Any newly added `unsafe` code is properly documented.

### DIFF
--- a/virtio-queue/src/queue/verification.rs
+++ b/virtio-queue/src/queue/verification.rs
@@ -565,3 +565,139 @@ fn verify_add_used() {
         kani::cover!();
     }
 }
+
+/// This proof checks that after setting the `next_used` field of the queue
+/// using `set_next_used(x)`, reading back the value of `next_used` returns the
+/// same value `x`.
+#[kani::proof]
+#[kani::unwind(0)]
+fn verify_used_ring_avail_event() {
+    let ProofContext {
+        mut queue,
+        memory: _,
+    } = kani::any();
+    let x = kani::any();
+    queue.set_next_used(x);
+    assert_eq!(x, queue.next_used.0);
+}
+
+/// # Specification (VirtIO 1.3, Section 2.7.6.1: "Driver Requirements: The Virtqueue Available Ring")
+///
+/// This proof checks that:
+/// - If there are pending entries in the avail ring (avail_idx != next_avail),
+///   `enable_notification` returns true.
+/// - If there are no pending entries (avail_idx == next_avail), it returns false.
+/// This matches the monotonicity property of the avail ring in VirtIO 1.3 Section 2.7.6.1.
+#[kani::proof]
+#[kani::unwind(0)]
+fn verify_enable_driver_to_device_notification() {
+    let ProofContext { mut queue, memory } = kani::any();
+
+    // The enable_notification method sets notification to true and returns
+    //   - true, if there are pending entries in the `idx` field of the
+    //     avail ring
+    //   - false, if there are no pending entries in the `idx` field of the
+    // avail ring The check for pending entries is done by comparing the
+    // current `avail_idx` with the `next_avail` field of the queue. If they
+    // are different, there are pending entries, otherwise there are no
+    // pending entries. The equality check is a consequence of the
+    // monotonicity property of `idx` (2.7.6.1) that it cannot be decreased
+    // by the driver.
+    if queue.enable_notification(&memory).unwrap() {
+        assert_ne!(
+            queue.avail_idx(&memory, Ordering::Relaxed).unwrap(),
+            queue.next_avail
+        );
+    } else {
+        assert_eq!(
+            queue.avail_idx(&memory, Ordering::Relaxed).unwrap(),
+            queue.next_avail
+        );
+    }
+}
+
+// Helper method that reads `val` from the `avail_event` field of the used ring, using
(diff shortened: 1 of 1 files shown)


priyasiddharth:
@MatiasVara FYI

virtio-queue/src/queue/verification.rs
```diff
@@ -565,3 +565,139 @@ fn verify_add_used() {
...
+/// This proof checks:
+/// - If event_idx is not enabled, `set_notification(false)` sets used.flags to 1 (NO_NOTIFY).
+/// - If event_idx is enabled, the call is a no-op.
+#[kani::proof]
+#[kani::unwind(0)]
+fn verify_set_notification_false() {
+    let ProofContext { mut queue, memory } = kani::any();
+    let result = queue.set_notification(&memory, false /* disable notification */);
+    if !queue.event_idx_enabled {
+        // Check for Sec 2.7.10
```
stefano-garzarella, line 692: `Sec`, should be `Spec` right?

Also the test doc talks about `Specification (VirtIO 1.3, Section 2.7.7: "Used Buffer Notification Suppression")` what 2.7.10 is? Is related to virtio 1.3?

Please be more verbose with that references.

virtio-queue/src/queue/verification.rs
```diff
@@ -459,6 +459,17 @@ impl kani::Arbitrary for ProofContext {
...
+    // Check if point is in the range [start, stop) in a wrapping manner.
+    if start <= stop {
+        start <= point && point < stop
+    } else {
+        // If start > stop, we wrap around the maximum value of u16.
+        point >= start || point < stop
+    }
+}
+
 /// # Specification (VirtIO 1.3, Section 2.7.7.2: "Device Requirements: Used Buffer Notification Suppression")
```
stefano-garzarella, line 473: The commit https://github.com/rust-vmm/vm-virtio/pull/363/commits/4a8b048604ee45dfcdc3e5219161e35b249a1b89 description is not really clear to me.

It seems now this function don't test only the suppression, so should we update this documentation?

MatiasVara: I think the proof is testing both w/wo suppression. I think the comment should only mention about the notification.

virtio-queue/src/queue/verification.rs
```diff
@@ -565,3 +565,139 @@ fn verify_add_used() {
...
+
+    mem.load::<u16>(addr, order).map_err(Error::GuestMemory)
+}
+
+// Get the value of the `flags` field of the used ring, applying the specified ordering.
+fn get_used_flags<M: GuestMemory>(queue: &Queue, mem: &M, order: Ordering) -> Result<u16, Error> {
+    mem.load::<u16>(queue.used_ring, order)
+        .map(u16::from_le)
+        .map_err(Error::GuestMemory)
+}
```
MatiasVara, line 643:
```suggestion
}

```

virtio-queue/src/queue/verification.rs
MatiasVara: I think here you could add that Section 2.7.10 of the Virtio spec explains how the device can indicates to the driver that notifications are not required when adding buffers into the avail right.

virtio-queue/src/queue/verification.rs
```diff
@@ -565,3 +565,139 @@ fn verify_add_used() {
...
+/// # Specification (VirtIO 1.3, Section 2.7.7: "Used Buffer Notification Suppression")
+///
+/// This proof checks:
+/// - If event_idx is not enabled, `set_notification(false)` sets used.flags to 1 (NO_NOTIFY).
+/// - If event_idx is enabled, the call is a no-op.
+#[kani::proof]
+#[kani::unwind(0)]
+fn verify_set_notification_false() {
+    let ProofContext { mut queue, memory } = kani::any();
+    let result = queue.set_notification(&memory, false /* disable notification */);
```
MatiasVara, line 690: Can't we use `kany::any` to cover the potential values of the second parameter of set_notification? This would prevent to have to proofs.

virtio-queue/src/queue/verification.rs
```diff
@@ -565,3 +565,139 @@ fn verify_add_used() {
...
+    queue.set_next_used(x);
+    assert_eq!(x, queue.next_used.0);
+}
+
+/// # Specification (VirtIO 1.3, Section 2.7.6.1: "Driver Requirements: The Virtqueue Available Ring")
+///
+/// This proof checks that:
+/// - If there are pending entries in the avail ring (avail_idx != next_avail),
+///   `enable_notification` returns true.
+/// - If there are no pending entries (avail_idx == next_avail), it returns false.
```
MatiasVara, line 589: I think `VirtIO 1.3, Section 2.7.6.1` states that `avail_idx` is always >= than `next_avail`. Although, it is correct that enable_notification checks if they are different.

stefano-garzarella:
@priyasiddharth are you still interested in getting this PR merged? If yes, could you please re-open it in https://github.com/rust-vmm/rust-vmm/. We moved `vm-virtio` crates there.

PR #371, opened 2025-10-25
Follow-up to #368 — change verification to follow new vm-memory code

priyasiddharth:
### Summary of the PR

Fix to virtio-queue verification code to follow new vm-memory code
### Requirements

requirements:

  test.

### DIFF
--- a/virtio-queue/src/queue/verification.rs
+++ b/virtio-queue/src/queue/verification.rs
@@ -571,8 +571,10 @@ fn get_used_idx(
 /// if the descriptor index is out of bounds, the operation must fail and the
 /// used index must not be incremented. Note that this proof does not verify
 /// Section 2.7.8.2: "Device Requirements: The Virtqueue Used Ring"
-#[kani::proof]
-#[kani::unwind(0)]
+// Re-enable this proof once https://github.com/rust-vmm/vm-virtio/issues/373
+// is fixed.
+// #[kani::proof]
+// #[kani::unwind(0)]
 fn verify_add_used() {
     let ProofContext { mut queue, memory } = kani::any();
     let used_idx = queue.next_used;


priyasiddharth:
Will fix the build failures

priyasiddharth:
The GuestMemory architecture has changed beyond the addition/deletion of methods. Currently load from our StubRegion is not working.
The symptom is :
```
Failed Checks: This is a placeholder message; Kani doesn't support message formatted at runtime
 File: "/Users/runner/.rustup/toolchains/nightly-2025-04-24-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/res$
```
Likely an unwrap is being called on the value being loaded (the happy path is that there is no failure and hence kani does not have to compute an error message) and I suspect https://github.com/rust-vmm/vm-memory/pull/339 is the source of the issue. I ran out of time to look more into this. Can someone who is familiar with vm-memory take a look?

XanClic:
Yes, at least a stack trace to know where the failure is would be good.

> suspect https://github.com/rust-vmm/vm-memory/pull/339 is the source of the issue.

As far as I can tell, the only panic that it adds to the `load()` path could happen when trying to load/store an object of size 0. Specifically, the panic checks that `.get_slices()` returns no empty iterator, which can only happen when the object would have size 0.

https://github.com/rust-vmm/vm-memory/pull/349 added another panic, checking that `GuestMemoryRegion::get_slice()` returns a slice of exactly the requested length (after we ensured that the `start..(start + len)` range will fit into the region). So it could be that `StubRegion` doesn’t do that.

XanClic:
Adding this to `impl GuestMemoryRegion for StubRegion` seems to work:
```rust
fn get_slice(
    &self,
    offset: MemoryRegionAddress,
    count: usize,
) -> GuestMemoryResult<VolatileSlice<()>> {
    Ok(unsafe {
        VolatileSlice::with_bitmap(
            self.buffer.add(offset.raw_value() as usize),
            count,
            (),
            None,
        )
    })
}
```

I don’t know if it’s right, I just copied it from https://github.com/rust-vmm/vm-memory/blob/main/src/mmap/unix.rs#L394 – but at least it does make the test pass.

MatiasVara:
> @XanClic thank you very much! After applying your fix, now I have another failure that is unrelated to that issue.
> 
> ```
> Check 5471: vm_memory::volatile_memory::copy_slice_impl::copy_slice_volatile::{closure
> #0}.unwind.0
>         - Status: FAILURE
>         - Description: "unwinding assertion loop 0"
>         - Location: ../../.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/vm-mem
> ory-0.17.1/src/volatile_memory.rs:1346:13 in function vm_memory::volatile_memory::copy
> _slice_impl::copy_slice_volatile::{closure#0}
> ```
> 
> I guess it's related `#[kani::unwind(0)]` on top of `verify_add_used()` but I don't know the details. @priyasiddharth @MatiasVara can you check it?
> 
> Otherwise we should disable kani in the CI, since we would like to do a new release with the new vm-memory to update vhost/vhot-devices.

I will give a look, apologies for the delay.

priyasiddharth:
I looked into this a bit.
1. The `with_bitmap` VolatileSlice call looks fine.
2. The problem seems to be that we expect to iterate through VolatileSlices in `load` and `store` methods in [1]. To get through this, we need to `unroll` [2] once. However, Kani does not support per loop unrolling and the setting all loops to unroll to one gets the execution into a bad case of forever unrolling (i killed after a couple of minutes).

To validate this hypothesis, a programmer can manually unroll the iterator once and then iterate when `len(volatile_slices)` are > 1 in [1]. We might need to have a verification friendly version of `GuestMemory` if this hypothesis is true.

References
* [1] - https://github.com/rust-vmm/vm-memory/commit/66b9427e5b9ca6523e763363fec82480eb21e313
* [2] - https://model-checking.github.io/kani/tutorial-loop-unwinding.html

MatiasVara:
Thanks @priyasiddharth! I think this PR can be split into two commits. First, the changes that correspond with following new vm-memory. Second, the disabling of the verify_add_used proof. WDYT?

MatiasVara:
Thanks @priyasiddharth! I think this PR can be split into two commits. First, the changes that correspond with following new vm-memory. Second, the disabling of the verify_add_used proof. WDYT?

stefano-garzarella (approved):

PR #379, opened 2025-12-17
Update vm-memory dependency to 0.18.0

XanClic:
### Summary of the PR

Update the vm-memory dependency to 0.18.0. This PR supersedes https://github.com/rust-vmm/vm-virtio/pull/378.

0.18.0 is a breaking change, first and foremost because it renames the `GuestMemory` trait to `GuestMemoryBackend`, replacing it with a new `GuestMemory` trait that can abstract not only physical but also virtual memory, but in turn has a limited interface and requires specifying the required permissions for any memory access.

(Luckily, most memory accesses use the `Bytes` trait, which already implicitly specifies the access mode by way of which method is used, e.g. `load()` vs. `store()`.)

This PR first changes some calls to use the methods that will still be available:
* `check_range()` instead of `check_address()`
* Specifically for `virtio_queue::descriptor_utils::{Reader, Writer}::new()`: Do not resolve memory accesses through guest memory regions, just collect slices through `get_slices()`
* For `virtio_vsock::packet`: Replace `get_slice()` by a helper function that calls `get_slices()` and returns an error if the given guest memory region maps to more than a single slice. This is not ideal: `get_slice()` has been removed to make users aware that guest memory regions may not always map to a single slice in their memory, especially for virtual memory, and this is just ignoring that. However, it shouldn’t be a regression, so support for this can be implemented later, I think. (Maybe via bounce buffers if the interface must stay compatible?)
* `get_host_address()` is removed; we only need it for testing, so we can just use `get_single_slice().as_ptr()` there, basically.

Then it bumps the vm-memory version, at which point we need to:
* Pass the required access permissions,
* Deal with `get_slices()` returning a `Result<Iterator, E>` instead of a plain `Iterator`,
* Get the bitmap type via `GuestMemory::Bitmap` instead of `<GuestMemory::R as GuestMemoryRegion>::B`, and
* Make the verification code’s `SingleRegionGuestMemory` implement `GuestMemoryBackend` instead of `GuestMemory`.

#### Compatibility Note

Note that users should still be able to depend on vm-memory 0.17.2+, even with vm-virtio updating to 0.18.0. This is because 0.17.2 internally depends on 0.18.0 and just re-exports all symbols, just some under different names (i.e. `GuestMemoryBackend` as `GuestMemory`).

### Requirements

requirements:

  test.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,5 +15,5 @@ lto = true
 codegen-units = 1
 
 [workspace.dependencies]
-vm-memory = "0.17.1"
+vm-memory = "0.18.0"
 vmm-sys-util = "0.15.0"
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -20,7 +20,7 @@ memfd = "0.6.3"
 virtio-queue = { path = "../virtio-queue", features = ["test-utils"] }
 virtio-vsock = { path = "../virtio-vsock" }
 virtio-queue-ser = { path = "../virtio-queue-ser" }
-vm-memory = { version = "0.17.1", features = ["backend-mmap", "backend-atomic"] }
+vm-memory = { version = "0.18.0", features = ["backend-mmap", "backend-atomic"] }
 common = { path = "common" }
 virtio-blk = { path = "../virtio-blk", features = ["backend-stdio"] }
 
--- a/fuzz/common/Cargo.toml
+++ b/fuzz/common/Cargo.toml
@@ -13,4 +13,4 @@ virtio-queue = { path = "../../virtio-queue", features = ["test-utils"] }
 virtio-vsock = { path = "../../virtio-vsock" }
 virtio-queue-ser = { path = "../../virtio-queue-ser" }
 virtio-blk = { path = "../../virtio-blk" }
-vm-memory = { version = "0.17.1", features = ["backend-mmap", "backend-atomic"] }
+vm-memory = { version = "0.18.0", features = ["backend-mmap", "backend-atomic"] }
--- a/fuzz/common/src/vsock.rs
+++ b/fuzz/common/src/vsock.rs
@@ -177,7 +177,7 @@ mod tests {
     use virtio_queue::desc::RawDescriptor;
     use virtio_queue::mock::MockSplitQueue;
     use virtio_vsock::packet::VsockPacket;
-    use vm_memory::{Bytes, GuestAddress, GuestMemory, GuestMemoryMmap};
+    use vm_memory::{Bytes, GuestAddress, GuestMemory, GuestMemoryMmap, Permissions};
 
     // Random values to be used by the tests for the header fields.
     const SRC_CID: u64 = 1;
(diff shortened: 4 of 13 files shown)


XanClic:
Split out the `check_range()` implementation for kani into a separate commit.

I don’t know why the x64 fuzzing tests failed, I can’t see anything in the logs. It looks like they all have been terminated from the outside at the same time (12:09:24 UTC), so it doesn’t look like this PR is the problem, but maybe a timeout.

stefano-garzarella (approved):
@XanClic I restarted failing jobs and everything is fine, I guess some issue on the worker side.

rbradford:
Thanks for working on this. I'd like to see us first make a release with bumping the dependency on vm-memory to 0.17.2 which is API compatible with the existing versions. Before we then adopt vm-memory 0.18.0 everywhere.

XanClic:
> Thanks for working on this. I'd like to see us first make a release with bumping the dependency on vm-memory to 0.17.2 which is API compatible with the existing versions. Before we then adopt vm-memory 0.18.0 everywhere.

~~I’m not sure that is true~~ I believe updating vm-virtio’s dependency to 0.18.0 will allow its users to continue to use 0.17.2, because 0.17.2+ is basically empty; it depends on 0.18.0 and re-exports everything, just sometimes renamed. So if vm-virtio depends on 0.18.0, that shouldn’t interfere with its users depending on 0.17.2+, because 0.17.2+ pulls in the 0.18.0 dependency anyway.

If an 0.17.2+ user uses `vm-memory-0.17.2::GuestMemory`, as far as I understand, that’s just an alias for `vm-memory-0.18.0::GuestMemoryBackend` (because vm-memory 0.17.2 just re-exports 0.18’s `GuestMemoryBackend` as `GuestMemory`).

(I’m saying 0.17.2+ because we’ll need an 0.17.3 release with https://github.com/rust-vmm/vm-memory/pull/369 to make it usable.)

rbradford (approved):

rbradford:
> > Thanks for working on this. I'd like to see us first make a release with bumping the dependency on vm-memory to 0.17.2 which is API compatible with the existing versions. Before we then adopt vm-memory 0.18.0 everywhere.
> 
> ~I’m not sure that is true~ I believe updating vm-virtio’s dependency to 0.18.0 will allow its users to continue to use 0.17.2, because 0.17.2+ is basically empty; it depends on 0.18.0 and re-exports everything, just sometimes renamed. So if vm-virtio depends on 0.18.0, that shouldn’t interfere with its users depending on 0.17.2+, because 0.17.2+ pulls in the 0.18.0 dependency anyway.
> 
> If an 0.17.2+ user uses `vm-memory-0.17.2::GuestMemory`, as far as I understand, that’s just an alias for `vm-memory-0.18.0::GuestMemoryBackend` (because vm-memory 0.17.2 just re-exports 0.18’s `GuestMemoryBackend` as `GuestMemory`).
> 
> (I’m saying 0.17.2+ because we’ll need an 0.17.3 release with [rust-vmm/vm-memory#369](https://github.com/rust-vmm/vm-memory/pull/369) to make it usable.)

Thanks for sharing - let's move ahead with this PR.

PR #382, opened 2026-01-15
virtio-bindings: fix several architecture-specific issues

alyssais:
### Summary of the PR

See commit messages.

Closes https://github.com/rust-vmm/vm-virtio/issues/381

### Requirements

Before submitting your PR, please make sure you addressed the following
requirements:

- [x] All commits in this PR have Signed-Off-By trailers (with
  `git commit -s`), and the commit message has max 60 characters for the
  summary and max 75 characters for each description line.
- [x] All added/changed functionality has a corresponding unit/integration
  test.
- [x] All added/changed public-facing functionality has entries in the "Upcoming 
  Release" section of CHANGELOG.md (if no such section exists, please create one).
- [x] Any newly added `unsafe` code is properly documented.

PR #385, opened 2026-01-18
virtio-bindings: regenerate with Linux 6.17

alyssais:
### Summary of the PR

6.18 is the current release, but it contains something we may consider a breaking change, so let's get this easy update out of the way first.

### Requirements

requirements:

  test.

### DIFF
--- a/virtio-bindings/CHANGELOG.md
+++ b/virtio-bindings/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Upcoming Release
 
+## Changed
+
+- Regenerate bindings with Linux 6.17.
+
 # v0.2.7
 
 ## Changed
--- a/virtio-bindings/src/bindings/aarch64/virtio_net.rs
+++ b/virtio-bindings/src/bindings/aarch64/virtio_net.rs
@@ -64,6 +64,12 @@ pub const VIRTIO_NET_F_RSS: u32 = 60;
 pub const VIRTIO_NET_F_RSC_EXT: u32 = 61;
 pub const VIRTIO_NET_F_STANDBY: u32 = 62;
 pub const VIRTIO_NET_F_SPEED_DUPLEX: u32 = 63;
+pub const VIRTIO_NET_F_GUEST_UDP_TUNNEL_GSO: u32 = 65;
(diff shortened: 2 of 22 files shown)


epilys (dismissed):

stefano-garzarella:
Thanks for this, BTW should we do a release before merging this? (I mean, we can merge and tag the commit before this, just to include #382 before updating to 6.17. WDYT?

stefano-garzarella:
> Thanks for this, BTW should we do a release before merging this? (I mean, we can merge and tag the commit before this, just to include #382 before updating to 6.17. WDYT?

Ah no, we need to update cargo.toml, changelog, etc. so better to do the release, and after that merge this. I'll work on that soon.

stefano-garzarella (changes requested):
@alyssais can you add a line in `virtio-bindings/CHANGELOG.md` and rebase on top of #386

stefano-garzarella (approved):

stsquad:
The net_v1_hdr to the upstream uapi headers broke QEMU so we might want to wait until that is sorted out before merging. There was a debate as to if they should be exposed to the UAPI at all. I believe @epilys was going to suggest a patch.

epilys:
@stsquad QEMU doesn't use `virtio-bindings` crate

stefano-garzarella:
> The net_v1_hdr to the upstream uapi headers broke QEMU so we might want to wait until that is sorted out before merging. There was a debate as to if they should be exposed to the UAPI at all. I believe @epilys was going to suggest a patch.

oh, good to know, do you have a link to the discussion ?

epilys (approved):

stsquad:
> @stsquad QEMU doesn't use `virtio-bindings` crate

I meant does the uapi breakage that affected QEMU with the C header affect rust the same way?

epilys:
@stsquad 

> I meant does the uapi breakage that affected QEMU with the C header affect rust the same way?

No, and I just realised this is 6.17, not 6.19 which was the problem with QEMU. In any case since we semver our releases, so if a consumer of this create depends on a struct definition that was changed in an ABI-compatible way can just change their code when they update the Cargo dependency.

We're good to go, in other words

PR #387, opened 2026-01-23
virtio-bindings: regenerate with Linux 6.18

alyssais:
So, how do we handle the breaking change here, if at all?

### Requirements

requirements:

  test.

### DIFF
--- a/virtio-bindings/CHANGELOG.md
+++ b/virtio-bindings/CHANGELOG.md
@@ -2,7 +2,11 @@
 
 ## Changed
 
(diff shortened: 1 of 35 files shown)


epilys:
> So, how do we handle the breaking change here, if at all?

New release?

rbradford:
A new major version should be fine - I don't think too many people use that struct member

epilys (approved):

stefano-garzarella (approved):

PR #391, opened 2026-03-04
virtio-queue: implement From for QueueSync

avagin:
Add From trait implementations for QueueSync to allow creating it from an existing Queue or Arc<Mutex<Queue>>.


stefano-garzarella (approved):

rbradford (approved):
I have no problem with this - but is this a normal Rusty thing to do?

epilys (approved):

PR #400, opened 2026-08-12
virtio-queue: fix SplitQueueRing address calculations

umangPokhriyall:
### Summary of the PR

SplitQueueRing is generic over the ring element type T, but two address
calculations did not account for the element size correctly:

- new() used size_of::<u16>() when locating the event field.
- end() treated the element count as a byte count and omitted the
  trailing event field.

The first happens to work for AvailRing, whose elements are u16, but is
incorrect for UsedRing, whose elements are VirtqUsedElem (8 bytes).
Because MockSplitQueue::create() places the used ring at avail.end(),
this causes the two rings to overlap.

For example, with a queue size of 8, the available ring occupies
128..150 while the used ring starts at 140. This also causes the 
available ring's used_event field to overlap the first UsedRing element.

The fix calculates the event-field offset using size_of::<T>() and makes
end() include the complete ring layout, including the trailing event
field.

Added regression tests covering both AvailRing and UsedRing and multiple
queue sizes. The tests fail on the unpatched code and pass with the fix.

Validation:
- cargo fmt --all -- --check
- cargo clippy --all-features --workspace --all-targets -- -D warnings
- cargo test --all-features --workspace
- cargo doc --all-features --workspace --no-deps
- cargo audit -q --deny warnings

All checks pass.

The ring addresses returned by MockSplitQueue change as a result of the
fix. Existing in-tree callers use the address accessors rather than
depending on the previous layout.

### Requirements

requirements:

  test.
- [x] All added/changed public-facing functionality has entries in the
  "Upcoming Release" section of CHANGELOG.md (if no such section exists,
  please create one).


stefano-garzarella:
@umangPokhriyall are you still interested in getting this PR merged? If yes, could you please re-open it in https://github.com/rust-vmm/rust-vmm/. We moved `vm-virtio` crates there.

